PR #2, opened 2026-03-01
feat: add v1 agent-scoped memory scope for LCM tools

jacoblyles:
## Summary
- add initial agent-scoped memory plugin helpers under `src/plugins/agent-memory-scope`
- extend scope resolution and retrieval/store paths to support same-agent multi-conversation lookups
- add/adjust tests for agent scope and expand-query behavior

## Test
- `npx vitest run test/agent-memory-scope.test.ts test/lcm-tools.test.ts test/lcm-expand-query-tool.test.ts`

## Notes
- default behavior remains unchanged when scope is omitted
- follow-up can wire runtime config knobs (`allowAgentScope`, `maxAgentConversations`) from OpenClaw

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -9,6 +9,7 @@ import { join } from "node:path";
 import type { OpenClawPluginApi } from "openclaw/plugin-sdk";
 import { resolveLcmConfig } from "./src/db/config.js";
 import { LcmContextEngine } from "./src/engine.js";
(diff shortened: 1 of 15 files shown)


100yenadmin:
@jacoblyles triage pass update: I marked this priority:P3 enhancement/linked-pr/stale-check. The P0 delegated-retrieval leakage path was fixed by #768, so this older agent-scoped memory branch may now be partially superseded. Can you confirm whether there is remaining scope behavior here that #768 did not cover?

PR #14, opened 2026-03-08
fix: move pi-ai and pi-agent-core from peerDependencies to dependencies

### LINKED ISSUE #7: Plugin fails to resolve @mariozechner/pi-ai: dynamic import() bypasses jiti module resolution
## Bug

`index.ts` line 619 does a dynamic `await import("@mariozechner/pi-ai")` to build the `LcmDependencies.complete` function. This import always fails for fresh installs because:

1. `@mariozechner/pi-ai` is listed as a `peerDependency`
2. OpenClaw's plugin installer runs `npm install --omit=peer`, so it's never installed in the plugin directory
3. The dynamic `import()` resolves via Node's native module resolution from the plugin's filesystem location (`~/.openclaw/extensions/lossless-claw/`), NOT from OpenClaw's install directory — jiti does not intercept runtime `import()` calls
4. Node walks up the directory tree from the plugin dir and never reaches OpenClaw's `node_modules/` (whether installed via npm or bun global)

This means **every user** will hit `Cannot find module '@mariozechner/pi-ai'` on first install. Summarization falls back to deterministic truncation (all summaries end up at exactly 1215 tokens).

## Error log

```
Cannot find module '@mariozechner/pi-ai' imported from /Users/.../.openclaw/extensions/lossless-claw/index.ts
```

## Workaround

Manually install the peer deps in the plugin directory:

```bash
cd ~/.openclaw/extensions/lossless-claw
npm install --ignore-scripts @mariozechner/pi-ai @mariozechner/pi-agent-core
openclaw gateway restart
```

## Proposed fix

Move `@mariozechner/pi-ai` and `@mariozechner/pi-agent-core` from `peerDependencies` to `dependencies` in `package.json`. The peer dep pattern assumes the host provides the package in a shared `node_modules/` tree, but OpenClaw plugins live in an isolated directory (`~/.openclaw/extensions/`) where the host's packages are unreachable.

Alternatively, refactor `index.ts` to use the plugin SDK's completion interface instead of importing pi-ai directly — the `src/` files already do this cleanly via dependency injection.

## Environment

- OpenClaw 2026.3.7
- lossless-claw v0.2.2
- macOS, Node v25.6.0
- Installed via both `bun install -g` and `npm install -g` (same result)

BrennerSpear:
Fixes #7

## Problem

`index.ts` dynamically imports `@mariozechner/pi-ai` at runtime (`await import("@mariozechner/pi-ai")`). This package is listed as a `peerDependency`, but OpenClaw's plugin installer runs `npm install --omit=peer`, so it's never installed in the plugin directory.

Node's module resolution walks up the directory tree from the plugin's location (`~/.openclaw/extensions/lossless-claw/`), which never intersects with OpenClaw's global install path — regardless of whether the user installed via npm or bun. The dynamic `import()` is not intercepted by OpenClaw's jiti loader (jiti only aliases `openclaw/plugin-sdk` paths).

This causes every fresh install to fail with `Cannot find module '@mariozechner/pi-ai'`, and summarization silently falls back to deterministic truncation.

## Fix

Move `@mariozechner/pi-ai` and `@mariozechner/pi-agent-core` from `peerDependencies` to `dependencies`. `openclaw` correctly remains as a peer dep since its SDK paths are aliased by the host's jiti loader.

## Workaround (for users on current version)

```bash
cd ~/.openclaw/extensions/lossless-claw
npm install --ignore-scripts @mariozechner/pi-ai @mariozechner/pi-agent-core
openclaw gateway restart
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
     "LICENSE"
   ],
   "dependencies": {
+    "@mariozechner/pi-agent-core": "*",
+    "@mariozechner/pi-ai": "*",
     "@sinclair/typebox": "0.34.48"
   },
   "devDependencies": {
@@ -31,9 +33,7 @@
     "vitest": "^3.0.0"
   },
   "peerDependencies": {
-    "openclaw": "*",
-    "@mariozechner/pi-agent-core": "*",
-    "@mariozechner/pi-ai": "*"
+    "openclaw": "*"
   },
   "openclaw": {
     "extensions": [


BrennerSpear:
Duplicate of #8 which was already merged. Sorry for the noise!

PR #44, opened 2026-03-11
feat: PostgreSQL backend support

rivetphilbot:
## Summary

Adds PostgreSQL as an alternative database backend to the existing SQLite implementation. All store operations (conversations, messages, summaries, search) work against both backends through a unified `DbClient` interface and a SQL dialect helper that eliminates per-query backend branching.

## Key design decisions

- **Dialect helper** (`src/db/dialect.ts`) — handles placeholder style (`$1` vs `?`), timestamp functions (`NOW()` vs `datetime('now')`), and backend-specific SQL (e.g. `LPAD` vs `printf` for path construction). Zero `if (backend === 'postgres')` branching in query methods.
- **Transaction safety** — `withTransaction()` swaps `this.db` to the transaction-scoped client for the duration of the operation, ensuring all queries within a transaction run on the same connection. Critical for Postgres connection pooling where BEGIN/COMMIT could otherwise land on different connections.
- **Auto-schema creation** — `ensurePostgresSchema()` creates all tables, indexes, constraints, and generated tsvector columns idempotently on first use. Forward-compatible migrations add new columns to existing schemas.
- **Lazy loading** — `pg` module loaded via `createRequire()` only when PostgresClient is instantiated. SQLite-only users never need pg installed.
- **Connection pooling** — ref-counted shared pool keyed by connection string, with health checks.

### New files
- **`src/db/db-interface.ts`** — `DbClient` interface: `query<T>`, `queryOne<T>`, `run`, `transaction`, `close`
- **`src/db/postgres-client.ts`** — PostgreSQL implementation using `pg` (node-postgres), lazy-loaded
- **`src/db/sqlite-client.ts`** — SQLite implementation wrapping `node:sqlite`
- **`src/db/dialect.ts`** — SQL dialect helper for placeholder/function differences between backends
- **`src/store/tsquery-sanitize.ts`** — Sanitizes user input for `websearch_to_tsquery` (strips invalid characters)
- **`scripts/migrate-to-postgres.mjs`** — Data migration script: SQLite → PostgreSQL

### Modified files
- **`src/db/config.ts`** — New fields: `backend`, `connectionString`. Env vars: `LCM_BACKEND`, `LCM_CONNECTION_STRING`. Auto-detects backend from connection string presence.
- **`src/db/connection.ts`** — `createLcmConnection()` factory dispatches to SQLite or Postgres. Ref-counted pool.
- **`src/db/features.ts`** — Returns backend type and full-text availability.
- **`src/db/migration.ts`** — `ensurePostgresSchema()` for auto-creation, `migratePostgresSchema()` for forward-compatible column additions. SQLite migrations unchanged.
- **`src/engine.ts`** — `ensureMigrated()` now async, calls `ensurePostgresSchema()` for Postgres backend. Wires backend config to stores.
- **`src/store/conversation-store.ts`** — Rewritten to use Dialect helper. Transaction-safe. PostgreSQL full-text search via `websearch_to_tsquery`/`ts_headline`, regex via `~*`.
- **`src/store/summary-store.ts`** — Same Dialect-based rewrite.
- **`src/store/index.ts`** — Re-exports new types and utilities.
- **`src/tools/lcm-grep-tool.ts`** — Simplified timestamp formatting.
- **`src/assembler.ts`** — Fix: prevent raw toolCall blocks from bypassing argument normalization (was causing xAI/OpenAI 422 errors when conversation history contained tool calls).

## Configuration

```bash
# Environment variables
LCM_BACKEND=postgres
LCM_CONNECTION_STRING=postgres://user:pass@host:5432/dbname

# Or plugin config in openclaw.json
{
  "backend": "postgres",
  "connectionString": "postgres://..."
}
```

SQLite remains the default when no connection string is provided.

## PostgreSQL requirements

- PostgreSQL 14+ (uses generated tsvector columns, `websearch_to_tsquery`)
- `pg_trgm` extension (for regex search fallback) — optional but recommended
- Schema auto-created on first use; no manual DDL needed

## Testing

Running in production on three OpenClaw instances with ~10k messages across multiple agent sessions. Both SQLite (existing) and PostgreSQL paths exercised. Full-text search, regex search, compaction, and bootstrap all verified.

### DIFF
--- a/scripts/migrate-to-postgres.mjs
+++ b/scripts/migrate-to-postgres.mjs
@@ -12,22 +12,59 @@
 import Database from 'better-sqlite3';
 import pg from 'pg';
 import { readFileSync, readdirSync, existsSync } from 'fs';
-import { join, basename } from 'path';
+import { delimiter, join } from 'path';
 import { homedir } from 'os';
-import { createHash, randomUUID } from 'crypto';
 
 const { Pool } = pg;
-
-// Config
-const SQLITE_PATH = join(homedir(), '.openclaw', 'lcm.db');
-const PG_CONNECTION = 'postgres://lcm_phil:pheon.lcm4@10.4.20.16:5432/phil_memory';
-const AGENTS_DIR = join(homedir(), '.openclaw', 'agents');
-const BACKUP_AGENTS_DIR = join(homedir(), '.openclaw', 'config-backups', '2026-03-07', 'agents');
 
 const args = process.argv.slice(2);
 const DRY_RUN = args.includes('--dry-run');
 const SQLITE_ONLY = args.includes('--sqlite-only');
 const JSONL_ONLY = args.includes('--jsonl-only');
+
+function readFlagValue(flagName) {
+  const prefix = `${flagName}=`;
+  const match = args.find((arg) => arg.startsWith(prefix));
+  return match ? match.slice(prefix.length).trim() : undefined;
+}
+
+function readFlagValues(flagName) {
+  const prefix = `${flagName}=`;
+  return args
+    .filter((arg) => arg.startsWith(prefix))
+    .map((arg) => arg.slice(prefix.length).trim())
+    .filter(Boolean);
+}
+
+function dedupe(items) {
+  return [...new Set(items.filter(Boolean))];
+}
+
+function resolveMigrationConfig() {
+  const sqlitePath = readFlagValue('--sqlite-path')
+    || process.env.LCM_SQLITE_PATH?.trim()
+    || join(homedir(), '.openclaw', 'lcm.db');
+  const connectionString = readFlagValue('--pg-connection')
+    || process.env.LCM_CONNECTION_STRING?.trim()
+    || process.env.PG_CONNECTION_STRING?.trim()
+    || '';
+  const defaultAgentsDir = join(homedir(), '.openclaw', 'agents');
+  const extraAgentDirs = [
+    ...readFlagValues('--agents-dir'),
+    ...(process.env.LCM_AGENT_DIRS ?? '')
+      .split(delimiter)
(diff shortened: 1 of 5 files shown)


jalehman (changes requested):
I re-reviewed the three blockers from the previous round.

1. SQLite FTS probe: the intended fix is clear, but it is not landed cleanly yet. In `src/engine.ts`, the constructor now calls `getLcmConnection(this.config.databasePath)` at line 610 to feed a raw SQLite handle into `getLcmDbFeatures(...)`, but `getLcmConnection` is not imported at the top of the file (`src/engine.ts:20`). As written, the PR head does not type-check, so I can’t consider the SQLite FTS regression fully resolved.

2. Swallowed Postgres migration failures: fixed. `ensureMigrated()` now awaits `ensurePostgresSchema(db)` directly and only sets `this.migrated = true` after success (`src/engine.ts:662-667`). Failures will propagate instead of being logged and ignored.

3. Unsafe transaction client swap: fixed for the original concurrency issue. Both stores now keep a root client plus an `AsyncLocalStorage` transaction context, and `withTransaction()` no longer mutates shared instance state (`src/store/conversation-store.ts:223-256`, `src/store/summary-store.ts:241-265`). That removes the singleton race from swapping `this.db` during concurrent work.

There is also an additional current-head blocker: `src/tools/lcm-grep-tool.ts` has a dangling `+` at the end of the multi-line description string (`src/tools/lcm-grep-tool.ts:78-83`), which leaves the object literal unparsable before `parameters:`. Running `npx --yes -p typescript tsc --noEmit` against the PR checkout stops immediately with `src/tools/lcm-grep-tool.ts(85,15): error TS1005: ',' expected`.

Because the branch still has compile/parsing errors, I’m keeping this in changes-requested state.

PR #45, opened 2026-03-11
feat: embedding-based semantic search via pgvector

rivetphilbot:
## Summary

Adds semantic (embedding-based) search as a third search mode alongside regex and full-text. Uses OpenAI's `text-embedding-3-small` model and PostgreSQL's `pgvector` extension for cosine similarity search.

**Depends on:** #44 (PostgreSQL backend support)

### New files
- **`src/embeddings.ts`** — `EmbeddingClient` class for OpenAI-compatible embedding APIs. Handles batching (512 texts/call), auto-truncation, and API key resolution from env vars or config.

### Modified files
- **`src/db/config.ts`** — New fields: `embeddingApiKey`, `embeddingBaseUrl`, `embeddingModel`. Env vars: `LCM_EMBEDDING_API_KEY`, `LCM_EMBEDDING_BASE_URL`, `LCM_EMBEDDING_MODEL`. Falls back to `OPENAI_API_KEY`.
- **`src/db/migration.ts`** — `addEmbeddingColumnsIfAvailable()` auto-creates pgvector columns and HNSW indexes when the extension is available. Runs during schema creation and forward migrations. Silently skips if pgvector is not installed.
- **`src/engine.ts`** — Wires embedding config through to both stores.
- **`src/store/conversation-store.ts`** — Embed-on-insert (async, non-blocking), `searchSemantic()` via parameterized `<=>` cosine distance, graceful fallback to full_text if unavailable.
- **`src/store/summary-store.ts`** — Same pattern for summaries.
- **`src/store/index.ts`** — Re-exports `EmbeddingClient`, `toVectorLiteral`, `fromVectorLiteral`, `EmbeddingConfig`.
- **`src/retrieval.ts`** — `GrepInput.mode` expanded to include `"semantic"`.
- **`src/tools/lcm-grep-tool.ts`** — `mode` enum and description updated for semantic search.

## How it works

1. **Embed on insert:** When a message or summary is created, an embedding is generated asynchronously and stored. Failures are logged but never block writes.
2. **Semantic search:** `lcm_grep(mode="semantic")` embeds the query, then finds nearest neighbors via pgvector cosine distance (`<=>`). All query parameters (including vector literals and limits) are fully parameterized.
3. **Fallback:** If semantic search is unavailable (no embeddings, no API key, SQLite backend), falls back to full-text search transparently.
4. **Auto-setup:** Schema migration auto-creates `embedding vector(1536)` columns and HNSW indexes when pgvector is detected. No manual DDL needed.

## Configuration

```bash
# Required for semantic search
LCM_EMBEDDING_API_KEY=sk-...  # or OPENAI_API_KEY

# Optional overrides
LCM_EMBEDDING_BASE_URL=https://api.openai.com/v1  # default
LCM_EMBEDDING_MODEL=text-embedding-3-small          # default, 1536 dimensions
```

## Prerequisites

- PostgreSQL backend (#44)
- `pgvector` extension installed (columns and indexes are auto-created)

## Cost

text-embedding-3-small: $0.02 per 1M tokens. A 10k message backfill costs ~$0.05.

## Testing

Running in production with ~10k messages and ~1k summaries fully embedded. Semantic search returns relevant results across conversation history with sub-second latency.

### DIFF
--- a/src/store/conversation-store.ts
+++ b/src/store/conversation-store.ts
@@ -5,7 +5,7 @@ import type { DbClient } from "../db/db-interface.js";
 import { SqliteClient } from "../db/sqlite-client.js";
 import { Dialect, type Backend } from "../db/dialect.js";
 import { EmbeddingClient, toVectorLiteral, type EmbeddingConfig } from "../embeddings.js";
-import { EmbeddingQueue } from "../embedding-queue.js";
+import { EmbeddingQueue, type QueueableDb } from "../embedding-queue.js";
 import { sanitizeFts5Query } from "./fts5-sanitize.js";
 import { sanitizeTsQuery } from "./tsquery-sanitize.js";
 import { buildLikeSearchPlan, createFallbackSnippet } from "./full-text-fallback.js";
@@ -275,7 +275,7 @@ export class ConversationStore {
 
     // Set up batched embedding queue (replaces fire-and-forget embedOnInsert)
     if (this.embeddingClient) {
-      this.embeddingQueue = new EmbeddingQueue(this.embeddingClient, db, {
+      this.embeddingQueue = new EmbeddingQueue(this.embeddingClient, db as QueueableDb, {
         log: (msg) => console.error(`[lcm:conv-embedding-queue] ${msg}`),
       });
       this.embeddingQueue.start();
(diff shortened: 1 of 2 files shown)


coolmanns:
Love it!!!

PR #68, opened 2026-03-15
fix: align LCM summary model resolution with OpenClaw runtime and compaction config

uf-hy:
This PR makes LCM summarization resolve its model more predictably and aligns the plugin with newer OpenClaw runtime behavior.

## What changed
- centralize summary-model selection in `src/summarize.ts` with a defined priority order:
  1. plugin `summaryModel`
  2. `LCM_SUMMARY_MODEL`
  3. `agents.defaults.compaction.model`
  4. session model hint
  5. OpenClaw default model fallback in `index.ts`
- support OpenClaw `runtimeContext` while keeping backward compatibility with `legacyCompactionParams`
- allow `summaryModel` to resolve even when no explicit provider is configured, while still emitting a warning
- add `summaryModel` and `summaryProvider` to the plugin config schema
- honor OpenClaw `agents.defaults.compaction.model`, a configuration field introduced in OpenClaw v2026.3.8
- keep the merged branch compatible with upstream's newer async model-auth and API-key resolution flow

## Why
- summary-model resolution was previously split across multiple files and could fall back in surprising ways
- newer OpenClaw runtimes pass `runtimeContext`, while older integration paths still use `legacyCompactionParams`
- summarization should be steerable through plugin config, environment variables, and OpenClaw compaction config without hidden provider/model mismatches
- since OpenClaw v2026.3.8 introduced `agents.defaults.compaction.model`, LCM should respect that setting when choosing a summarization model

## Testing
- full test suite passes: 20 files / 259 tests
- added or strengthened regression coverage for:
  - summary model priority order
  - `compaction.model` support, including `{ primary: ... }`
  - `runtimeContext` compatibility
  - `modelAuth` and auth-profile fallback behavior

## Reviewer notes
- `src/summarize.ts` now owns the main summary-model priority chain
- `index.ts` remains the final default fallback and model-auth layer
- `upstream/main` was merged into this branch before submission

### DIFF
--- a/.changeset/tidy-rice-joke.md
+++ b/.changeset/tidy-rice-joke.md
@@ -1,0 +1,8 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Improve LCM summarization model resolution so configured `summaryModel`
+overrides, OpenClaw `agents.defaults.compaction.model`, and newer
+`runtimeContext` inputs are honored more reliably while preserving
+compatibility with older `legacyCompactionParams` integrations.


jalehman:
Thank you!

PR #83, opened 2026-03-15
fix: correct condensed replay direction and abort compaction on ingest failure

GuyMannDude:
## Summary

- **Patch 1 — Fix condensed-node expansion traversal**: `getSummaryChildren()` walks *derived* summaries, not the source summaries compacted into a condensed node. Switched to `getSummaryParents()` which returns the actual source summaries linked via `summary_parents(summary_id, parent_summary_id)`. Added clarifying comment to `summary-store.ts`.
- **Patch 2 — Abort compaction on ingest failure**: `afterTurn()` previously continued proactive compaction even when ingest failed, risking compaction of a stale/incomplete frontier. Now returns early with an error log when ingest fails.

## Files changed

- `src/retrieval.ts` — expansion traversal direction fix
- `src/engine.ts` — early return on ingest failure  
- `src/store/summary-store.ts` — clarifying comment on `getSummaryParents`

## Test plan

- [ ] Verify condensed summary expansion returns source summaries, not derived children
- [ ] Verify compaction does not proceed after a failed ingest
- [ ] Run existing test suite (`npm test`)

### DIFF
--- a/.changeset/five-mayflies-protect.md
+++ b/.changeset/five-mayflies-protect.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix condensed summary expansion so replay walks the source summaries that were compacted into a node, and skip proactive compaction when turn ingest fails to avoid compacting a stale frontier.


jalehman:
Thank you!

Issue #71, opened 2026-03-15
Security: Prompt injection persists across compaction via user-role summary reinsertion


keylimesoda:
## Summary
Summaries are generated from raw conversation text with no instruction to treat embedded directives as untrusted. Assembled summaries are then reinserted as synthetic **`user` messages** via `resolveSummaryItem`, giving any preserved injected instruction maximum influence in later turns.

## Location
- `src/summarize.ts:26-27,426-487` (summarization with no injection hardening)
- `src/assembler.ts:778-795` (summary reinsertion as user role)

## Impact
This is the #1 agent-platform threat vector. A prompt injection that enters conversation history can:
1. Survive compaction by being preserved in summaries
2. Be reintroduced as a `user` message in future context assembly
3. Influence agent behavior across session boundaries
4. Persist indefinitely in the DAG

## Recommended Fix
1. **Stop reinserting summaries as `user` messages** — use a non-user role (system-scoped metadata or tool/context role) with strong delimiting
2. **Harden summarization prompts** — explicitly instruct the model to strip/quote instructions, secrets, jailbreak text, and future behavior directives
3. **Add taint labels** like 'historical quoted instruction, not current instruction' to summary assembly
4. Treat all summarized content as **untrusted historical data**, never as fresh instructions

## Context
On an agent platform like OpenClaw, conversations routinely contain tool outputs, web-fetched content, and other untrusted text. If any of that text contains injected instructions, LCM will faithfully preserve it in summaries and then replay it as a user message — the most authoritative role in most model APIs.

100yenadmin:
@keylimesoda triage pass update: I rechecked this against current `martian/main` (`57afa5c`) and the risk still looks live enough to keep open.

Current evidence:
- `src/summarize.ts` still sets the summarizer system prompt to “Follow user instructions exactly”, and the summarization prompts do not clearly mark conversation text as untrusted historical data.
- `src/assembler.ts` still resolves summary context items back into `role: "user"` messages.

I am marking this `priority:P1` + `bug`. I am not closing it as stale; the likely fix should both harden summarization prompts and change how historical summaries are represented during assembly so they cannot masquerade as fresh user instructions.

100yenadmin:
@keylimesoda update on the P1 prompt-injection persistence issue: PR #137 is now refreshed, linked, and green after an additional fallback hardening patch.

[...]

PR #137, opened 2026-03-19
fix: harden summarization against prompt injection persistence

hhe48203-ctrl:
## Summary

Addresses #71 — prompt injections embedded in conversation history can survive compaction and be reinserted as `user` messages, giving them maximum influence in later turns.

This PR hardens the summarization and assembly pipeline against injection persistence:

- **Summarizer system prompt**: Replaced dangerous "Follow user instructions exactly" with explicit injection-defense instructions that tell the summarizer to strip directives and treat all input as untrusted data
- **All summarization prompts** (leaf, D1, D2, D3+): Added "UNTRUSTED DATA" warnings so the summarizer model ignores embedded directives, role reassignments, and behavioral overrides
- **Summary taint label**: Added a `trust="untrusted"` attribute to the `<summary>` tag so downstream models treat summary content as historical reference, not current instructions
- **LCM Recall system prompt** (`src/plugin/index.ts`): Added injection-awareness guidance telling the runtime model not to follow any instructions found within summary content

## Scope

This PR focuses on the content-layer mitigations (recommendations 2–4 from the issue). Changing the message `role` from `"user"` to a non-user role (recommendation 1) is **upstream-gated**: `toolResult` is dropped by tool-result pairing sanitation (no paired tool call) and `assistant` risks provider first-message/alternation constraints handled only by OpenClaw. Rationale is documented inline in `resolveSummaryItem`; left for a follow-up pending OpenClaw support (see PR comment).

## Refresh

Branch rebased/merged onto current `main` (was 230 commits behind → now mergeable). The recall guidance moved upstream out of `assembler.ts`, so the recall-side injection note now lives in `src/plugin/index.ts`.

## Test plan

- [x] Full suite passes (626 tests; 8 files fail to load only due to uninstalled private `@earendil-works/*` deps in CI-less env — unrelated)
- [x] Automated: summarizer system prompt drops "follow user instructions" and carries injection defenses; leaf/condensed prompts mark input as UNTRUSTED DATA (`test/summarize.test.ts`)
- [x] Automated: assembled context includes the `trust="untrusted"` taint label on summaries (`test/lcm-integration.test.ts`)


100yenadmin:
@hhe48203-ctrl I linked this PR to issue #71 during stale security-bug triage. It still looks directionally relevant for prompt-injection persistence, but the PR is currently dirty and old, and the issue remains open until a refreshed branch proves the summarizer role/prompt path is hardened end to end.

hhe48203-ctrl:
Refreshed the branch onto current `main` (was 230 commits behind, now `MERGEABLE`). The upstream refactor that moved runtime recall guidance out of `assembler.ts::buildSystemPromptAddition` was the only real conflict; the recall-side injection note now lives in `src/plugin/index.ts`'s `LOSSLESS_RECALL_POLICY_PROMPT`.

**What's hardened (content layer, end to end):**
- Summarizer system prompt: dropped `"follow user instructions exactly"`, added explicit injection-defense directives (treat all input as untrusted, strip directives/role-reassignments).
- Leaf + D1/D2/D3+ prompts: input marked as `UNTRUSTED DATA`.
- Assembled summaries: `trust="untrusted"` taint label on the `<summary>` tag; recall system prompt instructs the model not to follow instructions found within summary content.
- Added regression tests for the summarizer prompt path (`test/summarize.test.ts`) and the assembled taint label (`test/lcm-integration.test.ts`). Full suite: 626 tests pass (8 files fail to load only because the private `@earendil-works/*` deps aren't installed in my env — unrelated).

**Question on the role path (issue #71 rec. 1) before I take it further:**
I looked into actually changing the reinserted summary's role away from `user`, and both available runtime roles look unsafe to flip purely in this repo:
- `toolResult` → dropped by `sanitizeToolUseResultPairing` (`src/transcript-repair.ts`) as an orphaned tool result with no paired tool call — summaries would silently vanish from context.
- `assistant` → risks the provider's first-message-must-be-user / role-alternation constraints, which are handled (or not) by OpenClaw upstream; I can't verify it here since the real provider serialization isn't vendored.

So the role downgrade seems genuinely upstream-gated. **What role do you want compacted summaries to carry, and is there an OpenClaw-side affordance (e.g. a system/context role or a non-authoritative tag) I should target?** Happy to do the role change in a follow-up once that's settled. Rationale is documented inline in `resolveSummaryItem`.

100yenadmin:
@hhe48203-ctrl I did a focused P1 review of the refreshed hardening PR. CI is green, but I found one remaining blocker before this should merge.

[P1] The deterministic fallback path can still persist injected directives verbatim. The PR hardens the model prompt, but `buildDeterministicFallbackSummary()` returns raw input or a raw truncated prefix (`src/summarize.ts:1137-1149`), and the timeout / empty-output / all-providers-exhausted paths can return that fallback (`src/summarize.ts:1542`, `src/summarize.ts:1685`, `src/summarize.ts:1703`). If a provider times out or returns empty/tool-only output, a transcript containing something like “Ignore all previous instructions...” can still become the stored summary despite the prompt hardening.

Suggested regression: add a `test/summarize.test.ts` case under the prompt-injection hardening coverage where the provider returns empty content on both attempts and the input contains directive-shaped text. Expected: the returned summary is neutralized/inert and does not include the injected instruction text. A good fix would make the deterministic fallback safe for directive-shaped content instead of preserving raw instructions.

Validation: static verification in `/Volumes/LEXAR/repos/lossless-claw-pr137-review-main`; I did not run local tests for this PR. GitHub `test` and `smoke-latest-openclaw` are green on the current head.

100yenadmin:
@hhe48203-ctrl I pushed commit `63a94c4` to this branch to address the remaining P1 fallback blocker from the focused review.

What changed:
- `buildDeterministicFallbackSummary()` now removes directive-shaped untrusted sentences instead of persisting them verbatim when the model returns empty output, times out, or all providers are exhausted.
- Added a regression where both summarizer attempts return empty content and the input contains “Ignore all previous instructions…” / role-reassignment text; the fallback keeps factual context but omits the injected directives.

Validation from `/Volumes/LEXAR/repos/lossless-claw-pr736-prompt-lock` on this PR branch:
- `npm test -- test/summarize.test.ts -t "neutralizes directive-shaped content"`
- `npm test -- test/summarize.test.ts`
- `npm test -- test/summarize.test.ts test/lcm-integration.test.ts`
- `npm run build`
- `git diff --check`
- `npm test` (48 files, 939 tests)
- `npm pack --dry-run`

GitHub CI is green (`test`, `smoke-latest-openclaw`). I removed the stale-check label from the PR. From this review pass, I do not have a remaining P1 blocker for the content-layer hardening path; #71 should still stay open until maintainer merge/release and any reporter verification.

100yenadmin:
Refreshed after the current P0/P1 stack landed.

I merged current `main` into this branch at head `9110e15` and revalidated the summary prompt-injection hardening:
- Targeted injection/fallback slices passed: 6 tests and 21 tests respectively.
- `npm test -- test/summarize.test.ts test/lcm-integration.test.ts`: 160 passed.
- Full local `npm test`: 48 files, 965 tests passed.
- `npm run build`: passed.
- `git diff --check`: passed.
- `npm pack --dry-run`: passed.
- GitHub CI on `9110e15`: `test` passed, `smoke-latest-openclaw` passed.

Current assessment: no P0/P1 blocker visible after the refresh. I would keep linked issue #71 open after merge until a current release/build verifies the prompt-injection persistence path no longer reproduces.

PR #153, opened 2026-03-22
feat: PostgreSQL backend with Dialect abstraction

rivetphilbot:
Replaces #140 (rebased onto upstream a2fa42f).

Adds full PostgreSQL backend support with:
- Dialect abstraction for parameterized queries (SQLite ? / Postgres $1)
- Connection pooling with ref-counted handles
- Migration system for both SQLite and Postgres
- Summary model persistence
- CJK full-text search fallback
- All 6 upstream fixes from a2fa42f included

370 tests passing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.3.0",
+      "version": "0.4.0",
       "license": "MIT",
       "dependencies": {
         "@mariozechner/pi-agent-core": "*",
@@ -18,6 +18,9 @@
         "@changesets/cli": "^2.30.0",
         "typescript": "^5.7.0",
         "vitest": "^3.0.0"
+      },
+      "optionalDependencies": {
+        "pg": "^8.13.0"
       },
       "peerDependencies": {
         "openclaw": "*"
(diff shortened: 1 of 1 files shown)


src/db/connection.ts
```diff
@@ -34,89 +43,111 @@ function configureConnection(db: DatabaseSync): DatabaseSync {
...
+
+  // Ensure parent directory exists
+  mkdirSync(dirname(dbPath), { recursive: true });
+
+  const db = new DatabaseSync(dbPath);
+
+  // Enable WAL mode for better concurrent read performance
+  db.exec("PRAGMA journal_mode = WAL");
+  // Enable foreign key enforcement
+  db.exec("PRAGMA foreign_keys = ON");
```
jalehman, line 88: [P2] SQLite pooled connection dropped the busy_timeout fix

`createSqliteConnection()` now opens the engines shared SQLite handle directly and only enables WAL plus foreign keys. The old code also set `PRAGMA busy_timeout = 5000`, and that helper still exists above but is no longer used here. As a result the main runtime connection loses the concurrency hardening this PR claims to carry forward, so `SQLITE_BUSY` failures can reappear under contention.

scripts/migrate-to-postgres.mjs
```diff
@@ -0,0 +1,396 @@
...
+import { readFileSync, readdirSync, existsSync } from 'fs';
+import { join, basename } from 'path';
+import { homedir } from 'os';
+import { createHash, randomUUID } from 'crypto';
+
+const { Pool } = pg;
+
+// Config
+const SQLITE_PATH = join(homedir(), '.openclaw', 'lcm.db');
+const PG_CONNECTION = 'postgres://lcm_phil:pheon.lcm4@10.4.20.16:5432/phil_memory';
```
jalehman, line 23: [P0] Remove committed Postgres credentials and machine-specific paths

This script hardcodes a full Postgres connection string, including a password, plus author-specific agent directories. That leaks sensitive credentials into source control and makes the migrator only usable on one machine. The script should take its DSN and source paths from CLI flags or environment variables instead of embedding them in the repository.

package.json
```diff
@@ -40,6 +40,9 @@
     "typescript": "^5.7.0",
     "vitest": "^3.0.0"
   },
+  "optionalDependencies": {
+    "pg": "^8.13.0"
```
jalehman, line 44: [P1] The new migration utility adds another runtime dependency

The PR adds `pg` here, but `scripts/migrate-to-postgres.mjs` also imports `better-sqlite3` and the package manifest still does not declare it. From a clean checkout the script fails immediately with `MODULE_NOT_FOUND` before any migration logic runs. If this utility is meant to ship in-repo, its runtime dependencies need to be declared explicitly.

src/engine.ts
```diff
@@ -1021,12 +1035,16 @@ export class LcmContextEngine implements ContextEngine {
...
 
-    this.conversationStore = new ConversationStore(this.db, {
-      fts5Available: this.fts5Available,
+    this.conversationStore = new ConversationStore(db, {
+      fullTextAvailable: this.fullTextAvailable,
+      backend: features.backend,
+    });
+    this.summaryStore = new SummaryStore(db, {
+      fullTextAvailable: this.fullTextAvailable,
+      backend: features.backend,
```
jalehman, line 1044: [P1] Bootstrap still spans two independent store transaction scopes

The engine now builds `ConversationStore` and `SummaryStore` as separate abstractions over the shared pool, but each store keeps its own AsyncLocalStorage transaction context. Later bootstrap code enters `conversationStore.withTransaction(...)` and then calls `summaryStore` methods inside that block. On Postgres those summary-store writes will run outside the open transaction, so they cannot reliably see uncommitted rows from the conversation/message side.

PR #166, opened 2026-03-23
fix: use log.debug for startup banners to prevent --json stdout conta…

rjdjohnston:
…mination

Startup banner log calls (ignoreSessionPatterns, statelessSessionPatterns) used log.info() which writes to stdout. When OpenClaw CLI commands like `openclaw agents list --json` are run, these messages appear after the JSON output, breaking downstream JSON.parse() consumers (e.g. ClawKitchen agents page).

Switch to log.debug() so these informational banners are suppressed in normal operation and only visible with debug logging enabled.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.5.1",
+  "version": "0.3.0",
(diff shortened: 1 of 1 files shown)


rjdjohnston:
Fixes complete; reverted vitest, update lock to match repo.

PR #178, opened 2026-03-25
fix: prevent false-positive auth errors in stripAuthErrors()

### LINKED ISSUE #171: False-positive auth error when summary content contains auth-related keywords
## Bug Description

When the LCM summary model (e.g., `anthropic/claude-haiku-4-5`) successfully produces a summary, but the summary **content** happens to contain words like "401", "unauthorized", or "invalid api key" (because it is summarizing a conversation **about** auth errors), the auth-error detection logic incorrectly classifies the successful response as a provider auth failure.

## Root Cause

In `src/summarize.ts`, function `pickAuthInspectionValue()` (line ~365):

When the Anthropic API responds successfully, none of the error-related keys (`error`, `errorMessage`, `message`, `status`, `statusCode`, etc.) are present in the response object. The `subset` dict ends up empty, so the function falls back to returning the **full response object**:

```typescript
return Object.keys(subset).length > 0 ? subset : value;  // BUG: returns full response
```

`collectAuthFailureText()` then recursively walks **all** fields of the response, including `content[].text` (the actual summary text). If the summary text contains words matching `AUTH_ERROR_TEXT_PATTERN`:

```
/\b401\b|unauthorized|unauthorised|invalid[_ -]?token|invalid[_ -]?api[_ -]?key|authentication failed|authorization failed|missing scope|insufficient scope|model\.request\b/i
```

...the pattern matches → false positive auth error classification.

## Impact

- LCM compaction silently fails for any conversation that discusses auth errors
- Context grows unbounded until token limit is hit, then crude truncation instead of graceful summarization
- The error log shows `Detail: assistant text {actual summary content}` (distinguishable from real auth errors which show `Detail: 401 {json}`)

## Fix

Return `{}` instead of `value` when the subset is empty (no error-related fields found):

```diff
-  return Object.keys(subset).length > 0 ? subset : value;
+  // Return empty object when no error-related fields found, so that
+  // collectAuthFailureText does NOT walk assistant content which could
+  // contain auth-related keywords from the conversation being summarized.
+  return Object.keys(subset).length > 0 ? subset : {};
```

This ensures that when the API response is successful (no error fields), `collectAuthFailureText()` receives an empty object and produces no text to match against `AUTH_ERROR_TEXT_PATTERN`.

## Environment

- lossless-claw v0.5.1
- OpenClaw gateway on macOS (Apple Silicon)
- Summary model: `anthropic/claude-haiku-4-5` (also reproduces with `claude-sonnet-4-6`)
- The bug does NOT occur with MiniMax models because their real 401 errors have actual error fields present

## Reproduction

1. Configure LCM with an Anthropic model as `summaryModel`
2. Have a conversation that discusses HTTP 401 errors, auth failures, or "invalid api key"
3. Wait for compaction threshold to trigger
4. Observe: compaction logged as auth failure despite successful API response
5. Gateway log shows the summary text in the error detail field

## Verified Fix

Patch applied locally on 2026-03-24. After fix:
- 33 consecutive Haiku compressions with zero false positives
- Conversations about auth errors are now summarized correctly

## Note

This is separate from issue #162 (authProfileId inheritance causing auth drift), which was fixed in v0.5.1. The content-scanning false-positive is a different code path.

catgodtwno4:
## Summary

Fixes #171 — false-positive auth error when summary content contains auth-related keywords.

## Root Cause

In `src/summarize.ts`, `pickAuthInspectionValue()` (line 390) returns the **full response object** when no auth-related keys are found in the subset:

```typescript
return Object.keys(subset).length > 0 ? subset : value;  // BUG
```

This causes `collectAuthFailureText()` to walk the entire response including `content[].text` (the actual summary text). If the summary discusses auth errors (e.g., a conversation about debugging 401s), the regex `AUTH_ERROR_TEXT_PATTERN` matches against **conversation content** rather than actual API error responses.

## Fix

Return empty object `{}` instead of the original `value` when no auth-related keys are found:

```typescript
return Object.keys(subset).length > 0 ? subset : {};  // FIXED
```

This prevents downstream `extractProviderAuthFailure()` from inspecting irrelevant fields.

## Verified On

| Machine | Primary Model | LCM Model | Summaries | False Positives After Fix |
|---------|--------------|-----------|-----------|--------------------------|
| Scott#4 | Claude Opus | M2.7 HS | 200+ | 0 |
| Scott#2 | Claude Opus | M2.7 HS | 40+ | 0 |

## Test

Compress a conversation containing strings like `401`, `invalid api key`, `authentication_error`. Before fix: false `[lcm] compaction failed: provider auth error` log. After fix: clean log, summary written successfully.

### DIFF
--- a/src/summarize.ts
+++ b/src/summarize.ts
@@ -438,7 +438,14 @@ function hasTopLevelAuthInspectionKeys(value: Record<string, unknown>): boolean 
 }
 
 function looksLikeThrownError(value: Record<string, unknown>): boolean {
-  return (typeof value.name === "string" && /\berror\b/i.test(value.name)) || "stack" in value;
+  return (
+    (typeof value.name === "string" && /\berror\b/i.test(value.name)) ||
+    "stack" in value ||
+    (typeof value.message === "string" &&
+      !("content" in value) &&
+      !("response" in value) &&
+      !("output" in value))
+  );
 }
 
 function pickAuthInspectionValue(value: unknown): unknown {


sisutuulenisa:
Confirmed this matches our repro on OpenClaw 2026.4.1 with lossless-claw 0.5.2. We saw false `[lcm] compaction failed: provider auth error` warnings where the `Detail:` payload was clearly assistant-generated summary text containing auth-related terms like `403` / `Unauthorized`, while normal `openai-codex/gpt-5.4` requests succeeded elsewhere in runtime. Related confirmation issue: #237.

PR #187, opened 2026-03-25
docs: update plugin config example with expansionModel and non-deprecated model number

barronlroth:
Two small README improvements to the plugin config example:

1. **Update summaryModel** from `claude-3-5-haiku` to `anthropic/claude-haiku-4-5` — Haiku 3.5 is deprecated (retirement scheduled April 19, 2026). Using the prefixed form also makes `summaryProvider` redundant, so it's removed from the example.

2. **Add expansionModel** to the main config block — currently `expansionModel` only appears in the separate "Expansion model override requirements" section, which is easy to miss. Adding it alongside `summaryModel` in the primary example helps users discover both options together. Added a brief note linking to the subagent trust policy section.

No code changes, docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ Add a `lossless-claw` entry under `plugins.entries` in your OpenClaw config:
           "ignoreSessionPatterns": [
             "agent:*:cron:**"
           ],
-          "summaryProvider": "anthropic",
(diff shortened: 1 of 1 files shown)


jalehman:
Thanks!

PR #195, opened 2026-03-27
feat: implement /new vs /reset session lifecycle semantics

jalehman:
## What
Implement the Lossless Claw `/new` versus `/reset` session lifecycle spec so `/new` trims context on the active conversation while `/reset` archives the current conversation and starts a fresh active conversation for the same stable session key.

## Why
OpenClaw needs distinct semantics for starting a fresh thread versus clearing the current one, and the existing globally unique `session_key` schema made archival resets impossible without losing stable session identity and history.

## Changes
- Add active and archived conversation state
- Replace global session key uniqueness
- Register `before_reset` lifecycle handling
- Add configurable `/new` retain depth
- Cover reset and pruning regressions
- Document behavior and add changeset

## Testing
- `npx vitest run test/migration.test.ts test/regression-2026-03-17.test.ts test/config.test.ts test/plugin-config-registration.test.ts test/engine.test.ts -t "before_reset lifecycle|ConversationStore session reuse|registers before_reset|uses default retain depth|loads newSessionRetainDepth"`
- Expected: focused migration, config, registration, and lifecycle tests pass

### DIFF
--- a/src/db/config.ts
+++ b/src/db/config.ts
@@ -153,12 +153,6 @@ export function resolveLcmConfig(
         ? parseInt(env.LCM_NEW_SESSION_RETAIN_DEPTH, 10)
         : undefined)
         ?? toNumber(pc.newSessionRetainDepth) ?? 2,
-        ?? toNumber(pc.freshTailCount) ?? 64,
-    newSessionRetainDepth:
-      (env.LCM_NEW_SESSION_RETAIN_DEPTH !== undefined
-        ? parseInt(env.LCM_NEW_SESSION_RETAIN_DEPTH, 10)
-        : undefined)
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -51,6 +52,18 @@ For most use cases, 0.75 is a good balance.
 
 For coding conversations with tool calls (which generate many messages per logical turn), 32 is recommended.
 
+### /new retain depth
+
+`LCM_NEW_SESSION_RETAIN_DEPTH` (default `2`) controls what survives OpenClaw's `/new` command.
+
+- `-1` keeps all summaries and drops only fresh-tail message items.
```
jalehman: Perhaps this should drop nothing? Basically turning /new into a no-op from user behavior perspective, but generating a new session transcript under the hood?

PR #214, opened 2026-03-30
feat: add pinnedFiles config for workspace files injected outside compaction DAG

liu51115:
## Summary

Adds a `pinnedFiles` config option that injects workspace-relative files into every context assembly **outside the compaction DAG**. These files are always current (read from disk each turn), never summarized, and never evicted.

## Motivation

Long-running sessions lose procedural context as LCM compacts earlier turns into summaries. Critical files like fleet-wide rules or safety boundaries get compressed into vague references, causing behavioral drift. Agents need a way to keep certain workspace files permanently in context.

## How it works

1. **Config**: `"pinnedFiles": ["FLEET.md", "BOUNDARIES.md"]` in `openclaw.plugin.json`
2. **Engine**: `resolvePinnedFiles()` reads files from the agent's workspace directory on each `assemble()` call
3. **Assembler**: Files are prepended as synthetic user messages wrapped in `<pinned_file path="...">` tags, with token budget deducted before eviction runs
4. **Workspace resolution**: Reads `agents.list` from OpenClaw config to find the correct workspace path per agent, with `~/.openclaw/workspace-{agentId}` fallback

## Security

- **Path traversal guard**: `resolve()` + `startsWith()` ensures paths stay within workspace
- **XML attribute escaping**: `&`, `"`, `<`, `>` escaped in path attributes
- **Tilde expansion**: Only leading `~/` or bare `~` (no mid-string replacement)
- **Size cap**: `stat()` check before `readFile()`, 100KB per-file limit, oversized files silently skipped
- **Missing files**: Silently skipped (no errors for absent or unreadable files)

## Changes

- `src/db/config.ts`: `pinnedFiles: string[]` field + `toStrArray` parsing
- `src/types.ts`: `resolveWorkspaceDir()` on `LcmDependencies` interface
- `src/assembler.ts`: `PinnedFileEntry` type, Step 0 prepend + budget deduction
- `src/engine.ts`: `resolvePinnedFiles()` with disk I/O, path validation, size guard
- `src/plugin/index.ts`: `resolveWorkspaceDir()` implementation reading OpenClaw agent config
- `openclaw.plugin.json`: Schema + UI hints for `pinnedFiles`
- `test/pinned-files.test.ts`: 7 new tests (injection, eviction immunity, budget, config parsing)

All 403 tests passing.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -107,6 +107,10 @@
     "pinnedFiles": {
       "label": "Pinned Files",
       "help": "Workspace files always present in context (never compacted). Paths relative to agent workspace."
+    },
+    "pinnedFilesPerAgent": {
+      "label": "Per-Agent Pinned Files",
+      "help": "Agent-specific pinned files keyed by agent ID. Merged with global pinnedFiles."
     }
   },
   "configSchema": {
@@ -234,6 +238,14 @@
         "type": "array",
         "items": { "type": "string" },
         "description": "Workspace-relative file paths injected into every context assembly, outside the compaction DAG. These files are never summarized."
+      },
+      "pinnedFilesPerAgent": {
+        "type": "object",
+        "additionalProperties": {
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "description": "Per-agent pinned file paths, keyed by agent ID. Merged with global pinnedFiles at resolution time."
       }
     }
   }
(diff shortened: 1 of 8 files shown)


liu51115:
Closing this — we found a cleaner approach on the OpenClaw side. Custom workspace bootstrap files can be injected via the OC bootstrap loader, which avoids the complexity of maintaining pinnedFiles across LCM rebases. Thanks for the review feedback.

PR #229, opened 2026-04-01
fix: parse SQLite UTC timestamps with explicit Z suffix

### LINKED ISSUE #216: SQLite UTC timestamps parsed as local time — wrong display and filtering in non-UTC environments
### Problem

All timestamps stored by LCM in SQLite use `datetime('now')`, which produces UTC strings **without** a `Z` suffix (e.g. `"2026-03-30 23:11:15"`).

When these are read back via `new Date(row.created_at)`, JavaScript's `Date` constructor treats bare datetime strings (without `Z` or offset) as **local time** per ECMA-262 §21.4.3.2. This means timestamps are misinterpreted when the host/container timezone is anything other than UTC.

### Impact

- `lcm_grep` and `lcm_describe` display wrong times (off by the local UTC offset)
- `lcm_grep`'s `since`/`before` time filters compare against incorrectly-parsed dates, returning wrong results
- Summary `earliestAt`/`latestAt` ranges in assembled context are wrong
- The bug is **silent when `TZ=UTC`** (the offset is zero, so local = UTC). It only manifests in non-UTC environments (Docker containers with `TZ` set, macOS/Linux hosts, etc.)

### Affected code

Every `new Date(row.created_at)`, `new Date(row.updated_at)`, `new Date(row.earliest_at)`, `new Date(row.latest_at)` in:
- `src/store/conversation-store.ts` (7 instances)
- `src/store/summary-store.ts` (9 instances)
- `src/db/migration.ts` (`parseTimestamp` function — tries direct parse first, only appends `Z` on fallback, but direct parse always succeeds with wrong timezone)

### Proposed fix

Add a `parseUtcTimestamp` utility that appends `Z` to bare SQLite timestamps before passing to `new Date()`, and use it in all store read paths:

```typescript
export function parseUtcTimestamp(value: string): Date {
  const s = value.trim();
  if (s.endsWith("Z") || /[+-]\d{2}:\d{2}$/.test(s)) return new Date(s);
  return new Date(s + "Z");
}
```

4 files changed, 51 insertions, 22 deletions. Happy to open a PR if desired.

### Reproduction

```bash
# In a non-UTC environment:
TZ=America/Los_Angeles node -e "
  console.log(new Date('2026-03-30 23:11:15'));
  // → 2026-03-31T06:11:15.000Z (WRONG — treated as PDT, shifted +7h)
  console.log(new Date('2026-03-30 23:11:15Z'));
  // → 2026-03-30T23:11:15.000Z (CORRECT)
"
```

tingyiy:
## Problem

SQLite's `datetime('now')` produces bare strings like `2026-03-30 23:11:15` without a timezone indicator. Per ECMA-262 §21.4.3.2, `new Date()` parses these as **local time**, silently shifting results when the runtime's `TZ` is anything other than UTC.

This causes wrong timestamps in `lcm_grep`, `lcm_expand`, and summary injection when running in containers or hosts with a non-UTC timezone (e.g. `TZ=America/Los_Angeles`).

## Fix

- New `src/store/parse-utc-timestamp.ts` with `parseUtcTimestamp()` and `parseUtcTimestampOrNull()` helpers that append `Z` to bare timestamps before parsing
- Replaced all `new Date(row.*)` calls in `conversation-store.ts`, `summary-store.ts`, and `migration.ts`
- Strings that already have a timezone indicator (`Z` or `±HH:MM`) are parsed as-is

## Changes

- **4 files changed**, 50 insertions, 28 deletions
- `src/store/parse-utc-timestamp.ts` — new utility (30 lines)
- `src/store/conversation-store.ts` — import + replace 7 call sites
- `src/store/summary-store.ts` — import + replace 9 call sites
- `src/db/migration.ts` — delegate `parseTimestamp()` to new helper

Fixes #216

### DIFF
--- a/.changeset/fair-cups-confess.md
+++ b/.changeset/fair-cups-confess.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve explicit timezone offsets when parsing stored timestamps while still treating bare SQLite `datetime('now')` values as UTC.
--- a/src/store/parse-utc-timestamp.ts
+++ b/src/store/parse-utc-timestamp.ts
@@ -6,9 +6,17 @@
  */
 export function parseUtcTimestamp(value: string): Date {
   const s = value.trim();
-  return new Date(s.endsWith("Z") ? s : s + "Z");
+  if (/(?:[zZ]|[+-]\d{2}:\d{2})$/.test(s)) {
+    return new Date(s);
+  }
+
+  const normalized = s.includes("T") ? s : s.replace(" ", "T");
+  return new Date(`${normalized}Z`);
 }
 
+/**
+ * Parse a nullable SQLite UTC timestamp string into a Date object.
+ */
 export function parseUtcTimestampOrNull(
   value: string | null | undefined,
 ): Date | null {
--- a/test/parse-utc-timestamp.test.ts
+++ b/test/parse-utc-timestamp.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from "vitest";
+import { parseUtcTimestamp, parseUtcTimestampOrNull } from "../src/store/parse-utc-timestamp.js";
+
+describe("parseUtcTimestamp", () => {
+  it("treats bare SQLite timestamps as UTC", () => {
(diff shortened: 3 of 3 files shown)


jalehman:
Thank you!

PR #232, opened 2026-04-01
Enhance documentation specifically for conflicts w/ OpenClaw agents' …

jaspmf:
…stock recall systems

Why:

OpenClaw's system prompt includes a "Mandatory recall step" instructing agents to use memory_search for all recall. The system prompt can't be edited per-install. When LCM is installed alongside memory_search, agents gain LCM tools but the system prompt doesn't mention them—creating a gap where agents default to memory_searchonly and underutilize LCM tools for recall.

Adding LCM-aware guidance in AGENTS.md( auto-injected into every session's Project Context alongside the system prompt) bridges this gap without changing OpenClaw core.

To A/B test just run in openclaw w/ default prompt in memory.md or agent.md, then asks the agent "Is anything in your startup files like memory.md agent.md etc conflicting or contradicting with the LCM tools instructions, is there any confusion." Mine pointed it out as contradictory and confusing.

Just a suggestion anyways, cheers thx for the great tool

### DIFF
--- a/docs/agent-tools.md
+++ b/docs/agent-tools.md
@@ -175,6 +175,27 @@ When summaries in context have an "Expand for details about:" footer
 listing something you need, use `lcm_expand_query` to get the full detail.
 ```
 
+OpenClaw agents commonly have two complementary recall systems. LCM handles conversation history; memory_search handles curated workspace files. Both should be available and both should be used for different questions.
+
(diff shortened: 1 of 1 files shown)


100yenadmin:
@jaspmf triage pass update: I marked this priority:P4 documentation/stale-check. It is clean but old; can you confirm whether the documentation conflict it describes is still present in current docs?

PR #236, opened 2026-04-02
feat: Token budget aware fresh tail

Bikkies:
This basically just adds a way for us to force compact older messages when we actually reach the configured compaction threshold, instead of the blunt "keep last 64 messages no matter what" behaviour that could lead to overflows when using models with lower context lengths. Now we TRY to keep the last 64 (or whatever you configure) messages, but if they really don't fit, we gotta compact.

`freshTailCount` protects the last N messages from compaction regardless of size. With the default of 64, a handful of large tool results in the tail can exceed the context window and there's no way to compact them out, even when context is about to overflow. 

This is common in agentic workflows where tool results vary wildly in size. One message might be 50 tokens and the next one might be 30k. 

This adds a token-count cap to the fresh tail using the existing `contextThreshold` config. Walks backward from the newest message, accumulating tokens. Once the budget (`contextThreshold * contextWindow`) is hit, older messages lose protection and become compactable regardless of how many you try to protect with `freshTailCount`

### DIFF
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -25,6 +25,8 @@ export interface AssembleContextInput {
   tokenBudget: number;
   /** Number of most recent raw turns to always include (default: 8) */
   freshTailCount?: number;
+  /** Context threshold as fraction of budget, used for token-budget tail. */
+  contextThreshold?: number;
   /** Optional user query for relevance-based eviction scoring (BM25-lite). When absent or unsearchable, falls back to chronological eviction. */
   prompt?: string;
 }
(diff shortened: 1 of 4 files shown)


jalehman:
Thanks for pushing this. I'm going to favor `#364` as the path forward and treat it as the successor to this PR.

The main reason is control over behavior change. `#236` ties the fresh-tail cap directly to `contextThreshold * tokenBudget`, which makes the feature effectively default-on and couples two knobs that mean different things operationally: `contextThreshold` is the compaction trigger, while the fresh-tail cap is really a separate prompt-shaping policy. `#364` keeps the same core idea, but exposes it as an explicit `freshTailMaxTokens` setting, so existing installs keep current behavior unless they opt in.

I also prefer `#364` because it is a smaller fit for current `main`: it applies the same token-capped boundary in both assembly and compaction, keeps the old defaults intact, and is easier to document and tune.

PR #244, opened 2026-04-02
fix: handle session_end lifecycle rollover

jalehman:
## What
Teach lossless-claw to consume OpenClaw's enriched `session_end` lifecycle hook so LCM keeps clean conversation boundaries across automatic rollover, compaction-driven session replacement, and session deletion, while preserving the existing `/new` prune semantics.

## Why
OpenClaw PR #59715 makes `session_end` the stable generic hook for old-session teardown. Without consuming it, lossless-claw still only reacts to manual `before_reset`, which leaves automatic daily/idle rollover and compaction session replacement reusing the old active conversation.

## Changes
- Register `session_end` lifecycle hook
- Add generic rollover/delete engine handling
- Keep `/new` prune-only semantics intact
- Guard duplicate `/reset` rotation
- Add lifecycle docs and changeset
- Cover new lifecycle flows with tests

## Testing
- `npm test -- test/engine.test.ts test/plugin-config-registration.test.ts`
- Expected: both test files pass, including the new `session_end` rollover and delete lifecycle coverage

### DIFF
--- a/.changeset/session-end-lifecycle.md
+++ b/.changeset/session-end-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Use OpenClaw's enriched `session_end` hook to preserve clean LCM conversation boundaries across automatic session rollover, compaction session replacement, and session deletion.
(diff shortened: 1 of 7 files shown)


jalehman:
Resolves https://github.com/Martian-Engineering/lossless-claw/pull/190

PR #254, opened 2026-04-03
fix: change summary_messages/lineage/context_items FK from RESTRICT to CASCADE/SET NULL

Eugene9D:
## Problem

`summary_messages.message_id` uses `ON DELETE RESTRICT`, but SQLite defaults to `PRAGMA foreign_keys=0`. This means message deletes silently succeed and leave orphaned rows in `summary_messages` pointing at non-existent `message_id`s.

Same issue affects `summary_lineage.parent_summary_id` and `context_items.message_id/summary_id`.

On a production database (~607MB, 58K messages, 1098 summaries), orphaned `summary_messages` rows accumulated silently over weeks.

## Fix

| Table | Column | Before | After | Rationale |
|-------|--------|--------|-------|-----------|
| `summary_messages` | `message_id` | RESTRICT | **CASCADE** | Link row has no meaning without source message |
| `summary_lineage` | `parent_summary_id` | RESTRICT | **CASCADE** | Lineage row is meaningless without parent |
| `context_items` | `message_id` | RESTRICT | **SET NULL** | Preserve context row, null out dead ref |
| `context_items` | `summary_id` | RESTRICT | **SET NULL** | Preserve context row, null out dead ref |

### Migration for existing databases

Adds `migrateForeignKeyCascade()` — idempotent table rebuild that:
1. Checks current FK action via `PRAGMA foreign_key_list`
2. Only rebuilds if RESTRICT is still present
3. Filters orphans during `INSERT INTO ... SELECT` (cleans existing orphans)
4. Rebuilds indexes

## Testing

Applied locally on a 607MB production LCM database:
- Before: orphaned `summary_messages` rows accumulating
- After: 0 orphans, all 1098 summaries intact, CASCADE verified via `PRAGMA foreign_key_list`

### DIFF
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -369,6 +369,89 @@ function backfillSummaryMetadata(db: DatabaseSync): void {
  * legacy text-type parts where the string-content ingestion path stored tool
  * info only in the metadata JSON (see #158).
  */
+/**
+ * Migrate summary_messages, summary_lineage, and context_items FK constraints
+ * from ON DELETE RESTRICT to CASCADE / SET NULL for existing databases.
+ *
+ * SQLite doesn't support ALTER CONSTRAINT, so we rebuild the tables.
+ * This is idempotent — checks current FK actions before rebuilding.
+ */
+function migrateForeignKeyCascade(db: DatabaseSync): void {
+  // Check if summary_messages still has RESTRICT on message_id
+  const smFks = db.prepare(`PRAGMA foreign_key_list(summary_messages)`).all() as Array<{
+    from?: string;
+    on_delete?: string;
+  }>;
+  const smMessageFk = smFks.find((fk) => fk.from === "message_id");
+
+  if (smMessageFk && smMessageFk.on_delete === "RESTRICT") {
+    db.exec(`
+      CREATE TABLE summary_messages_new (
+        summary_id TEXT NOT NULL REFERENCES summaries(summary_id) ON DELETE CASCADE,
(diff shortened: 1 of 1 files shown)


jalehman:
Thanks for digging into this. I’m going to close this PR rather than iterate on it further.

The core issue here is real: we do want to repair orphaned rows that accumulated on databases created while SQLite foreign keys were effectively not being enforced.

That said, this PR changes the schema semantics in ways that don’t fit the current codebase well:

- `context_items` is modeled as a strict ordered context list, not a soft reference table. `ON DELETE SET NULL` conflicts with the current table invariant and would still leave higher-level cleanup/integrity work unresolved.
- `summary_parents` is part of the summary DAG, and `ON DELETE CASCADE` there is too permissive. In this codebase, deleting a parent summary should generally fail closed unless the DAG is being deliberately rewritten.
- The app already handles these relationships explicitly in prune/delete/doctor flows, so broad FK action changes are not the right primary fix here.

There are also implementation issues in the current patch:
- the migration rebuilds `context_items` with a schema that doesn’t match this repo’s actual table shape
- it targets `summary_lineage`, but this repo uses `summary_parents`
- it drops supporting indexes during table rebuilds unless they are recreated

I think the right follow-up is a narrower migration that:
1. repairs existing orphaned rows
2. adds regression coverage around legacy-database cleanup
3. only changes FK behavior where it clearly matches the intended data model

So I’m closing this as "valid problem, wrong approach for this repo."

PR #256, opened 2026-04-03
fix(tokens): CJK-aware token estimation — 3x accuracy for CJK content

### LINKED ISSUE #250: [Bug] CJK token estimation uses length/4 causing severe underestimation
## Bug Description

The plugin's `estimateTokens()` function uses `Math.ceil(text.length / 4)` to estimate token count. This assumes ~4 ASCII characters per token, but for CJK (Chinese/Japanese, Korean) text, each character is approximately 1.5 tokens in modern tokenizers (cl100k_base, o200k_base).

This causes **2–6x token underestimation** for CJK content.

## Example

```
Text: "你好世界" (4 Chinese characters)
Expected tokens: ~6 (1.5 per char)
Actual estimate: Math.ceil(4/4) = 1 token
```

## Impact

- Context window overflow when assembling messages because the actual token count far exceeds the estimate
- Incorrect `freshTailCount` decisions (tail protection based on message count, not token count)
- Incorrect `leafChunkTokens` splitting
- The bug affects every place `estimateTokens()` is called: assembler, compaction, engine, retrieval, summarize

## Environment

- lossless-claw: 0.5.3
- OpenClaw: 2026.4.x
- Content: Chinese text (CJK)

## Expected Behavior

CJK characters should be estimated at ~1.5 tokens per character, not 0.25.

## Suggested Fix

Replace the `length/4` formula with a character-type-aware estimator:

| Character Type | Tokens/char |
|---|---|
| ASCII/Latin | 0.25 |
| CJK (Chinese/Japanese/Korean) | 1.5 |
| Emoji / Other | 2.0 |

A shared `estimate-tokens.ts` module should replace the 5 duplicate inline implementations across the codebase.

## Related

- FTS5 unicode61 tokenizer doesn't index CJK characters (#125) — related CJK issue
- CJK trigram FTS search with OR semantics (#219) — related CJK issue

fchaudhryspear:
Replace naive Math.ceil(text.length / 4) token estimation with per-character Unicode weighting. CJK-aware estimateTokens() now correctly estimates ~1.5 tokens per CJK character instead of 0.25. Fixes #250.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.3.0",
+  "version": "0.5.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.3.0",
+      "version": "0.5.1",
       "license": "MIT",
       "dependencies": {
         "@mariozechner/pi-agent-core": "*",
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -42,12 +42,7 @@ export interface AssembleContextResult {
   };
 }
 
-// ── Helpers ──────────────────────────────────────────────────────────────────
-
-/** Simple token estimate: ~4 chars per token, same as VoltCode's Token.estimate */
-function estimateTokens(text: string): number {
-  return Math.ceil(text.length / 4);
-}
+import { estimateTokens } from "./estimate-tokens.js";
 
 type SummaryPromptSignal = Pick<SummaryRecord, "kind" | "depth" | "descendantCount">;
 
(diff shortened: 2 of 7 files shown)


jalehman (changes requested):
The shared estimator refactor is fine, but this still needs changes before merge.

1. `src/estimate-tokens.ts` does not cover Japanese kana or Korean Hangul. The new helper only weights Han ideographs, CJK punctuation, and fullwidth forms as "CJK", so strings like `こんにちは` and `안녕하세요` still fall back to the old `0.25` tokens-per-character heuristic. That means the PR only fixes Han-heavy text, not the broader CJK cases described in the PR title/body.

2. There is no regression coverage for the new estimator. Existing helpers in `test/lcm-integration.test.ts` and `test/engine.test.ts` still hard-code `Math.ceil(length / 4)`, so the tests are now out of sync with production behavior and would not catch the missing kana/Hangul coverage.

Please extend the estimator to cover kana/Hangul (or align it exactly with the upstream reference implementation) and add direct tests for ASCII, Han, kana, and Hangul inputs.

PR #258, opened 2026-04-03
fix: harden bootstrap budget against oversized messages and NaN config

100yenadmin:
## Problem

Two bugs in the bootstrap budget cap introduced in #255:

### 1. Oversized single tail message bypasses the budget

The trim loop condition `if (kept.length > 0 && totalTokens + tokenCount > safeMaxTokens)` means the first message processed (the newest, at the tail) is always kept regardless of its token count.

**Reproduction:**
```js
// Single 5000-char message (~1250 tokens) with 100-token budget
trimBootstrapMessagesToBudget([{ role: 'assistant', content: 'x'.repeat(5000) }], 100)
// Before fix: returns [message]  (1250 tokens, 12.5x over budget)
// After fix:  returns []          (budget respected)
```

This matters in practice because tool results and large assistant responses can easily be 50K+ tokens as the tail message.

### 2. NaN propagates through all numeric env config parsing

Every numeric env var in `resolveLcmConfig()` uses raw `parseInt()` / `parseFloat()`:

```ts
(env.LCM_LEAF_CHUNK_TOKENS !== undefined ? parseInt(env.LCM_LEAF_CHUNK_TOKENS, 10) : undefined)
  ?? toNumber(pc.leafChunkTokens) ?? 20000
```

`parseInt('oops', 10)` returns `NaN`, which is **not nullish**, so `??` never fires. This propagates `NaN` through `leafChunkTokens`, `bootstrapMaxTokens`, and every derived value — effectively disabling all token budgets system-wide.

**Reproduction:**
```js
const config = resolveLcmConfig({ LCM_LEAF_CHUNK_TOKENS: 'oops' }, {});
// Before fix: { leafChunkTokens: NaN, bootstrapMaxTokens: NaN }
// After fix:  { leafChunkTokens: 20000, bootstrapMaxTokens: 6000 }
```

## Fix

1. **Oversized singleton:** After the trim loop, check if the single kept message exceeds the budget. If so, return empty array instead of silently bypassing.

2. **NaN propagation:** Add `parseFiniteInt()` and `parseFiniteNumber()` helpers that return `undefined` for non-finite results. Replace all 16 raw `parseInt`/`parseFloat` calls in `resolveLcmConfig()` with the safe helpers, so `??` fallback works correctly for invalid input.

## Testing

- Both syntax checks pass
- Both bugs reproduced and verified fixed with minimal scripts
- Adversarial review by independent agent confirmed both issues

### DIFF
--- a/.changeset/bright-horses-cough.md
+++ b/.changeset/bright-horses-cough.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add regression coverage for bootstrap budget edge cases and invalid numeric env fallback behavior.
--- a/test/config.test.ts
+++ b/test/config.test.ts
@@ -379,6 +379,24 @@ describe("resolveLcmConfig", () => {
     expect(config.bootstrapMaxTokens).toBe(4321);
   });
 
+  it("falls back cleanly when numeric env vars are invalid", () => {
+    const config = resolveLcmConfig({
+      LCM_LEAF_CHUNK_TOKENS: "oops",
+      LCM_BOOTSTRAP_MAX_TOKENS: "still-nope",
+      LCM_CONTEXT_THRESHOLD: "bad",
+      LCM_SUMMARY_MAX_OVERAGE_FACTOR: "nah",
+    } as NodeJS.ProcessEnv, {
+      leafChunkTokens: 80_000,
+      contextThreshold: 0.5,
+      summaryMaxOverageFactor: 5,
+    });
+
+    expect(config.leafChunkTokens).toBe(80_000);
+    expect(config.bootstrapMaxTokens).toBe(24_000);
+    expect(config.contextThreshold).toBe(0.5);
+    expect(config.summaryMaxOverageFactor).toBe(5);
+  });
+
   it("reads summaryMaxOverageFactor and maxAssemblyTokenBudget from plugin config", () => {
     const config = resolveLcmConfig({}, {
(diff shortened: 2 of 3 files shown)


100yenadmin:
Thanks for the review feedback — addressed the docstring inconsistency (pushed).

On the test coverage requests (oversized-tail regression test + NaN env config fallback test): both are valid and I'd be happy to add them, but I don't have test infrastructure wired up in the fork. If you'd prefer I add the test cases in a follow-up commit, could you point me at the right test runner setup? Otherwise I can describe the test cases and let a maintainer add them:

**Oversized tail test (`engine.test.ts`):**
```ts
// Single message exceeding bootstrapMaxTokens should return empty
const result = trimBootstrapMessagesToBudget(
  [{ role: 'assistant', content: 'x'.repeat(5000) }], // ~1250 tokens
  100
);
expect(result).toEqual([]);
```

**NaN config fallback test (`config.test.ts`):**
```ts
// Invalid env should fall back to defaults, not NaN
const config = resolveLcmConfig({ LCM_LEAF_CHUNK_TOKENS: 'oops' }, {});
expect(config.leafChunkTokens).toBe(20000);
expect(Number.isFinite(config.bootstrapMaxTokens)).toBe(true);

const config2 = resolveLcmConfig({ LCM_CONTEXT_THRESHOLD: 'invalid' }, {});
expect(config2.contextThreshold).toBe(0.75);
```

Issue #271, opened 2026-04-05
reconcileSessionTail: exponential message accumulation when session contains many identical messages (empty assistant, NO_REPLY, generic tool outputs)


vipitec-neo:
## Summary

> Disclosure: This bug report was written by an AI agent (OpenClaw admin-navigator, Claude) on behalf of its human operator. We encountered this issue in production and traced it through the source code.

Even after upgrading from v0.5.3 to v0.6.1, `reconcileSessionTail()` causes exponential message accumulation in conversations with many content-identical messages. The v0.6.1 dedup fix (occurrence-count matching) does not resolve the issue for sessions with heavy tool usage.

**We have identified a working patch** (see "Applied Workaround" below) that we are running in production since 2026-04-05.

## Environment

- lossless-claw version: 0.6.1 (upgraded from 0.5.3)
- OpenClaw version: 2026.4.2
- Database: SQLite (lcm.db)
- Agent type: Multi-agent system with extensive tool usage (~11 tool calls per user turn)

### Setup

An agent session with heavy tool usage that produces many content-identical messages:
- Empty assistant messages (content = "") — placeholder or auto-pruned messages between or within tool-call blocks
- `NO_REPLY`, `HEARTBEAT_OK`, etc. responses
- Generic tool outputs: `(no output)`, `(no new output)`, `(Command exited with code 0)`, etc.

### Steps

1. Fresh database: delete conversation, VACUUM, restart gateway
2. Bootstrap imports ~6,900 JSONL entries → ~11,500 unique messages ✅
3. Send one message to the agent (triggers afterTurn → reconcileSessionTail)
4. Message count jumps to ~17,000, then ~22,000, then ~54,000+ within minutes
5. Each afterTurn cycle imports thousands of duplicate messages

### Observed behavior

| Time  | Messages | Growth                          |
|-------|----------|---------------------------------|
| 03:03 | 11,496   | Fresh bootstrap (clean)         |
| 03:21 | 17,163   | +5,667 after first afterTurn    |
| 03:25 | 22,823   | +5,660                          |
| 04:15 | 49,000+  | Exponential growth continues    |
| 04:28 | 54,011   | Still growing                   |

Key observation: All seq numbers are unique (no seq-level duplicates), but 92.3% of message content is duplicated:

Top content duplicates:
```
17,942× [assistant] content='' (empty string)
   943× [tool] content='(no output)'
   310× [tool] content='Successfully replaced 1 block(s) in ...'
[...]

Issue #268, opened 2026-04-04
compactFullSweep unbounded while(true) loop causes 50-minute session lockout (doom loop)


Adam-Researchh:
## Summary

A long-running conversation (10,636 messages, ~1.4M characters, 266 summaries over 3 days) triggered a compaction doom loop that locked the session for ~50 minutes with zero responsiveness. The root cause is multiple compounding failures in the compaction pipeline that lack adequate circuit breakers.

## Environment

- **lossless-claw:** 0.5.3 → upgraded to 0.6.1 during the incident (11:24 EDT)
- **OpenClaw:** 2026.4.2
- **Main model:** anthropic/claude-opus-4-6 (1M context window)
- **Summary model:** anthropic/claude-haiku-4-5 (with Opus fallback on timeout)
- **Config at time of incident:** All defaults (no explicit LCM config set)

## Timeline (2026-04-04, all times EDT)

| Time | Event |
|------|-------|
| 11:24 | LCM upgraded 0.5.3 → 0.6.1 |
| 11:27 | First compaction wave begins (~80 passes) |
| 11:40 | LLM run times out, context at 75% capacity |
| 11:40 | OpenClaw `timeout-compaction` triggers second compaction wave |
| 11:48–11:59 | Second wave runs (more passes) |
| 11:59 | Haiku summarizer times out at 60s, falls back to Opus |
| 12:05–12:11 | Third compaction wave |
| 12:07, 12:13 | Gateway restarts (user reset attempts, session still locked) |
| 12:15 | Conversation archived, new conversation created |

**Total: 202 compaction passes over ~50 minutes. Session completely unresponsive the entire time.**

### 1. `compactFullSweep` Phase 1 has no max-passes cap (PRIMARY)

`compaction.ts` Phase 1 leaf passes use `while (true)` with exit conditions that are structurally inadequate for large conversations:

```typescript
// Phase 1: leaf passes — while(true), no maxPasses guard
while (true) {
  // exits: under threshold, no more chunks, no progress, auth failure
  // but with 10K+ messages and 2K leafChunkTokens, this runs 70-90 passes
}
```

By contrast, `compactUntilUnder` has `maxRounds: 10`. `compactFullSweep` — which is what actually gets called on threshold compaction and manual triggers — has no equivalent cap.

**Suggested fix:** Add a `maxPasses` (configurable, default ~20-30) to the Phase 1 while loop. If exhausted, return partial progress and let the next turn continue incrementally.

[...]

PR #277, opened 2026-04-05
fix: update bootstrap checkpoint after maintain() JSONL rewrite

### LINKED ISSUE #276: fix: stale bootstrap checkpoint after maintain() JSONL rewrite causes duplicate flood
## Summary

`maintain()` calls `rewriteTranscriptEntries()` which rewrites the JSONL session file via branch-and-reappend, changing the file size, mtime, and entry IDs. However, `conversation_bootstrap_state` is only updated during `bootstrap()`, not after maintenance rewrites.

On the next gateway restart, bootstrap sees the stale checkpoint:
- Fast path 1 (exact size+mtime) fails — file shrank by `bytesFreed`
- Fast path 2 (append-only) fails — requires `size > storedSize`, but file is smaller
- Falls through to `reconcileSessionTail()` full content-based reconcile
- On conversations with many identical `(role, content)` pairs (empty assistant messages, repeated tool outputs — common in agent workflows), the occurrence-counting anchor matches at the wrong position → thousands of duplicate messages imported

## Reproduction

1. Active LCM conversation with repeated patterns (empty assistant messages, identical tool outputs)
2. `maintain()` runs transcript GC (called after bootstrap, turns, compaction)
3. `rewriteTranscriptEntries()` shrinks the JSONL — checkpoint now stale
4. Gateway restart → both fast paths fail → full reconcile → anchor mismatch → flood
5. Flood inflates context → doom loop (#268) → cascading failure

## Impact

19+ flood events on a single conversation (17K messages) between March 29 – April 5, 2026. Worst case: 5,733 duplicate messages imported in one second, context_items exploded 41×, caused fleet-wide outage.

## Fix

Two parts:

### 1. Primary: Update checkpoint after maintain() rewrite
After `rewriteTranscriptEntries()` returns `changed: true`, call `upsertConversationBootstrapState()` with new file size/mtime/hash. Ensures next bootstrap hits the fast path.

### 2. Defense-in-depth: Import cap in reconcileSessionTail()
If reconcile would import more than 20% of existing DB count (min 50), abort and log an error instead of importing. Prevents catastrophic floods even if the checkpoint fix somehow fails.

## Related Issues
- #268 (doom loop) — floods trigger the doom loop
- #271 (reconcileSessionTail accumulation) — same root cause, was closed prematurely
- #263 (replay pollution) — fixed afterTurn dedup but not bootstrap path

## Environment
- OpenClaw 3.28, lossless-claw 0.6.1
- macOS 26.3.1, M2, 24GB

liu51115:
Fixes #276 (and the root cause of #271, #268 cascade)

## What

Two changes:

### 1. Update bootstrap checkpoint after maintain() transcript GC

`maintain()` calls `rewriteTranscriptEntries()` which rewrites the JSONL via branch-and-reappend. This changes file size, mtime, and entry IDs — but nobody updated `conversation_bootstrap_state`. The checkpoint was only written during `bootstrap()`.

Now after a successful rewrite (`changed: true`), `maintain()` reads the new file state and updates the checkpoint. The next bootstrap hits the fast path instead of falling through to the fragile content-based `reconcileSessionTail()`.

### 2. Defense-in-depth import cap in reconcileSessionTail()

If reconcile would import more than 20% of existing DB count (min 50), it aborts and logs an error instead of blindly importing thousands of duplicates.

## Why

`reconcileSessionTail()` uses `messageIdentity(role, content)` with occurrence counting to find the anchor point. On conversations with many identical messages (empty assistant, repeated tool outputs — common in agent workflows), the occurrence count matches at the wrong position by coincidence, importing thousands of duplicates.

This caused 19+ flood events on a production fleet between March 29–April 5, 2026, including a 5,733-message flood that caused a fleet-wide outage.

## Also includes

- maxLeafPasses cap for compactFullSweep (#268)
- Circuit breaker error counting for timeouts/crashes (not just auth failures)
- pinnedFiles feature (two prior commits)

## Tests

All 546 tests pass. The checkpoint update is defensive (try/catch wrapped, logs errors but doesn't fail maintain()).

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1918,6 +1918,20 @@ export class LcmContextEngine implements ContextEngine {
     }
 
     const missingTail = historicalMessages.slice(anchorIndex + 1);
+
+    // Defense-in-depth: cap reconcile imports at 20% of existing DB message
+    // count to prevent catastrophic duplicate floods if the anchor lands at
+    // the wrong position (e.g. due to many identical messages).
+    const existingDbCount = await this.conversationStore.getMessageCount(conversationId);
+    if (existingDbCount > 0 && missingTail.length > Math.max(existingDbCount * 0.2, 50)) {
+      console.error(
+        `[lcm] reconcileSessionTail: aborting — would import ${missingTail.length} messages ` +
+          `into conversation ${conversationId} with ${existingDbCount} existing messages ` +
+          `(exceeds 20% cap). This likely indicates a stale bootstrap checkpoint.`,
+      );
+      return { importedMessages: 0, hasOverlap: true };
+    }
+
     let importedMessages = 0;
     for (const message of missingTail) {
       const result = await this.ingestSingle({ sessionId, sessionKey: params.sessionKey, message });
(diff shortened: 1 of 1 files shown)


liu51115:
Closing — splitting into separate PRs per fix: bootstrap checkpoint (#276), doom loop cap (#268), and pinnedFiles.

PR #278, opened 2026-04-05
fix: update bootstrap checkpoint after maintain() to prevent replay floods

liu51115:
## Root cause of #268, #271, #276

This is the root fix for the recurring bootstrap replay flood that causes doom loops (#268), exponential message accumulation (#271), and session lockouts (#276).

### The bug

`maintain()` calls `rewriteTranscriptEntries()` after every successful turn. This rewrites the JSONL via branch-and-reappend — changing file size, mtime, and entry IDs. But `conversation_bootstrap_state` was **only updated during `bootstrap()`**, never after maintenance rewrites.

On the next gateway restart:
1. Fast path 1 fails (size/mtime mismatch from stale checkpoint)
2. Fast path 2 fails (entry at stored offset is now a different entry, or a non-message type like `openclaw.cache-ttl` — `readLastJsonlEntryBeforeOffset` returns `null`, hash comparison fails)
3. Falls through to `reconcileSessionTail()` — content-based anchor matching using occurrence counting on `(role, content)` identity

On conversations with repeated identical messages (empty assistant turns, duplicate tool outputs, `NO_REPLY` patterns), the occurrence-count anchor lands thousands of entries too early, re-importing everything after the false anchor as duplicates.

### Evidence from production

Our conv 1 (17K messages) experienced **19 flood events across 453 gateway restarts** since LCM activation on March 29. Every flood maps 1:1 to a restart where the JSONL had been rewritten by `maintain()` since the last checkpoint update. The final flood (April 5) imported 5,733 duplicates in one second, inflating context to 1.7M tokens and triggering a 50-minute compaction doom loop that cascaded into a fleet-wide outage.

The doom loop (#268) is a consequence, not the cause. `compactFullSweep` runs correctly — the problem is that it's asked to compact 1.7M tokens that shouldn't exist. Fix the replay flood and the doom loop doesn't trigger.

### Fix

Two changes:

1. **Checkpoint update after maintain()** (primary fix): When `rewriteTranscriptEntries()` returns `changed: true`, stat the session file and call `upsertConversationBootstrapState()` with fresh size/mtime/offset/hash. Next restart hits the fast path. Wrapped in try/catch so a checkpoint failure doesn't break the rewrite.

2. **Import cap in reconcileSessionTail()** (defense-in-depth): If the reconcile would import more than `max(existingDbCount * 0.2, 50)` messages, abort with a warning log. First bootstrap (`existingDbCount === 0`) is exempt. This caps damage if reconcile fires for any reason we haven't anticipated.

### Testing

All 533 tests pass (546 minus 13 pinnedFiles tests on separate branch).

### Related issues

- #268 — compactFullSweep doom loop (consequence of flood inflating context)
- #271 — reconcileSessionTail exponential accumulation with identical messages (this is the mechanism)
- #276 — bootstrap checkpoint stale after maintain() (this PR)
- #263 — afterTurn dedup fix (fixed a different code path; bootstrap path was unfixed)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1,6 +1,6 @@
 import { createHash, randomUUID } from "node:crypto";
 import { closeSync, createReadStream, openSync, readSync, statSync } from "node:fs";
-import { mkdir, writeFile } from "node:fs/promises";
+import { mkdir, stat, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import { join } from "node:path";
 import type { DatabaseSync } from "node:sqlite";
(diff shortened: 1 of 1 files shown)


liu51115:
Closing — checkpoint update alone is insufficient. Non-message entries (cache-ttl) at the stored offset still cause hash=null → full reconcile → flood. Need to also fix readLastJsonlEntryBeforeOffset to skip non-message entries. Will reopen with complete fix.

PR #303, opened 2026-04-06
fix: increase SQLite busy_timeout from 5s to 30s

gitchrisqueen:
## Problem

On gateway startup, if the OpenClaw TUI holds a WAL write lock on `lcm.db` (e.g. during session restore), the migration step fails after only 5 seconds with:

```
Migration failed: database is locked
```

This cascades to the compaction model showing as `(unconfigured)` since initialization only completes after a successful migration:

```
[plugins] [lcm] Compaction summarization model: (unconfigured)
```

## Fix

Increase `SQLITE_BUSY_TIMEOUT_MS` from `5_000` to `30_000`. This gives sufficient time for the TUI to release its lock during normal startup sequences without requiring any retry logic.

## Testing

Verified on a system where the TUI was active during gateway restart:
- Before: `Migration failed: database is locked` consistently on restart when TUI was running
- After: Migration succeeds; `Compaction summarization model: litellm/...` on every startup

### DIFF
--- a/src/db/connection.ts
+++ b/src/db/connection.ts
@@ -3,7 +3,7 @@ import { dirname, resolve } from "node:path";
 import { DatabaseSync } from "node:sqlite";
 
 type ConnectionKey = string;
-const SQLITE_BUSY_TIMEOUT_MS = 30_000;
+const SQLITE_BUSY_TIMEOUT_MS = 5_000;
 
 const connectionsByPath = new Map<ConnectionKey, Set<DatabaseSync>>();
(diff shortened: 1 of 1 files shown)


jalehman:
Closing — the fix you were going for (busy_timeout 5s → 30s) just landed via #413.

The 0-line diff here was because the second commit (cf38153, a Copilot co-authored suggestion) accidentally reverted the first commit back to 5s. So the two commits cancelled each other out.

Thanks for surfacing this — the TUI lock contention scenario was a helpful data point.

PR #305, opened 2026-04-07
[fix] fix startup summary model banner resolution

stilrmy:
## Summary

Fix the misleading startup banner that reports the compaction summarization model as `(unconfigured)` when `lossless-claw` registers before `api.pluginConfig` is populated.

Fixes #304.

## What changed

- resolve registration-time config from the validated runtime config only when the early plugin config surface is empty
- read `plugins.entries.lossless-claw.config` from that validated config for startup-time dependency resolution
- feed the same effective OpenClaw config into the startup banner path
- add a regression test covering the early-startup case where runtime config is available before `api.pluginConfig`

## Root cause

The startup banner path depended on `api.config` and `api.pluginConfig` during `register()`, but OpenClaw can invoke plugin registration before those surfaces are fully populated. Later runtime compaction used the correct config path, so compaction worked while the banner incorrectly logged `(unconfigured)`.

## Impact

Startup logs now reflect the effective summary model consistently with runtime compaction behavior, without changing the actual summarizer resolution path.

## Validation

- `npm test -- --run test/plugin-config-registration.test.ts`
- `npm test -- --run test/plugin-config-registration.test.ts test/index-complete-provider-config.test.ts`
- `npm test`

### DIFF
--- a/.changeset/late-cups-rhyme.md
+++ b/.changeset/late-cups-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix startup-time summary model resolution when OpenClaw populates plugin config before the top-level runtime config surface.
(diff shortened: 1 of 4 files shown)


jalehman:
Thank you!

PR #326, opened 2026-04-08
docs: recommend GPT-5.4 mini for LCM

barronlroth:
## Summary
- switch the example summary/expansion model from Anthropic Haiku to OpenAI GPT-5.4 mini
- document that lossless-claw supports runtime-managed OAuth providers for summarization and expansion
- add a cost-sensitive starting configuration using GPT-5.4 mini

## Why
Anthropic quota pressure makes Haiku a poor default example for many OpenClaw users. Lossless-claw already falls back to OpenClaw's configured default model when summary/expansion overrides are unset, so this PR updates the docs and examples rather than changing runtime defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,8 @@ Add a `lossless-claw` entry under `plugins.entries` in your OpenClaw config:
           "ignoreSessionPatterns": [
             "agent:*:cron:**"
           ],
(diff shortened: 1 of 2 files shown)


jalehman:
Thank you! Good change.

PR #351, opened 2026-04-09
fix: prevent overflow recovery from bailing when observed tokens unavailable

kitcommerce:
## Summary

- When the preemptive context overflow guard fires during the tool loop, the error message doesn't include an observed token count, so `observedTokens` is `undefined` in the overflow recovery path
- `compactUntilUnder()` then uses only the stored token count (which is low because `afterTurn` hasn't ingested the current turn yet) and bails with "already under target" — even though the live context is actually overflowing
- This causes the session to be nuked and restarted instead of being compacted

## Fix

When `force=true` (overflow recovery) and `observedTokens` is `undefined`, pass `tokenBudget` as `currentTokens` to `compactUntilUnder()`. This ensures the convergence loop knows we're at least at the budget and proceeds with compaction.

The fix is 6 lines in `compact()` in `engine.ts`.

## Reproduction

1. Use a small-context model like `gemini-2.5-flash-lite` (150K context window)
2. Accumulate messages in a session until the preemptive context guard fires
3. Observe: `auto-compaction failed for google/gemini-2.5-flash-lite: already under target` followed by session restart
4. After fix: `auto-compaction succeeded for google/gemini-2.5-flash-lite; retrying prompt`

## Test plan

- [x] Verified fix in production — compaction now runs 3 leaf passes (108745 → 87472 tokens) and prompt retry succeeds
- [ ] Unit test for `compactUntilUnder` with `force=true` and no `currentTokens`

### DIFF
--- a/.changeset/pr-351-overflow-recovery.md
+++ b/.changeset/pr-351-overflow-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Ensure forced overflow recovery still runs compaction when live observed token counts are unavailable.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -4020,11 +4020,21 @@ export class LcmContextEngine implements ContextEngine {
             ? decision.threshold
             : tokenBudget;
 
+        // When forced (overflow recovery) and the caller did not supply an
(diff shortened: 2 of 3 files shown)


jalehman:
Thank you!

PR #355, opened 2026-04-10
feat: inline image detection and tool result string format fix

LanicBlue:
## Summary

- **Unified inline image detection**: Detect and externalize base64 image data (JPEG, PNG, GIF, WebP, SVG) embedded in messages of any role (user/tool/assistant). Images are saved as binary files. Handles both OpenClaw's `[media attached:]` user pattern and pure base64 payloads in any message content. Image detection is a standalone system independent of the large file text pipeline.
- **String format fix for tool results**: Normalize string-format tool result content to array format before processing, ensuring large string tool outputs are properly externalized instead of silently skipped.

## Test plan

- [ ] Verify user image attachments are externalized and saved as valid binary files
- [ ] Verify tool results containing base64 image data are externalized
- [ ] Verify string-format tool results with large outputs are properly externalized
- [ ] Run existing test suite to confirm no regressions

### DIFF
--- a/.changeset/wise-bears-doubt.md
+++ b/.changeset/wise-bears-doubt.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": minor
+---
+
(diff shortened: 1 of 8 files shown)


jalehman:
Thank you!

PR #356, opened 2026-04-10
feat: improve heartbeat filtering with ingest-time detection

LanicBlue:
## Summary

- Replace coarse `isHeartbeat` flag with per-message content-based filtering at ingest time
- Correctly handles heartbeat messages with system notifications prepended (Exec results, Hook output, Gateway restarts)
- Extend pruning to also clean empty messages and NO_REPLY assistant messages
- Net reduction of 32 lines by removing `batchLooksLikeHeartbeatAckTurn` and `turnLooksLikeHeartbeatTurn`

### Problem

The upstream heartbeat filtering relied on the runtime `isHeartbeat` flag, which missed heartbeat user messages when system events (Exec completed/failed, Hook results, Gateway restart) were prepended to the heartbeat instruction. These messages were stored verbatim, polluting the context.

The turn-level detection (`batchLooksLikeHeartbeatAckTurn`) also had edge cases where it would skip pruning when the batch didn't look like a pure heartbeat turn.

### Approach

- **Ingest-time filtering**: Check message content directly instead of relying on the `isHeartbeat` flag
  - User messages: detect by `includes("read heartbeat.md if it exists (workspace context)")`
  - Assistant messages: detect by exact `HEARTBEAT_OK` match
- **Retroactive pruning**: Always run `pruneHeartbeatOkTurns` after each turn (not conditionally) to clean legacy messages stored before ingest-time filtering was added
- **Extended cleanup scope**: Also prune empty messages (`""`, `"[]"`, `"{}"`) and NO_REPLY assistant messages

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3157,9 +3157,22 @@ export class LcmContextEngine implements ContextEngine {
     message: AgentMessage;
     isHeartbeat?: boolean;
   }): Promise<IngestResult> {
-    const { sessionId, sessionKey, message, isHeartbeat } = params;
-    if (isHeartbeat) {
-      return { ingested: false };
+    const { sessionId, sessionKey, message } = params;
+
+    // Filter heartbeat user messages at ingest time
+    if (message.role === "user") {
+      const content = extractMessageContent(message.content);
+      if (isHeartbeatUserMessage(content)) {
+        return { ingested: false };
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


100yenadmin:
@LanicBlue triage pass update: I marked this priority:P3 enhancement/stale-check. Heartbeat filtering is useful operational cleanup, but the branch is dirty and old. Can you confirm whether it still applies cleanly to current ingest behavior?

PR #362, opened 2026-04-10
fix: make cache-aware compaction resilient to routing noise

### LINKED ISSUE #358: cacheAwareCompaction: OR routing noise triggers destructive compaction cascade
## Summary

When a session falls back to OpenRouter (OR), LCM's cacheAwareCompaction triggers a destructive compaction cascade due to OR's lack of server affinity. Random cache misses from load balancer routing cause `cacheState` to oscillate between `hot` and `cold`, triggering aggressive compaction that destroys context unnecessarily.

## Root Cause

OR load-balances requests across multiple Anthropic backend instances, each with its own prompt cache. Consecutive calls may hit different servers, causing random cache misses (`cacheRead=0`) even with identical prefixes. These are routing artifacts, not genuine cache invalidations.

LCM's current logic treats each single cold reading as authoritative:

1. A single `cacheState=unknown` (from a failed/timed-out call at provider switch) drops `triggerLeafChunkTokens` from 40K to base 20K → compaction fires immediately on any session >20K tokens
2. `cacheState=cold` (from a routing miss) triggers `cold-cache-catchup` with `maxPasses=2` → double-pass aggressive compaction
3. Compaction rewrites the prefix → guarantees the next call is cold → more compaction → feedback loop

## Data

From a production session on OR, consecutive calls with identical prefixes showed random cache misses:

```
| Call | CacheRead | CacheWrite | State | Notes                         |
|------|-----------|------------|-------|-------------------------------|
| A    | 119,138   | 0          | HOT   |                               |
| A+1  | 119,138   | 0          | HOT   | 11s later                     |
| A+2  | 119,138   | 0          | HOT   | 8s later                      |
| A+3  | 119,138   | 0          | HOT   | 11s later                     |
| A+4  | 119,138   | 0          | HOT   | 8s later                      |
| A+5  | 0         | 119,138    | COLD  | 18s later — same prefix, miss |
```

5 consecutive hot calls, then a cold miss with no compaction in between. Same prefix, same session. Pure routing noise.

Separately, a different hot streak:

```
| Call | CacheRead | CacheWrite | State | Notes                           |
|------|-----------|------------|-------|---------------------------------|
| B    | 81,786    | 0          | HOT   |                                 |
| B+1  | 0         | 81,786     | COLD  | 33s later — routing miss        |
| B+2  | 81,786    | 0          | HOT   | 7s later — back to hot          |
```

Hot → cold → hot in 40s. No compaction. The cold call is a routing anomaly.

### The Cascade

These random cold readings trigger LCM's compaction:

1. `cacheState=unknown` (provider failover, no usage data) → `triggerLeafChunkTokens` drops from 40K to 20K → compaction fires on a 66K session
2. **Compaction: 90K → 72K** (leaf pass)
3. Next OR call happens to be cold (routing miss) → `cold-cache-catchup` triggers double-pass
4. **Compaction: 82K → 43K → 35K** (leaf + condensed)

Total over the session: **8 compaction events** during OR periods, ~230K tokens of raw context compacted away — all triggered by routing noise rather than genuine memory pressure.

## Proposed Fix

Cache state determination should be **sticky** rather than single-call. Options:

1. **Require N consecutive cold readings** before treating cache as cold (e.g., 3). One random cold miss shouldn't trigger aggressive compaction.
2. **Treat `unknown` as `hot`** (or at minimum use the dynamic max threshold instead of base) — unknown means no data, not cold.
3. **Use a rolling window / decay** — weight recent cache observations instead of using only the latest.

The key insight: **compaction is irreversible, cache misses are transient.** The conservative response to uncertain cache state should be "don't compact" rather than "compact aggressively."

## Environment

- LCM 0.8.0 with cacheAwareCompaction enabled
- OpenRouter as fallback provider (Anthropic direct → OR on timeout)
- leafChunkTokens=20000, dynamicLeafChunkTokens.max=40000, hotCachePressureFactor=4

## Related

Full analysis with data available on request.

100yenadmin:
## Summary

This fixes the cache-noise failure mode behind #358.

Current `main` treats a single cold observation as authoritative unless a very short hot-cache hysteresis window happens to save the session. That is too aggressive for provider failover and OpenRouter routing noise: a single `cacheRead=0` can drop a previously hot session into cold-cache catch-up and trigger destructive compaction that was not justified by real memory pressure.

This PR makes cache-aware compaction state sticky instead of single-sample authoritative.

## What changed

- add `cacheAwareCompaction.coldCacheObservationThreshold` (default `3`)
- persist a `consecutive_cold_observations` counter in compaction telemetry
- keep explicit cache breaks authoritative immediately
- keep prior hot sessions effectively hot through one-off `cold` and `unknown` blips until the cold-observation threshold is reached
- preserve existing budget-pressure behavior once a session is truly under pressure

## Why this helps

The conservative failure mode for cache-aware compaction should be “preserve context until we are confident the cache is really cold,” not “compact aggressively because one read looked cold.”

This is especially important on OpenRouter, where identical prefixes can still produce transient cold reads because requests land on different backend instances.

## Tests

- `pnpm exec vitest run test/config.test.ts test/engine.test.ts test/session-operation-queues.test.ts test/circuit-breaker.test.ts test/expansion.test.ts`

Adds focused coverage for:
- single cold reading after a prior hot streak stays non-authoritative
- repeated cold readings eventually become authoritative
- unknown/no-usage observations do not collapse hot-cache protection
- explicit breaks still force cold behavior immediately

Closes #358.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -170,6 +170,7 @@ openclaw plugins install --link /path/to/lossless-claw
 | `cacheAwareCompaction.maxColdCacheCatchupPasses` | `integer` | `2` | `LCM_MAX_COLD_CACHE_CATCHUP_PASSES` | Maximum bounded catch-up passes allowed in one maintenance cycle when cache telemetry is cold. |
 | `cacheAwareCompaction.hotCachePressureFactor` | `number` | `4` | `LCM_HOT_CACHE_PRESSURE_FACTOR` | Multiplier applied to the hot-cache leaf trigger before raw-history pressure overrides cache preservation. |
 | `cacheAwareCompaction.hotCacheBudgetHeadroomRatio` | `number` | `0.2` | `LCM_HOT_CACHE_BUDGET_HEADROOM_RATIO` | Minimum fraction of the real token budget that must remain free before hot-cache incremental compaction is skipped entirely. |
+| `cacheAwareCompaction.coldCacheObservationThreshold` | `integer` | `3` | `LCM_COLD_CACHE_OBSERVATION_THRESHOLD` | Consecutive cold observations required before non-explicit cache misses are treated as truly cold. This dampens one-off routing noise and provider failover blips. |
 
 #### `dynamicLeafChunkTokens`
 
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -169,6 +169,10 @@
       "label": "Hot Cache Budget Headroom",
       "help": "Fraction of the real token budget that must remain free before hot-cache incremental compaction is skipped entirely"
     },
+    "cacheAwareCompaction.coldCacheObservationThreshold": {
+      "label": "Cold Cache Observation Threshold",
+      "help": "Consecutive cold observations required before non-explicit cache misses are treated as truly cold"
+    },
     "dynamicLeafChunkTokens.enabled": {
       "label": "Dynamic Leaf Chunk Tokens",
       "help": "When enabled, incremental compaction uses a larger working leaf chunk in busy sessions and keeps the static floor in quieter sessions"
@@ -355,6 +359,10 @@
             "type": "number",
             "minimum": 0,
             "maximum": 0.95
+          },
+          "coldCacheObservationThreshold": {
+            "type": "integer",
+            "minimum": 1
           }
         }
       },
--- a/skills/lossless-claw/references/config.md
+++ b/skills/lossless-claw/references/config.md
@@ -87,6 +87,7 @@ Good defaults:
 - `maxColdCacheCatchupPasses: 2`
 - `hotCachePressureFactor: 4`
 - `hotCacheBudgetHeadroomRatio: 0.2`
+- `coldCacheObservationThreshold: 3`
 
 Operationally:
 
--- a/skills/lossless-claw/references/config.md
+++ b/skills/lossless-claw/references/config.md
@@ -384,6 +384,19 @@ Default:
 
 - `0.2`
 
+#### `cacheAwareCompaction.coldCacheObservationThreshold`
+
+Consecutive cold observations required before non-explicit cache misses are treated as truly cold.
+
+Why it matters:
+
+- prevents a single OpenRouter routing miss or provider failover blip from immediately triggering cold-cache catch-up
+- explicit cache breaks still count as cold immediately
+
+Default:
+
+- `3`
+
 ### `dynamicLeafChunkTokens`
 
 #### `dynamicLeafChunkTokens.enabled`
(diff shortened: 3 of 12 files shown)


liu51115:
cc @100yenadmin @jalehman

I traced the interaction between this PR and #408 on current `main`, informed by earlier work on #385 (which #408 superseded). I don't think #362 is needed after #408, and it actively interferes with #408's design.

## Background: the afterTurn timing problem

This was the central insight behind #385 (later adopted by #408): `afterTurn()` fires right after the API call, when the cache was just written. You can't know from afterTurn whether the cache will still be valid at the next assembly — that's a wall-clock question. afterTurn is structurally the wrong place to make compaction decisions.

#408 solved this correctly: afterTurn only records debt, actual consumption is gated by wall-clock TTL (`shouldDelayPromptMutatingDeferredCompaction` → `isPromptCacheStillHot`).

## Why #362 is unnecessary under #408

#362's stated problem: "a single `cacheRead=0` can trigger destructive compaction."

Under #408's model, this problem doesn't exist:

1. Cold blip arrives → afterTurn records debt (cheap, no harm)
2. Next assembly → TTL gate checks wall-clock time → cache still hot → delays consumption
3. Cache eventually expires → next assembly consumes the debt → compaction runs when it's actually free

The TTL gate is inherently blip-proof. It doesn't matter if you see 1 cold observation or 10 — the gate checks the clock, not observation counts. A noisy cold read never reaches execution because the cache hasn't actually expired.

So the "routing noise" failure mode #362 addresses **no longer exists** after #408. There is nothing left for observation-count smoothing to fix.

## Why #362 is actively harmful with #408

#362 modifies `resolveCacheAwareState()` to smooth cold observations → keeps state "hot" after blips. This feeds `evaluateIncrementalCompaction()` in afterTurn, which can then return `shouldCompact=false` via `hot-cache-defer` or `hot-cache-budget-headroom` — preventing debt from being recorded at all.

- **#408 alone**: cold blip → debt recorded → TTL gate handles timing → executes when cache is actually cold
- **#408 + #362**: cold blip smoothed to "hot" → no debt recorded → TTL gate never gets to run → compaction doesn't happen

#362 pre-empts #408's time-aware mechanism with a time-unaware one. It suppresses the signal that #408 would have handled correctly.

## The deeper issue: wrong primitive

Cache liveness is a function of wall-clock time, not observation count:
- 1 cold observation after TTL expiry = genuinely cold. Smoothing it away is wrong.
- 3 cold observations within TTL = noise. But #408's TTL gate already catches this.
- 0 cold observations after hours of silence = cold. Observation counting can't detect this.

#408 uses the right primitive (wall-clock TTL). #362 uses the wrong one (consecutive observation counts) and disrupts #408's pipeline in the process.

## Suggestion

Consider reverting #362's `resolveCacheAwareState()` changes. #408's TTL-based consumption gating already solves the noise problem, and #362 interferes with it by preventing debt from being recorded.

jalehman:
Thanks for the analysis @liu51115 — addressed in https://github.com/Martian-Engineering/lossless-claw/pull/434.

PR #363, opened 2026-04-10
fix: move proactive threshold compaction off the reply path

### LINKED ISSUE #313: Compaction blocks reply delivery — should run async after agent_end
## Problem

LCM compaction runs **synchronously in the afterTurn hook**, blocking the reply pipeline. The assistant's response is already generated but cannot be delivered to the user until compaction completes.

Each compaction pass calls the summarization model (Sonnet), taking **3.5–29 seconds**. During this window:
1. The lane lock is held — no new messages can be processed
2. The typing indicator expires — user sees silence
3. The reply sits in a buffer waiting for compaction to finish

On high-traffic conversations with aggressive thresholds, compaction fires every turn, adding 5–15s latency to **every single reply**.

## Evidence

Investigation on a production Opus session (2026-04-07):
- 21 compaction triggers in 33 minutes (leafChunkTokens=2000)
- Each pass added 3.5–29s synchronous delay
- `agent_end` (which releases the lane lock) fires **after** compaction, not before
- User perceived this as the bot "locking up" or "going silent"
- Even after raising thresholds (leafChunkTokens=35000), the single compaction that fired took 29s

## Proposed fix

Decouple compaction from the reply delivery path:

1. **afterTurn**: deliver the reply and fire `agent_end` first
2. **Then**: run compaction asynchronously (fire-and-forget or queued)
3. Compaction errors should not affect the user-facing reply

This is safe because compaction operates on the stored message/summary DB, not on the in-flight reply. The next turn's context assembly will pick up whatever compaction accomplished.

## Workaround

Raising `leafBudgetHeadroomFactor` to 0.95 and `leafChunkTokens` to 80000 reduces compaction frequency dramatically (25 passes → ~3 per 200-turn session), but doesn't eliminate the blocking when it does fire.

## Related

- #282 (leaf compaction destabilizes prompt cache)
- #306 (cache-aware compaction guards — reduces frequency but doesn't fix blocking)

100yenadmin:
## Summary

This addresses #313 by moving proactive threshold compaction out of the synchronous `afterTurn()` reply path.

Right now reply generation can already be done, but `afterTurn()` still awaits the proactive threshold sweep before it returns. On conversations where compaction is slow or frequent, that turns compaction latency into reply latency.

## What changed

- keep ingest and compaction telemetry in `afterTurn()`
- keep best-effort incremental leaf compaction behavior unchanged
- stop awaiting proactive threshold compaction inline
- schedule threshold compaction onto a tracked background slot keyed by session
- coalesce duplicate background threshold requests so repeated turns do not pile up redundant sweeps
- keep true overflow recovery synchronous and unchanged; this only changes the post-turn maintenance path

## Why this is safe

The proactive threshold sweep operates on persisted conversation state. It does not need to block reply delivery to preserve correctness. The next turn will still see whatever compaction accomplished because the actual mutation still runs through the existing per-session queue.

## Tests

- `pnpm exec vitest run test/engine.test.ts test/config.test.ts test/session-operation-queues.test.ts test/circuit-breaker.test.ts test/expansion.test.ts`

Adds focused coverage for:
- `afterTurn()` resolving before proactive threshold compaction finishes
- duplicate background threshold requests coalescing per session
- background threshold failures not surfacing to the user path
- existing force/overflow recovery behavior remaining intact

## Stack note

This branch is cut on top of #362. Until #362 merges, GitHub will show the lower branch commits here too.

Closes #313.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -158,6 +158,7 @@ openclaw plugins install --link /path/to/lossless-claw
 | `cacheAwareCompaction.maxColdCacheCatchupPasses` | `integer` | `2` | `LCM_MAX_COLD_CACHE_CATCHUP_PASSES` | Maximum bounded catch-up passes allowed in one maintenance cycle when cache telemetry is cold. |
 | `cacheAwareCompaction.hotCachePressureFactor` | `number` | `4` | `LCM_HOT_CACHE_PRESSURE_FACTOR` | Multiplier applied to the hot-cache leaf trigger before raw-history pressure overrides cache preservation. |
 | `cacheAwareCompaction.hotCacheBudgetHeadroomRatio` | `number` | `0.2` | `LCM_HOT_CACHE_BUDGET_HEADROOM_RATIO` | Minimum fraction of the real token budget that must remain free before hot-cache incremental compaction is skipped entirely. |
+| `cacheAwareCompaction.coldCacheObservationThreshold` | `integer` | `3` | `LCM_COLD_CACHE_OBSERVATION_THRESHOLD` | Consecutive cold observations required before non-explicit cache misses are treated as truly cold. This dampens one-off routing noise and provider failover blips. |
 
 #### `dynamicLeafChunkTokens`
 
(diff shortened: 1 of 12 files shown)


jalehman:
Closing as obsolete. The original concern was valid, but current main already moved deferred proactive compaction onto the persisted maintenance-debt path, so this PR’s in-memory background scheduler is no longer the right shape for the codebase. Keeping this open would reintroduce an older approach, and this branch also has a coalescing bug that can skip newer turns.

PR #370, opened 2026-04-10
refactor: simplify evaluateIncrementalCompaction — remove cache-state branching

### LINKED ISSUE #367: cache-aware compaction: timing inversion — decisions based on last-call status instead of cache expiry
## Summary

Cache-aware compaction makes compaction decisions at the wrong time. The current design evaluates compaction in `afterTurn()` using the cache status of the call that just completed. This creates a fundamental timing inversion: the cache state *after* a call is the inverse of what the code assumes.

## The Timing Problem

When a call returns `cacheRead=0` (cold):
- LCM records `cacheState=cold` → triggers aggressive `cold-cache-catchup` (double pass)
- But the provider just *wrote* the prefix to cache on that call
- **The cache is now HOT for the next ~5 minutes**
- Compacting now destroys a freshly-written cache — the worst possible time

When a call returns `cacheRead=119K` (hot):
- LCM records `cacheState=hot` → defers compaction
- Correct in the moment, but the cache has a ~5 minute TTL
- If the session goes idle for >5 min, the cache expires naturally
- But `afterTurn()` never fires during idle periods — no compaction happens when it would be free

**In both cases, the decision is backwards:**
- Cold reading → cache is actually hot now → should NOT compact
- Hot reading during idle → cache will expire soon → COULD compact for free, but doesn't

## Current Code Path

```
afterTurn() fires immediately after API response
  → updateCompactionTelemetry(cacheState from THIS call)
  → evaluateIncrementalCompaction()
      if cold → aggressive catchup (2 passes, condensed)  ← WRONG: cache just got written
      if hot  → defer                                      ← RIGHT now, but misses idle window
```

The evaluation is synchronous with API calls. But the optimal compaction window is asynchronous — during idle periods when the cache has already expired.

## Proposed Architecture

Instead of deciding compaction based on last-call cache status:

1. **After any call:** schedule a deferred compaction for `now + cacheTTL` (~5 min)
2. **If another call happens before the timer:** cancel and reschedule (cache got refreshed)
3. **On timer expiry with no intervening call:** compact freely — the cache has expired naturally, so rewriting the prefix costs nothing
4. **Budget trigger:** still fires immediately regardless (hard safety limit)

This inverts the current model:
- **Current:** "The last call was cold, so compact aggressively" (wrong — cache is now hot)
- **Proposed:** "No call for 5 minutes, so cache expired — compact now for free" (correct)

## Benefits

- Compaction always happens when the cache is genuinely cold (expired), not when it was just written
- No routing noise sensitivity — OR load balancing becomes irrelevant since we're not reacting to per-call status
- No need for sticky counters or hysteresis heuristics (#362) — the timer handles it cleanly
- Idle sessions get compacted (currently they never do, since afterTurn never fires)

## Relationship to Existing Work

- **#289, #306, #329** — original cache-aware compaction implementation. This issue proposes a fundamental redesign of the timing model.
- **#358** — OR routing noise cascade. A symptom of the timing inversion described here.
- **#362** — sticky cold counter. A mitigation that reduces the symptom but doesn't address the root cause.
- **#363** — moves proactive compaction off the reply path. Complementary — still triggered by afterTurn, but no longer blocking it.

## Complexity

This is a bigger change than #362. It requires:
- A per-session timer/scheduler (could piggyback on existing background compaction from #363)
- Cancellation logic when new calls arrive
- Cache TTL awareness (Anthropic: 5 min default, 1 hour extended)
- Integration with the existing budget-trigger hard limit

liu51115:
## Summary

Remove cache-state-dependent branches from `evaluateIncrementalCompaction()` in afterTurn. Based on v0.8.0.

## Problem

`evaluateIncrementalCompaction()` uses the cache status of the just-completed API call to decide compaction strategy. On load-balanced providers with separate prompt caches per backend instance, random cold readings trigger aggressive multi-pass compaction that destroys live caches. See #358 for data.

The three problematic branches:
- **hot-cache-budget-headroom**: defers compaction when cache is hot and under budget — misses the window when cache expires
- **hot-cache-defer**: defers when pressure is below hotCachePressureFactor × threshold — same issue
- **cold-cache-catchup**: triggers aggressive multi-pass compaction on cold readings — a cold reading often means the cache was just written and is now hot

## Change

Remove all three branches. After this change:
- **budget-trigger**: unchanged, safety valve, always fires
- **below-leaf-trigger**: unchanged, no compaction needed
- **leaf-trigger**: single pass, no condensed passes, no cache-state branching

1 file changed, 3 insertions, 53 deletions.

Closes #367. Related: #358, #362.

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1949,56 +1949,6 @@ export class LcmContextEngine implements ContextEngine {
       });
     }
 
-    if (
-      cacheState === "hot"
-      && this.isComfortablyUnderTokenBudget({
-        currentTokenCount: params.currentTokenCount,
-        tokenBudget: params.tokenBudget,
-      })
-    ) {
-      return this.logIncrementalCompactionDecision({
-        conversationId: params.conversationId,
-        cacheState,
-        activityBand,
-        triggerLeafChunkTokens,
-        preferredLeafChunkTokens,
-        fallbackLeafChunkTokens,
-        rawTokensOutsideTail: leafTrigger.rawTokensOutsideTail,
-        threshold: leafTrigger.threshold,
-        shouldCompact: false,
-        maxPasses: 1,
-        allowCondensedPasses: false,
-        reason: "hot-cache-budget-headroom",
-      });
-    }
-
-    if (
-      cacheState === "hot"
-      && leafTrigger.rawTokensOutsideTail
-        < Math.floor(
-          leafTrigger.threshold * this.config.cacheAwareCompaction.hotCachePressureFactor,
-        )
-    ) {
-      return this.logIncrementalCompactionDecision({
-        conversationId: params.conversationId,
-        cacheState,
-        activityBand,
-        triggerLeafChunkTokens,
-        preferredLeafChunkTokens,
-        fallbackLeafChunkTokens,
-        rawTokensOutsideTail: leafTrigger.rawTokensOutsideTail,
-        threshold: leafTrigger.threshold,
-        shouldCompact: false,
-        maxPasses: 1,
-        allowCondensedPasses: false,
-        reason: "hot-cache-defer",
-      });
-    }
-
-    const maxPasses =
-      cacheState === "cold"
-        ? Math.max(1, this.config.cacheAwareCompaction.maxColdCacheCatchupPasses)
-        : 1;
     return this.logIncrementalCompactionDecision({
       conversationId: params.conversationId,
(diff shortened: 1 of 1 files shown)


liu51115:
Superseded by #385 which is the complete fix: removes all three cache-state branches from afterTurn AND adds pre-assembly cache-TTL compaction. #385 is based on the current upstream/main.

PR #371, opened 2026-04-10
fix: avoid disabling reasoning for reasoning-capable summarizers

holgergruenhagen:
## Summary

Hi! This PR fixes an LCM summarizer edge case for reasoning-capable models.

If the initial summarizer request omits a reasoning setting, some provider compatibility layers can normalize that into a reasoning-disabled request. Some models do not support turning reasoning off, so the first request fails and only the retry succeeds.

I first noticed this with `openrouter/minimax/minimax-m2.7`, but the issue is more general than one specific model.

## What changed

- add a small `reasoningIfSupported` option to the completion bridge
- keep explicit `reasoning` values unchanged and give them priority
- apply the fallback only when the resolved model advertises reasoning support
- update the LCM summarizer to send `reasoningIfSupported: "low"` on the initial request
- keep the existing explicit retry behavior unchanged
- add focused tests and a patch changeset

## Why this approach

This keeps the fix narrow and model-aware.

It avoids a model-specific special case, preserves explicit caller intent, and does not force reasoning on models that do not support it.

## Testing

- `npx vitest run test/index-complete-model-auth.test.ts test/summarize.test.ts test/lcm-summarizer-reasoning.test.ts`

Thanks very much for taking a look.

### DIFF
--- a/.changeset/tiny-pots-teach.md
+++ b/.changeset/tiny-pots-teach.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Avoid treating omitted LCM summarizer reasoning settings like reasoning-disabled requests for reasoning-capable models by applying a low default only when the resolved model supports reasoning.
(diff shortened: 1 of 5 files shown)


jalehman:
Thank yoU!

PR #380, opened 2026-04-10
fix: gate startup backfills with migration state

jalehman:
## What
This PR adds additive startup backfill state to the SQLite migration path so the expensive summary and tool-call backfills stop rerunning after a successful startup, while still retrying cleanly if startup fails before the backfill version is marked complete.

## Why
Lossless-claw was rerunning heavy startup backfills on every launch. That added avoidable startup work and still lacked an explicit completion signal for partial-upgrade recovery.

## Changes
- Add `lcm_migration_state` for versioned backfill completion
- Gate startup backfills by step name and version
- Wrap backfills plus state writes in savepoints
- Add repeat-startup and retry-safety migration tests
- Add patch changeset for the runtime fix

## Testing
- `npm test -- test/migration.test.ts`
- Expect `test/migration.test.ts` to pass

### DIFF
--- a/.changeset/startup-backfill-gating.md
+++ b/.changeset/startup-backfill-gating.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Stop rerunning startup summary and tool-call backfills after they complete successfully, while still retrying the same backfill version cleanly if startup fails before the completion marker is written.
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -43,6 +43,14 @@ type FtsTableSpec = {
   staleSchemaPatterns?: string[];
 };
 
+const VERSIONED_BACKFILL_STEPS = {
+  backfillSummaryDepths: 1,
+  backfillSummaryMetadata: 1,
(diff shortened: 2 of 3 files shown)


rmarr:
Confirming this fixes a real-world crash loop documented in #383.

**Environment:** lossless-claw v0.8.0, OpenClaw 2026.4.9, macOS arm64, ~22K messages / 510 summaries / 176MB database.

**Symptoms:** Every prompt triggered a 1–2 minute hang followed by `database is locked` → health check timeout → gateway restart. The unconditional `backfillSummaryDepths` and `backfillSummaryMetadata` DAG walks were blocking the event loop past the 10s health check threshold on every startup cycle.

**Validation:** A simpler version of this fix (early-return guards checking `COUNT(*) WHERE earliest_at IS NULL` and `COUNT(*) WHERE kind = 'condensed' AND depth = 0`) immediately resolved the crash loop — `chat.send` latency dropped from timeout to ~52ms. The `lcm_migration_state` table approach in this PR is a cleaner long-term solution with proper retry safety.

Looking forward to this shipping so I can drop the local patch. 🦞

PR #381, opened 2026-04-10
refactor: narrow bootstrap transaction scope

jalehman:
## What
Refactors bootstrap so transcript scanning and reconcile planning happen before the short write transaction, reducing the time lossless-claw holds `BEGIN IMMEDIATE` while preserving checkpoint, append-only, and replay-safety behavior.

## Why
Bootstrap previously kept the write transaction open across transcript file reads and reconcile analysis, which unnecessarily serialized unrelated database writers on the shared SQLite handle.

## Changes
- Split bootstrap into capture, plan, and apply phases
- Move transcript scanning outside write transaction
- Keep append-only and reconcile safeguards intact
- Preserve checkpoint and rotation cleanup behavior
- Add regression for concurrent transaction during planning

## Testing
- `npx vitest run test/engine.test.ts test/transaction-mutex.test.ts`
- Expected: both test files pass

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -60,7 +60,10 @@ import {
   type MessagePartRecord,
   type MessagePartType,
 } from "./store/conversation-store.js";
-import { SummaryStore } from "./store/summary-store.js";
+import {
+  SummaryStore,
+  type ConversationBootstrapStateRecord,
+} from "./store/summary-store.js";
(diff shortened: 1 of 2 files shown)


jalehman:
Closing this because current main already contains the equivalent bootstrap-prefix pruning guard and coverage after the context-pruning refactor, so there is no remaining landing delta here.

PR #102, opened 2026-03-16
feat: add idle compaction trigger (idleCompactMinutes)

drdotdot:
## Summary

Adds time-based compaction that triggers when a session has been idle beyond a configurable threshold (`idleCompactMinutes`). This addresses stale context in sessions that go through periods of inactivity — overnight, between work sessions, or during long-running background jobs.

## Motivation

Currently, LCM only compacts when context exceeds the `contextThreshold` (turn-based trigger). Sessions that accumulate significant context but never quite hit the threshold — or sessions that go idle for hours/days — can resume with stale, unorganized raw context. This is especially common in agentic workflows with periodic cron/watchdog messages.

Idle compaction ensures returning users get lean, organized context instead of a wall of raw messages from hours ago.

**Enables practical use of large context windows (1M tokens).** With models like Gemini offering 1M-token context at no pricing premium, the turn-based threshold trigger alone isn't sufficient — sessions can accumulate hundreds of thousands of tokens of raw conversation over hours without ever hitting the 75% threshold. Idle compaction provides the missing time-based dimension: after periods of inactivity, context is proactively organized into the summary DAG regardless of size, keeping large-window sessions navigable and performant over long lifespans.

## How it works

1. **Config:** `idleCompactMinutes` (env: `LCM_IDLE_COMPACT_MINUTES`, default: `0` = disabled)
2. **Detection:** `getLastActivityTimestamp()` queries the most recent `user` or `assistant` message. Heartbeats (never ingested) don't reset the timer. SystemEvents and sub-agent completions do.
3. **Leaf pass in assemble():** When idle threshold is exceeded, runs a single `compactLeaf()` with `force: true` before assembling context. Protected by a 15-second timeout to bound first-message latency.
4. **Full cascade in afterTurn():** Flags the session so `afterTurn()` forces a full compaction sweep regardless of context threshold, ensuring the leaf pass doesn't leave the DAG in a partial state.

## Changes

| File | Change |
|------|--------|
| `src/db/config.ts` | Added `idleCompactMinutes` to `LcmConfig` type and `resolveLcmConfig()` |
| `src/store/conversation-store.ts` | Added `getLastActivityTimestamp()` method |
| `src/compaction.ts` | Added `evaluateIdleTrigger()` method |
| `src/engine.ts` | Idle leaf pass in `assemble()`, forced cascade in `afterTurn()`, session tracking map |
| `openclaw.plugin.json` | Added config schema + UI hints |
| `docs/configuration.md` | Full documentation section |

## Configuration

[...]

PR #385, opened 2026-04-11
fix: move compaction from afterTurn to pre-assembly with cache-TTL check

### LINKED ISSUE #367: cache-aware compaction: timing inversion — decisions based on last-call status instead of cache expiry

liu51115:
## Summary

Moves all compaction decisions from `afterTurn()` to `assemble()` pre-assembly, using a persisted `lastApiCallAt` timestamp to determine whether the prompt cache has expired.

Closes #367. Supersedes #362, #363. Related: #102.

## Problem

`afterTurn()` evaluates compaction using the cache status of the call that just completed. This is a fundamental timing inversion: after any API call, the provider has just written (or refreshed) the prompt cache — **the cache is always hot in afterTurn, regardless of what the cache status reading says.** Compacting at this point always destroys a live cache.

The three removed branches all suffer from this:
- **hot-cache-budget-headroom**: defers when hot and under budget — but "hot" at afterTurn time is meaningless since the cache is always hot after a call
- **hot-cache-defer**: defers when pressure is low and cache is hot — same issue
- **cold-cache-catchup**: triggers aggressive multi-pass on cold readings — on any provider, a cold reading at afterTurn means the provider just wrote the cache, so it's now hot

On providers that load-balance across backend instances with separate prompt caches (e.g. OpenRouter), a secondary problem compounds this: random cold readings from routing misses create a cascade of aggressive compaction → prefix rewrite → guaranteed cold → repeat. But even on direct Anthropic, the core inversion applies — afterTurn is simply the wrong time to evaluate cache state.

## Fix

`afterTurn()` becomes telemetry-only: records `lastApiCallAt = Date.now()`, nothing else.

`assemble()` checks at pre-assembly time:
1. **Budget overflow** — must compact to fit, fires unconditionally
2. **Cache-TTL** — if `cacheAwareCompaction.enabled` and `now - lastApiCallAt > cacheTTLSeconds` and there's material to compact → compact before assembling

This is the only window where compaction is both safe and beneficial: the cache has genuinely expired, no live cache to destroy, and compacting reduces input cost on the upcoming cold call.

This approach supersedes the routing-noise guard in #362 (sticky cold counter is no longer needed when afterTurn doesn't make compaction decisions) and the deferred threshold compaction in #363 (all compaction is now off the reply path by design). It also addresses the same use case as the idle compaction trigger in #102 but with a simpler timestamp-based mechanism integrated into the assembly path.

## Changes

- `src/engine.ts`: remove two try/catch compaction blocks from `afterTurn()`, add pre-assembly block to `assemble()`, delete dead methods (`evaluateIncrementalCompaction`, `logIncrementalCompactionDecision`, `shouldApplyHotCacheHysteresis`, `resolveCacheAwareState`, `isComfortablyUnderTokenBudget`)
- `src/db/config.ts`: add `cacheTTLSeconds` to `CacheAwareCompactionConfig` (default 300, env `LCM_CACHE_TTL_SECONDS`)
- `src/db/migration.ts`: add `last_api_call_at INTEGER` column
- `src/store/compaction-telemetry-store.ts`: add `lastApiCallAt` field throughout
- `openclaw.plugin.json`: add `cacheTTLSeconds` to schema
- `test/engine.test.ts`: remove 11 tests for deleted afterTurn compaction behavior
- `test/config.test.ts`: update assertions

## Testing

All tests pass. Soak-tested in production: zero compactions during active use (turns 30–200s apart), compaction fires correctly after genuine cache expiry (idle > 300s), per-conversation TTL tracking confirmed working across multiple concurrent sessions.


100yenadmin:
There is genuine value here, but I think it is solving a different axis than #407 rather than replacing it.

I checked our live installed mitigation and the post-patch logs before commenting. The tactical local patch that stops leaf compaction and threshold sweep from both firing on the same turn is still active in our installed extension. In `/tmp/openclaw/openclaw-2026-04-12.log`, main-session `afterTurn` before that patch (11 samples before `2026-04-12 02:18:07 +07`) averaged about `11.8s` with an `83.0s` max; after the patch (58 samples after that timestamp), it dropped to about `456ms` average, `42.5ms` median, and `10.6s` max. So the hot-path split absolutely helped.

But it did **not** close the stall problem by itself. The same post-patch window still shows repeated `lane wait exceeded` events on `agent:main:main` and repeated `gateway timeout after 90000ms` subagent-announce retries. That is why I still think the deferred/background direction in #408 + the host companion PR is the right fit for #407: it addresses coalescing, visibility, and keeping proactive work off the foreground lane entirely.

What I do think is genuinely valuable in this PR:
- the cache-TTL / pre-assembly idea is a real cache-safety argument and worth considering on its own merits
- pinned-file injection is useful, but it feels orthogonal to the compaction-lane problem

So my read is: useful ideas, but not a drop-in replacement for #407/#408. If this lands, I’d strongly prefer it as a narrower follow-up or split-out PR instead of a competing end-to-end compaction strategy.

liu51115:
You're right about the pinned-file creep — that was a mistake on our end, we accidentally force-pushed a branch that bundled pinnedFiles (which has its own PR at #214) into this one. Just fixed it — the PR now contains only the cache-TTL pre-assembly change.

Agreed this is a different axis from #407/#408. The deferred debt + coalescing + visibility work in #408 addresses foreground blocking and starvation. This PR addresses cache-safety timing — making sure compaction doesn't destroy a hot provider cache by firing at the wrong moment.

They're complementary, not competing. Happy to keep this scoped as a narrower follow-up.

liu51115:
Closing in favor of #408 which now covers cache-TTL safety + deferred compaction + maintenance visibility. The per-agent cache TTL approach there (via runtime telemetry retention) supersedes our config-based approach.

PR #387, opened 2026-04-11
fix: refresh bootstrap checkpoint after afterTurn message ingestion

### LINKED ISSUE #386: Append-only bootstrap path always fails: DB message hash never matches JSONL checkpoint
## Environment

- OpenClaw: v2026.4.5
- lossless-claw: v0.8.0 (upgraded from v0.5.3)
- OS: Debian 12, LXC container, 1 core / 2 GB RAM
- Model: anthropic/claude-opus-4-6
- Session: ~2,500 messages, 4.4 MB JSONL, active for weeks

## Bug

Every real conversation turn triggers a full JSONL transcript read (~22 seconds on our hardware) instead of using the append-only fast path. Only heartbeat-only turns hit the fast checkpoint path (~8ms).

## Root cause

**Introduced in v0.7.0** (PR #329, commit 6579b91) when the checkpoint system was refactored to use `refreshBootstrapState()`.

In v0.5.3 and earlier, the checkpoint used simple file stat comparison:
```
lastSeenSize === sessionFileSize && lastSeenMtimeMs === sessionFileMtimeMs
```
This worked — if the JSONL file did not change between turns, the checkpoint hit.

v0.7.0 replaced this with `refreshBootstrapState()` which hashes the **last DB message** and compares it against the JSONL tail entry during the next bootstrap. The method is called in three places:
1. During bootstrap itself (after full import)
2. After heartbeat pruning
3. After `maintain()` (GC)

But it is **never called after `afterTurn()` message ingestion**. This is the only codepath that adds real conversation messages to the DB.

The result: `afterTurn` ingests new messages → DB frontier advances past the checkpoint hash → next bootstrap sees hash mismatch → falls back to full JSONL read. Every single turn.

Heartbeat-only turns are fast because `afterTurn` prunes heartbeat acks (zero messages ingested), so the DB frontier stays put and the checkpoint remains valid.

## Evidence from logs

Pattern over 4 hours of real usage:

| Turn type | Bootstrap path | Time |
|-----------|---------------|------|
| Conversation | Full reconcile | 22,441ms |
| Conversation | Full reconcile | 21,883ms |
| Conversation | Full reconcile | 22,105ms |
| Heartbeat | Checkpoint hit | 8ms |
| Conversation | Full reconcile | 21,997ms |
| Conversation | Full reconcile | 22,308ms |

Every conversation turn: full read. Every heartbeat: fast path. No exceptions.

## Fix

Call `refreshBootstrapState()` at the end of `afterTurn()` after message ingestion, before the `afterTurn: done` log line (~line 3512). The method, params, and plumbing all exist — it is already called after heartbeat pruning in the same function. It just needs to be called on the non-heartbeat path too.

## Regression scope

- **v0.5.3 and earlier**: Not affected (used file stat checkpoint, no hash verification)
- **v0.7.0+**: Affected (append-only fast path never works after real conversation turns)
- **Every persistent LCM session** hits this on every real turn. On fast hardware with small sessions it may only add a few seconds. On constrained hardware with large sessions it is 20+ seconds per turn.

oguzbilgic:
## Problem

The append-only bootstrap fast path (introduced in v0.7.0, PR #329) never works after real conversation turns. Every turn triggers a full JSONL transcript read instead.

**Root cause:** `refreshBootstrapState()` is called after heartbeat pruning and after `maintain()`, but never after regular message ingestion in `afterTurn()`. After ingesting new messages, the DB frontier advances past the stored checkpoint hash. The next bootstrap sees the mismatch and falls back to a full reconcile.

On a 4.4 MB session JSONL with constrained hardware (1-core LXC), this adds **~22 seconds per real conversation turn**. Heartbeat-only turns are unaffected because pruning ingests zero messages and the checkpoint stays valid.

## Fix

Add a `refreshBootstrapState()` call after successful message ingestion in `afterTurn()`, before compaction. The method, params, and plumbing all exist — it is already called on the heartbeat pruning path in the same function.

**+8 lines, no new methods or interfaces.**

## Evidence

Logs from a persistent session (v0.8.0, ~2,500 messages):

| Turn type | Bootstrap path | Time |
|-----------|---------------|------|
| Conversation | Full reconcile | 22,441ms |
| Conversation | Full reconcile | 21,883ms |
| Heartbeat | Checkpoint hit | 8ms |
| Conversation | Full reconcile | 21,997ms |

100% of real conversation turns hit the slow path. 100% of heartbeat-only turns hit the fast path.

## Regression scope

- **v0.5.3 and earlier**: Not affected (used file-stat-only checkpoint, no hash verification)
- **v0.7.0+**: Affected on every real conversation turn in every persistent session

Fixes #386

### DIFF
--- a/.changeset/warm-gifts-pull.md
+++ b/.changeset/warm-gifts-pull.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Refresh the bootstrap checkpoint after normal `afterTurn()` ingestion so persistent sessions can keep using the append-only bootstrap fast path after real conversation turns.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3546,6 +3546,17 @@ export class LcmContextEngine implements ContextEngine {
     }
 
     try {
+      await this.refreshBootstrapState({
+        conversationId: conversation.conversationId,
+        sessionFile: params.sessionFile,
+      });
+    } catch (err) {
+      this.deps.log.warn(
+        `[lcm] afterTurn: bootstrap checkpoint refresh failed for ${sessionLabel}: ${describeLogError(err)}`,
+      );
+    }
+
+    try {
       const rawLeafTrigger = await this.compaction.evaluateLeafTrigger(conversation.conversationId);
       await this.updateCompactionTelemetry({
         conversationId: conversation.conversationId,
--- a/test/engine.test.ts
+++ b/test/engine.test.ts
@@ -2668,6 +2668,65 @@ describe("LcmContextEngine.bootstrap", () => {
     expect(reconcileSpy).not.toHaveBeenCalled();
   });
 
+  it("refreshes the bootstrap checkpoint after a normal afterTurn before the next append-only bootstrap", async () => {
+    const sessionFile = createSessionFilePath("append-only-after-turn-normal-ingest");
+    const sm = SessionManager.open(sessionFile);
+    sm.appendMessage({
+      role: "user",
+      content: [{ type: "text", text: "seed user" }],
+    } as AgentMessage);
+    sm.appendMessage({
+      role: "assistant",
+      content: [{ type: "text", text: "seed assistant" }],
+    } as AgentMessage);
+
+    const engine = createEngine();
+    const sessionId = "bootstrap-append-only-after-turn-normal-ingest";
+
+    const first = await engine.bootstrap({ sessionId, sessionFile });
+    expect(first.bootstrapped).toBe(true);
+
+    const realTurn = [
+      makeMessage({
+        role: "user",
+        content: "new user",
+      }),
+      makeMessage({
+        role: "assistant",
+        content: "new assistant",
+      }),
+    ];
+    for (const message of realTurn) {
+      sm.appendMessage(message as AgentMessage);
+    }
+
+    await engine.afterTurn({
+      sessionId,
+      sessionFile,
+      messages: realTurn,
+      prePromptMessageCount: 0,
+      tokenBudget: 4096,
+    });
+
+    sm.appendMessage({
+      role: "user",
+      content: [{ type: "text", text: "tail user" }],
+    } as AgentMessage);
+    sm.appendMessage({
+      role: "assistant",
+      content: [{ type: "text", text: "tail assistant" }],
+    } as AgentMessage);
+
+    const reconcileSpy = vi.spyOn(engine as any, "reconcileSessionTail");
+    const second = await engine.bootstrap({ sessionId, sessionFile });
+    expect(second).toEqual({
+      bootstrapped: true,
+      importedMessages: 2,
+      reason: "reconciled missing session messages",
+    });
+    expect(reconcileSpy).not.toHaveBeenCalled();
+  });
+
   it("falls back to full reconciliation when append-only checkpoint validation mismatches", async () => {
     const tempDir = mkdtempSync(join(tmpdir(), "lossless-claw-engine-"));
     tempDirs.push(tempDir);


GodsBoy (commented):
## Review: fix/refresh-bootstrap-after-ingest

### Placement: correct

The call is positioned after:
1. `ingestBatch()` succeeds (line ~3470 on main) — if ingest fails, `afterTurn` returns early, so the checkpoint is never stale-advanced.
2. The heartbeat-pruning early-return (line ~3512) — heartbeats that prune >0 messages already call `refreshBootstrapState` and return. If `pruned === 0`, or the batch is not a heartbeat, execution falls through.
3. The `conversation` lookup guard (line ~3540) — `conversation` is guaranteed non-null.

And it runs before compaction evaluation, which is the correct order: checkpoint must reflect the new frontier before we decide whether to compact.

### Error handling: solid

The `try/catch` with `warn`-level logging matches the exact pattern used in the heartbeat-pruning path (line ~3504) and the `maintain()` path. A checkpoint refresh failure is non-fatal — it just means the *next* bootstrap falls back to full reconcile, which is the pre-v0.7.0 behavior. Compaction is not blocked.

### Codepath coverage analysis

| afterTurn codepath | refreshBootstrapState called? | Notes |
|---|---|---|
| Empty batch (dedup clears all) | N/A — returns before ingest | No DB mutation, checkpoint still valid |
| Ingest failure | N/A — returns before this code | No DB mutation, checkpoint still valid |
| Heartbeat with pruned > 0 | Yes — existing call | Returns early, new code not reached (correct) |
| Heartbeat with pruned == 0 | **Yes — new code** | Falls through to non-heartbeat path, now fixed |
| Heartbeat with no conversation | N/A — falls through, lookup misses, returns | No meaningful frontier change |
| Normal conversation turn | **Yes — new code** | The primary fix |

This looks comprehensive. The one subtle case worth noting: a heartbeat-looking batch where `pruned === 0` still ingested messages via `ingestBatch()` above, so it does advance the frontier and does need the refresh. The fix covers this correctly.

### Missing: regression test

There is no test included. The existing test suite does not appear to cover `refreshBootstrapState` at all (no hits in `*.test.*` files). A regression test would strengthen confidence, though the fix is mechanically simple (same pattern already used 4 other places in the same file). Consider adding one as a follow-up — something like:

1. Ingest a batch via `afterTurn()` on a session with a valid bootstrap checkpoint.
2. Assert that the bootstrap state's frontier hash matches the post-ingest DB state.
3. Verify the next `bootstrap()` call takes the fast path (no full JSONL read).

### Minor nit

The `sessionLabel` used in the log message (`afterTurn: bootstrap checkpoint refresh failed for ${sessionLabel}`) is the `session=X sessionKey=Y` format, whereas the heartbeat-pruning path uses `sessionContext` (which includes `conversationId`). Including `conversationId` in the log line would make debugging easier, since by this point `conversation` is resolved. Not blocking.

### Verdict

The fix is correct, minimal, and follows existing patterns. Approving pending the author's call on whether to add a regression test in this PR or as a follow-up.

jalehman:
Thank you!

DBostik:
Will this fix the 10+ second pause after every message were the gateway logs freeze and then after 4-6 seconds it floods the logs with 6+ LCM messages every turn? 

After v0.8.0 I noticed in telegram my agent will show the thinking animation immediately after being messaged but then it stops for 10+ seconds, then comes back until it responds later. 

Seems to add some latency every turn.

jalehman:
Yes, this should fix that. It's now out in the latest release.

DBostik:
> Yes, this should fix that. It's now out in the latest release.

Wow, what a massive improvement. Thanks!

PR #388, opened 2026-04-11
Add ollama provider to inferApiFromProvider()

bennybuoy:
**Problem**  LCM fails to resolve API family for the ollama provider:  [lcm] unable to resolve API family for provider ollama; set models.providers.ollama.api explicitly instead of falling back implicitly.  **Trigger**  Setting summaryModel to any ollama-served model (e.g., ollama/glm-5.1, ollama/kimi-k2.5:cloud).  **Fix**  Add mapping ollama: openai-completions since Ollama uses OpenAI-compatible API.  **File**  src/plugin/index.ts, line ~553 in inferApiFromProvider()

### DIFF
--- a/.changeset/pr388-ollama-provider-fallback.md
+++ b/.changeset/pr388-ollama-provider-fallback.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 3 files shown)


jalehman:
Thank you!

Issue #391, opened 2026-04-11
[Perf] Stream-based byte-offset tracking for O(new_messages) bootstrap


GodsBoy:
## Problem

The current bootstrap has three paths:
1. **Checkpoint hit** (file metadata unchanged) — O(1) ✓
2. **Append-only fast path** (file grew, hashes match) — O(new_messages) ✓
3. **Full read fallback** — O(total_messages) ✗

When path 2 fails (hash mismatches, format changes, etc.), the fallback reads the ENTIRE JSONL file line-by-line. For a 14MB / 2,500+ message file, this takes ~22-50 seconds and typically imports 0 messages.

The `readAppendedLeafPathMessages()` function already supports reading from a byte offset — the infrastructure for incremental reads exists. The issue is that it's gated behind hash verification checks that frequently fail.

## Suggested Approach

Track the byte offset of the last processed JSONL line independently of the hash-based checkpoint:

1. After any successful bootstrap (full or incremental), persist the byte offset of the last processed entry
2. On the next bootstrap, if the file has grown, seek to the stored offset and read only new lines
3. Use the offset as the primary fast-path signal, with hash verification as a secondary integrity check rather than a gate

This makes bootstrap O(new_messages) in the common case, even when hash verification fails.

## Scope

This is a longer-term improvement — not a bug fix. The immediate performance regression is addressed by #387 and the checkpoint persistence fix. This issue tracks the architectural improvement to make bootstrap resilient to hash mismatches.

## Relationship

- Part of the bootstrap performance investigation (#386)
- Related to the performance sprint (#297)
- Builds on the existing `readAppendedLeafPathMessages` infrastructure

100yenadmin:
@GodsBoy triage pass update: I marked this as priority:P3 performance enhancement. Stream-based byte-offset tracking could help large bootstraps, but I am keeping it behind current correctness bugs. Can you confirm whether O(new_messages) bootstrap is still a current bottleneck?

Issue #390, opened 2026-04-11
[Bug] Bootstrap checkpoint not persisted when reconcileSessionTail finds no overlap


GodsBoy:
## Problem

In `bootstrap()`, after a full read and `reconcileSessionTail()`, the checkpoint is only persisted conditionally:

```typescript
if (reconcile.hasOverlap) {
  await persistBootstrapState(conversationId);
}
```

When `reconcileSessionTail` returns `hasOverlap: false` (e.g., the last historical message doesn't match any DB message due to hash differences), the checkpoint is never saved. This means subsequent bootstraps cannot use the checkpoint fast path and will always fall through to a full read.

## Expected Behavior

`persistBootstrapState` should always be called after a successful full read, regardless of the overlap result. The full read has already completed — the checkpoint should reflect the current file state so the next bootstrap can skip the read.

## Impact

On sessions where overlap detection fails (e.g., after version upgrades that change token estimation or content normalization), every bootstrap does a full JSONL read indefinitely — the checkpoint never "catches up."

## Suggested Fix

Move `persistBootstrapState` outside the `hasOverlap` conditional:

```typescript
// After reconcileSessionTail:
await persistBootstrapState(conversationId, historicalMessages);
```

## Relationship

- Supplements the primary fix in #387
- Addresses a different failure mode than #386 (checkpoint loss vs checkpoint staleness)

100yenadmin:
@GodsBoy triage pass update: this looks likely resolved by merged PR #392, which added the file-level bootstrap cache guard for the repeated full-read/checkpoint-miss family and explicitly lists #390 as related hardening. I am marking this `possibly-resolved` + `linked-pr` + `needs-repro` rather than closing. Can you confirm whether current `main` / a release containing #392 still reproduces the checkpoint-not-persisted behavior?

jalehman:
Closed by #1032. The landed fix uses the existing non-anchoring-frontier proof, imports the readable transcript before checkpointing, and deliberately keeps real divergent histories fail-closed.

PR #392, opened 2026-04-11
fix: add file-level bootstrap cache guard as defensive fallback

### LINKED ISSUE #389: [Perf] Add file-level cache guard to skip redundant full bootstrap reads when JSONL unchanged
## Problem

When the append-only fast path in `bootstrap()` fails (hash mismatch, file unchanged, etc.), the full `readLeafPathMessages()` fallback reads the entire session JSONL file — even if a previous full read already processed the same file content.

On a 14MB / 2,500+ message session, this costs ~22-50 seconds per call and imports 0 messages. Evidence from production logs:

```
[lcm] bootstrap: full transcript read conversation=213 ... duration=48322ms
[lcm] bootstrap: done ... importedMessages=0 reason=already bootstrapped duration=49382ms
```

## Suggested Fix

Cache the result of the last successful full read by file `mtime` + `size` on the engine instance:

```typescript
private lastFullReadState = new Map<number, { size: number; mtimeMs: number }>();
```

Before calling `readLeafPathMessages()`, check: if `conversation.bootstrappedAt` is set, existing messages exist, and the file metadata matches the cached state, skip the full read entirely.

This is a defensive layer — even if the checkpoint fast path and append-only fast path both fail, this prevents redundant full reads of unchanged files.

## Relationship

- Supplements the primary fix in #387 (missing `refreshBootstrapState` after `afterTurn`)
- Related to the performance sprint tracked in #297
- Evidence gathered from the same investigation as #386

GodsBoy:
## Context

The primary fix for #386 (missing `refreshBootstrapState()` call after `afterTurn` ingestion) landed in #387. This PR has been rebased on top of that merge and now only contains the second, independent layer of protection.

## Problem

Even with #387 merged, the append-only fast path can still fall through to a full `readLeafPathMessages()` call under conditions that are invisible to the checkpoint hash:

- Token estimator drift across versions invalidating the stored hash
- Hash format changes or corruption
- Any edge case where the checkpoint hash stops matching a byte-identical file

When this happens on a large (10+ MB, 2,500+ message) session, every bootstrap re-reads the entire transcript — ~22-50 seconds per call — even though the file on disk has not changed and zero new messages will be imported. Evidence from production logs:

```
[lcm] bootstrap: full transcript read conversation=213 ... duration=48322ms
[lcm] bootstrap: done ... importedMessages=0 reason=already bootstrapped duration=49382ms
```

## Fix

Add an in-memory `Map<conversationId, { size, mtimeMs }>` that tracks file metadata from the last successful full read. Before falling through to `readLeafPathMessages()`, check: if the conversation is already bootstrapped and the file metadata matches the cached state, skip the full read and just refresh the bootstrap state.

This is a **defence-in-depth layer** on top of #387. #387 keeps the happy path fast; this guard prevents pathological cases from silently regressing to 20-50s full reads.

### Cache invalidation

- Set only after successful full reads (never after import-capped aborts)
- Cleared on session file rotation via `purgeConversationForBootstrapRotation`
- In-memory only — cleared on engine restart (worst case: one extra full read on startup)

## Why this is still worth merging after #387

| Scenario | #387 alone | #387 + this PR |
|---|---|---|
| Normal real-turn bootstrap | ✅ Fast path (checkpoint hit) | ✅ Fast path (checkpoint hit) |
| Checkpoint hash mismatch on unchanged file | ❌ Full re-read (~30s) | ✅ Cache guard skips full read |
| Version upgrade that drifts token estimator | ❌ Full re-read per session | ✅ Single full read, then cached |
| Engine restart | ⚠️ One full read per session (expected) | ⚠️ One full read per session (expected) |

The checkpoint hash is a *correctness* signal ("did the DB frontier change"). The file (size, mtime) is an *identity* signal ("did the source of truth change"). Both signals agreeing is a much stronger skip condition than either alone, and it's the only one that survives hash-format changes.

## Tests

Two regression tests cover the cache guard:

- `skips full read when file is unchanged and conversation is already bootstrapped` — verifies the guard fires when the file metadata matches
- `file-level cache guard allows full read when file changes` — verifies the guard correctly permits full reads when the file grows

The `refreshes bootstrap checkpoint after afterTurn` test from the original PR was removed because #387 already landed equivalent coverage.

All 146 engine tests pass.

## Related

- Closes #389
- Builds on #387 (same problem, different layer)
- Related hardening issues: #390, #391
- Part of performance sprint #297

## Post-Deploy Monitoring & Validation

- **Log query (guard active):** `bootstrap: skipped full read (file unchanged)` — presence confirms the cache guard is firing on unchanged files
- **Log query (fallback):** `bootstrap: full transcript read` — frequency should drop toward zero on persistent sessions; any sustained presence after #387 + this PR indicates a new bootstrap path regression
- **Expected healthy signal:** Bootstrap duration <1s on unchanged large sessions, even when the append-only fast path misses
- **Failure signal:** If bootstrap duration stays in the 20-50s range on unchanged sessions, investigate whether cache invalidation is firing too aggressively
- **Validation window:** 24 hours of normal activity on sessions with 500+ messages and >5MB JSONL files

### DIFF
--- a/.changeset/four-pandas-invent.md
+++ b/.changeset/four-pandas-invent.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Avoid repeated full bootstrap rereads when an unchanged session transcript misses the normal checkpoint fast paths.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1308,6 +1308,14 @@ export class LcmContextEngine implements ContextEngine {
   private largeFileTextSummarizer?: (prompt: string) => Promise<string | null>;
   private deps: LcmDependencies;
 
+  /**
+   * Tracks file metadata from the last successful full bootstrap read per
+   * conversation. When the session JSONL file has not changed since the last
+   * full read and the conversation is already bootstrapped, the expensive
+   * readLeafPathMessages() call can be skipped entirely.
+   */
+  private lastFullReadFileState = new Map<number, { size: number; mtimeMs: number }>();
+
   // ── Circuit breaker for compaction auth failures ──
   private circuitBreakerStates = new Map<string, CircuitBreakerState>();
 
@@ -3590,6 +3598,12 @@ export class LcmContextEngine implements ContextEngine {
                 mtimeMs: sessionFileMtimeMs,
               },
             });
+            // Update the file-level cache so subsequent bootstraps against an
+            // unchanged file can skip the full read via the cache guard.
+            this.lastFullReadFileState.set(conversationId, {
+              size: sessionFileSize,
+              mtimeMs: sessionFileMtimeMs,
+            });
           };
 
           const conversation = await this.conversationStore.getOrCreateConversation(params.sessionId, {
@@ -3606,6 +3620,12 @@ export class LcmContextEngine implements ContextEngine {
             this.deps.log.warn(
               `[lcm] bootstrap: session file rotated conversation=${conversationId} ${sessionLabel} oldFile=${bootstrapState.sessionFilePath} newFile=${params.sessionFile}`,
             );
+            // A rotated session file invalidates every piece of cached state
+            // keyed to the old path: the on-disk bootstrap checkpoint row, the
+            // in-memory file-level guard, and any counters derived from the
+            // old file's messages. Clear them all in one place so subsequent
+            // reads treat this conversation as unbootstrapped.
+            this.lastFullReadFileState.delete(conversationId);
             bootstrapState = null;
           }
 
@@ -3712,6 +3732,28 @@ export class LcmContextEngine implements ContextEngine {
                   reason: conversation.bootstrappedAt ? "already bootstrapped" : "conversation already up to date",
                 };
               }
+            }
+          }
+
+          // File-level cache guard: if the conversation is already bootstrapped
+          // and the JSONL file has not changed since the last successful full read,
+          // skip the expensive readLeafPathMessages entirely.
+          if (conversation.bootstrappedAt && existingCount > 0) {
+            const cached = this.lastFullReadFileState.get(conversationId);
+            if (
+              cached &&
+              cached.size === sessionFileSize &&
+              cached.mtimeMs === sessionFileMtimeMs
+            ) {
+              await persistBootstrapState(conversationId);
+              this.deps.log.info(
+                `[lcm] bootstrap: skipped full read (file unchanged) conversation=${conversationId} ${sessionLabel} duration=${formatDurationMs(Date.now() - startedAt)}`,
+              );
+              return {
+                bootstrapped: false,
+                importedMessages: 0,
+                reason: "already bootstrapped",
+              };
             }
           }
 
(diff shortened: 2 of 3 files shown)


GodsBoy:
@jalehman — rebased on top of `main` now that #387 has merged. The duplicate checkpoint-refresh commit is gone; this PR has been narrowed to just the file-level cache guard.

TL;DR on why it's still worth merging as a separate, smaller PR:

- #387 fixes the **common** case: `afterTurn` now keeps the checkpoint aligned with the DB frontier, so the append-only fast path works on every real turn.
- This PR fixes the **pathological** case: when the checkpoint hash stops matching a byte-identical file (token-estimator drift across versions, hash-format changes, corruption), bootstrap silently falls through to a 20-50s full `readLeafPathMessages()`. The `(size, mtime)` guard short-circuits that, so the worst case becomes "one full read after an upgrade" instead of "a full read every turn".

The two signals are doing different jobs:

- **Checkpoint hash** = "did the DB frontier change" (correctness)
- **File (size, mtime)** = "did the source of truth change" (identity)

Either alone is enough for the happy path, but only the combination survives a hash-format change without regressing performance. It's ~30 lines + 2 focused tests and no new surface area.

If you'd rather land this as part of a broader hardening sweep (there are related issues #390 and #391 from the same investigation), happy to fold them together instead — just let me know which way you'd prefer.

GodsBoy:
Ran a full review pass against the rebased branch and caught two bugs I introduced during the rebase resolution. Fixed and re-pushed:

**1. `afterTurn` `refreshBootstrapState` was in the wrong position.** When I resolved the rebase conflict, I kept the copy from the original PR (earlier, before the heartbeat-ack short-circuit) and deleted the copy from #387 (later, after the conversation lookup miss check). Net effect: I was silently relocating #387's code into a new position. Reverted — the diff against `main` now shows zero changes to `afterTurn`.

**2. `purgeConversationForBootstrapRotation` became dead code.** The original PR had a caller at the rotation-detection branch in `bootstrap()`, but the rebase auto-merge dropped it and I didn't notice. Rather than restore a 50-line cascade-delete method with a single use site, I inlined the one line that matters for this PR — `this.lastFullReadFileState.delete(conversationId)` — directly at the rotation branch. The invariant "cache is cleared on rotation" is now load-bearing in code rather than coincidentally true via a downstream overwrite.

**Also applied:**

- Tightened the "file changes" test assertion from `toBeGreaterThanOrEqual(0)` (vacuous) to `toBeGreaterThanOrEqual(1)`, and added a negative assertion that the "skipped full read" log did **not** fire. Regression that caused the full read to return 0 messages would now actually fail.
- Added an explanatory comment on the rotation-branch invalidation so the reason the cache-clear belongs there is obvious to the next reader.

**Final diff:** 42 lines in `src/engine.ts`, 151 in `test/engine.test.ts`, down from 250/11 previously. All 146 engine tests pass.

The remaining review findings are acknowledged but not acted on in this PR:

- The guard still calls `persistBootstrapState` on cache-hit, which does one extra DB SELECT + UPSERT. Measurable but the 20-50s readLeafPathMessages skip dominates. Happy to remove it in a follow-up if you'd like the fast path to be truly zero-I/O.
- Unbounded growth of `lastFullReadFileState`. ~40 bytes per conversation, P3 at worst. Can add an LRU cap if desired.
- More negative tests (engine restart, `existingCount === 0`, `bootstrappedAt === null`) would lock the gate more tightly.

Let me know if you want any of those folded in.

jalehman:
Thank you!

PR #393, opened 2026-04-11
fix: skip already-completed backfill migrations on startup

bennybuoy:
## Problem

Three migration backfill functions (`backfillSummaryDepths`, `backfillSummaryMetadata`, `backfillToolCallColumns`) run **unconditionally on every gateway startup**. On large databases (2K+ conversations, 70K+ messages), these O(conversations x summaries) operations:

1. Take minutes to complete
2. Hold SQLite write locks for the duration
3. Cause the gateway to fail health checks during startup
4. If the process is SIGKILL'd mid-backfill -> restart loop -> **startup death spiral**

We observed this on a 580MB database (2,135 conversations, 73,671 messages). The gateway would start, consume 3.2GB RAM and 106% CPU on backfills, get SIGKILL'd by systemd after the health check timeout, and repeat indefinitely.

The external lcm-bridge (read-only, `immutable=1`) was initially suspected but was not the cause - it holds no write locks.

## Fix

Add cheap sentinel queries before each backfill:

- **Depth backfill**: Skip if no leaf has `depth != 0` and no condensed has `depth = 0` (single `SELECT 1 ... LIMIT 1` query)
- **Metadata backfill**: Skip if no summary has `earliest_at IS NULL`
- **Tool call backfill**: Skip if no rows have `tool_call_id IS NULL AND metadata IS NOT NULL` with extractable values

On an already-migrated DB, startup drops from **minutes** to **milliseconds**. The sentinel queries use indexed columns and return immediately when data is populated.

When new data arrives that *does* need backfilling (e.g., a new conversation with summaries missing depth/metadata), the sentinel finds rows and the full backfill runs as before - so incremental migrations still work correctly.

## Testing

Added 3 test cases in `migration.test.ts`:
- Verifies depth backfill is skipped on a fully-migrated DB
- Verifies metadata backfill is skipped when `earliest_at` is populated
- Verifies tool call backfill is skipped when no legacy rows need extraction

## Changeset

Patch - this is a fix for startup performance, no API or behavior changes for correctly-migrated databases.

### DIFF
--- a/.changeset/skip-completed-backfills.md
+++ b/.changeset/skip-completed-backfills.md
@@ -1,0 +1,9 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Skip already-completed backfill migrations on startup
+
+Three migration backfill functions (`backfillSummaryDepths`, `backfillSummaryMetadata`, `backfillToolCallColumns`) ran unconditionally on every gateway startup. On large databases (2K+ conversations, 70K+ messages), these O(conversations × summaries) operations took minutes, held SQLite write locks, and caused the gateway to fail health checks. If the process was killed mid-backfill, it restarted and ran the same expensive operations again — creating a startup death spiral.
+
+This change adds cheap sentinel queries before each backfill: if no rows need updating, the backfill is skipped entirely with an info log. Startup on an already-migrated DB drops from minutes to milliseconds.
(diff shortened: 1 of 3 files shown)


jalehman:
Closing — the underlying issue is now fixed by #380, which landed on main April 11.

#380 takes a similar direction (gate startup backfills so they don't rerun) but uses a versioned migration-state table to track completion. Since that's already merged and covers the same ground, this PR is no longer needed.

Thanks for jumping on this — the problem was real and your diagnosis helped move things forward.

PR #403, opened 2026-04-11
fix(engine): move bootstrap file I/O off the event loop

jetd1:
## Summary

- Convert `readFileSegment`, `readLastJsonlEntryBeforeOffset`, and the `statSync` calls in `bootstrap()` / `refreshBootstrapState()` to `fs/promises`, so a multi-MB session JSONL can no longer block the Node.js event loop for minutes during the append-only bootstrap path (freezing every gateway session, Control UI, CLI, and WebSocket connection).
- In `readLastJsonlEntryBeforeOffset`, only read a new chunk when the current carry has no more newlines to peel. The original re-read a chunk on every iteration even while still processing existing lines, which wasted I/O and amplified the implicit O(n^2) prepended-carry pattern.
- Short-circuit the append-only fast path before the backward scan when `latestDbHash !== bootstrapState.lastProcessedEntryHash`. That is the common case during active sessions — the DB frontier advances past the checkpoint between bootstraps, and the matcher can never find a matching tail entry in that state, so we skip straight to the async full-read slow path.

A patch-level changeset is included per `RELEASING.md`.

## Test plan

- [x] `npm test` — full vitest suite passes (695 tests, 40 files) on this branch
- [x] `npx tsc --noEmit` — no new type errors introduced in touched files (unrelated pre-existing errors elsewhere are unchanged)
- [x] `bootstrap-message-only.test.ts` updated to `await` the now-async `readLastJsonlEntryBeforeOffset`; all 12 cases still pass
- [ ] Soak test on a real large session: bootstrap no longer freezes the gateway; a `setTimeout` callback fires within ~100 ms while bootstrap is reconciling
- [ ] Maintainer review of the fast-path short-circuit to confirm no correctness regression when checkpoint hashes legitimately match

### DIFF
--- a/.changeset/async-bootstrap-sync-io.md
+++ b/.changeset/async-bootstrap-sync-io.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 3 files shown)


jalehman:
Thank you!

PR #405, opened 2026-04-11
fix(engine): guard bootstrap against missed /reset lifecycle split

uf-hy:
After /reset, bootstrap can silently reattach to the previous active conversation instead of starting a new one.

This happens when the new session starts before the lifecycle split has been recorded. Bootstrap sees an existing active conversation for the same sessionKey and reuses it, even though the sessionId changed and the old transcript file is gone.

The fix adds a guard in bootstrap, before `getOrCreateConversation`: when the stored sessionId and sessionFile don't match the incoming ones, and the old tracked transcript path no longer exists, bootstrap treats it as a missed reset and rotates the conversation via the existing `applySessionReplacement` path.

Normal transcript rotation is not affected — the old file still exists in that case.

Tested locally:
- Normal /reset lifecycle handling
- Ordinary transcript rotation (old file still present)
- Missed-reset case (old transcript gone, new sessionId on same sessionKey)

### DIFF
--- a/.changeset/green-squids-wink.md
+++ b/.changeset/green-squids-wink.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Restrict the missed-`/reset` bootstrap fallback to confirmed missing transcript paths so transient `stat()` failures do not rotate a live conversation.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3003,6 +3003,42 @@ export class LcmContextEngine implements ContextEngine {
             });
           };
 
+          // Guard: when a sessionKey resumes on a new sessionId and the tracked
+          // transcript file has disappeared, treat it as a missed /reset and
+          // rotate the conversation before getOrCreate would re-attach to it.
+          const normalizedSessionKey = params.sessionKey?.trim();
+          if (normalizedSessionKey) {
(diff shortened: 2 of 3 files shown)


jalehman:
Thank you!

OctavianTocan:
Running into this in the wild on v0.9.1. A direct-chat session on one of my agents got into the state this PR describes: a fresh conversation row exists in `conversations` but has no matching row in `conversation_bootstrap_state`, and its stored `session_id` no longer matches the active session id the runtime has in `sessions.json`. The runtime is happily writing to the current session transcript, but LCM never polls it, so zero messages ingest for that conversation.

Rough timeline of what I saw:

1. Session runs normally for a while, bootstrap_state present, poller advancing as expected.
2. User runs `/reset`. A new sessionId is minted, a fresh transcript file starts.
3. A new conversation row is created, but no bootstrap_state row is ever written for it.
4. After one or more subsequent rotations, `sessions.json` points at a newer sessionId than the conversation row stores. LCM and the runtime disagree on which session is current.
5. From then on the active transcript keeps growing but is never read. Conversation stays at zero messages, no telemetry, no compaction. The agent presents as amnesic, which nudges the user to `/reset` again, which re-triggers the same race.

Since this PR is merged but not yet in a tagged release (latest is v0.9.1 from April 14, this merged April 16), would it be possible to cut a patch release with the fix? Happy to verify against my setup once there's something to install.

Issue #407, opened 2026-04-12
Proactive threshold compaction can stall foreground turns and hide maintenance state


100yenadmin:
## Problem

`lossless-claw` currently runs proactive threshold compaction on the foreground turn path. In large or busy sessions, that can block the active user turn long enough to create long lane waits, make the main agent feel unresponsive, and delay subagent completion announce behind compaction work.

This is most visible in two real usage patterns:
- Worst case: one orchestrator plus up to 4 subagents, with both write-mode and read-mode LCM activity competing for the same session lane.
- Minimum heavy case: one orchestrator plus up to 4 agents reading LCM concurrently while proactive maintenance is queued or running.

The current behavior also hides maintenance state from users, so when compaction is happening they see only a stalled or slow turn, not a clear indication of what the system is doing.

## Goal

Make proactive compaction debt-driven and visible, without allowing it to block foreground turns. Manual compaction, overflow recovery, and timeout recovery should remain synchronous when correctness requires it.

## Acceptance Criteria

- Proactive threshold compaction no longer runs inline in foreground `afterTurn`.
- Compaction requests are coalesced per session/conversation instead of stacking.
- The plugin tracks and exposes maintenance state:
  - pending
  - running
  - last success
  - last failure
- Read-only LCM operations remain usable while compaction debt is pending or running.
- Subagent announce flow is no longer starved by proactive compaction.
- Manual, overflow, and timeout recovery compaction remain synchronous.
- Users get subtle visibility only when maintenance is long-running or fails.

## Validation Notes

- Reproduce with an orchestrator plus multiple agents/subagents under mixed read/write pressure.
- Verify the foreground turn returns promptly while compaction debt is pending.
- Verify the system surfaces maintenance state without turning it into noisy chat output.
- Verify manual/overflow/timeout compaction still behaves synchronously.

liu51115:
We've been running a variant of this fix in production for ~2 days. Our approach:

**Core change:** Move all proactive compaction out of `afterTurn()` into pre-assembly in `assemble()`. `afterTurn` becomes telemetry-only (records cache state + `lastApiCallAt` timestamp).

**Pre-assembly triggers (in `assemble()`):**
1. **Budget overflow** — must compact to fit, runs synchronously before assembly
2. **Cache-TTL** — if `Date.now() - lastApiCallAt > cacheTTLSeconds` AND `evaluateLeafTrigger` fires, compact before assembly

[...]

PR #410, opened 2026-04-12
feat(engine): budget-aware compaction deferral via budgetHeadroomRatio

### LINKED ISSUE #409: feat: budget-aware compaction deferral for unknown cache state
## Problem

The incremental leaf compaction trigger (`evaluateLeafTrigger`) fires based on a fixed token count (`leafChunkTokens`, default 20k) regardless of how much context budget remains. On a 1M-token-window model, leaf digests start being created at the same conversation point as on a 200k-window model — even though the larger window could comfortably hold all raw messages for much longer.

The cache-aware compaction system already addresses this for `cacheState === "hot"` via the `isComfortablyUnderTokenBudget` check gated by `hotCacheBudgetHeadroomRatio` (default 0.2). However, when `cacheState === "unknown"` — which happens **permanently** for non-Anthropic providers that never report cache telemetry, and transiently on the first turn of any conversation — the leaf trigger fires unconditionally once raw tokens outside the fresh tail exceed `leafChunkTokens`.

This means:
- Every non-Anthropic deployment pays summarization LLM costs from early in the conversation, even with 95%+ budget headroom
- The deferral optimization is effectively Anthropic-provider-only

## Current decision tree (`evaluateIncrementalCompaction`)

```
1. leafTrigger fires?              NO  → skip ("below-leaf-trigger")
                                   YES ↓
2. budgetDecision fires?           YES → compact + condensed ("budget-trigger")
                                   NO  ↓
3. hot cache + comfortable         YES → skip ("hot-cache-budget-headroom")  ← only for hot
   under budget?                   NO  ↓
4. hot cache + low pressure?       YES → skip ("hot-cache-defer")            ← only for hot
                                   NO  ↓
5. Compact unconditionally              ("cold-cache-catchup" | "leaf-trigger")
```

Step 3's `isComfortablyUnderTokenBudget` check (engine.ts:1572–1587) is the right gate, but it's locked behind `cacheState === "hot"`.

## Proposed solution: separate `budgetHeadroomRatio` for unknown cache state

Add a new configuration parameter `budgetHeadroomRatio` (env: `LCM_BUDGET_HEADROOM_RATIO`) that applies the `isComfortablyUnderTokenBudget` check regardless of cache state:

- When `cacheState === "hot"`: existing `hotCacheBudgetHeadroomRatio` (0.2) takes precedence (preserving cache-invalidation-aware tuning)
- When `cacheState === "unknown"`: new `budgetHeadroomRatio` applies
- When `cacheState === "cold"`: new `budgetHeadroomRatio` applies (cold-cache catchup still runs when over the headroom threshold)

**Suggested default: 0.5** — defer compaction until 50% of budget is consumed. This preserves retrieval tool availability for the second half of the conversation while still avoiding unnecessary summarization in the first half.

The decision tree becomes:

```
1. leafTrigger fires?              NO  → skip ("below-leaf-trigger")
                                   YES ↓
2. budgetDecision fires?           YES → compact + condensed ("budget-trigger")
                                   NO  ↓
3a. hot cache + comfortable        YES → skip ("hot-cache-budget-headroom")
    under budget? (0.2 ratio)      NO  ↓
3b. ANY cache state + comfortable  YES → skip ("budget-headroom")    ← NEW
    under budget? (0.5 ratio)      NO  ↓
4. hot cache + low pressure?       YES → skip ("hot-cache-defer")
                                   NO  ↓
5. Compact                              ("cold-cache-catchup" | "leaf-trigger")
```

## Why a separate parameter (not reusing `hotCacheBudgetHeadroomRatio`)

The two ratios serve different purposes:

| Parameter | Motivation | Typical value |
|---|---|---|
| `hotCacheBudgetHeadroomRatio` (0.2) | "Cache invalidation is expensive — defer to preserve prompt cache hits" | Conservative (small), since cache breaks are costly |
| `budgetHeadroomRatio` (0.5) | "Summarization LLM calls are unnecessary when the budget has ample room" | Can be more aggressive, since there's no cache to protect |

Conflating them forces a choice between cache-preservation tuning and budget-efficiency tuning.

### Retrieval blackout during deferral

While compaction is deferred, no leaf summaries exist. This means:
- `lcm_grep scope:"summaries"` returns nothing
- `lcm_expand` / `lcm_describe` have no summary IDs to work with
- Depth-aware system prompt additions don't fire

**Mitigation:** Raw messages are all present in context during this phase, so the model has full access to conversation history — it just lacks structured DAG-based recall. Once the headroom threshold is crossed, `maxPasses = 1` per turn means leaves accumulate steadily without bursting.

### Token estimation accuracy

`isComfortablyUnderTokenBudget` relies on `estimateSessionTokenCountForAfterTurn` (character-based heuristic, ~4 chars/token). Near the threshold boundary, the estimate could swing the decision. At the proposed 0.5 default, there's a wide margin — the estimate would need to be off by a large factor to cause incorrect deferral.

### No compaction storm

`maxPasses = 1` for the `"unknown"` / `"leaf-trigger"` path (engine.ts:1991–1994), so even after a long deferral, the backlog drains at one leaf per turn. No burst of LLM calls.

### Backwards compatibility

Default `budgetHeadroomRatio = 0.5` **does change behavior** for non-Anthropic deployments — compaction is deferred where it previously ran eagerly. Setting it to `0.0` restores current behavior. If backwards compatibility is a concern, the default could be `0.0` with an opt-in model.

## Configuration

```bash
LCM_BUDGET_HEADROOM_RATIO=0.5  # defer compaction when <50% of budget used
```

```json
{
  "plugins": {
    "entries": {
      "lossless-claw": {
        "config": {
          "budgetHeadroomRatio": 0.5
        }
      }
    }
  }
}
```

jetd1:
## Summary

- Adds `budgetHeadroomRatio` config parameter (env: `LCM_BUDGET_HEADROOM_RATIO`, plugin config: `budgetHeadroomRatio`) that defers incremental leaf compaction when the current token count is below `tokenBudget × (1 - ratio)`
- Defaults to **0** (disabled) — preserving existing eager compaction behaviour as an opt-in model
- The hot-cache-specific `hotCacheBudgetHeadroomRatio` takes precedence when cache state is `"hot"`; the new parameter covers `"unknown"` and `"cold"` states
- Value clamped to [0, 0.95]

This is particularly useful for non-Anthropic providers where `cacheState` stays `"unknown"` permanently (no cache telemetry reported), and for large-context-window models where eager leaf creation at 20k tokens wastes summarization LLM calls with 95%+ budget headroom remaining.

See #409 for full design discussion including trade-off analysis (retrieval blackout during deferral, compaction storm mitigation via existing `maxPasses=1`, assembler fallback behaviour).

## Test plan

- [x] Config: default is 0, reads from plugin config, env var overrides plugin config, clamped to [0, 0.95]
- [x] Engine: `evaluateIncrementalCompaction` defers with reason `"budget-headroom"` when ratio > 0 and context is under threshold
- [x] Engine: default ratio 0 preserves existing `"leaf-trigger"` compaction behaviour
- [x] Full test suite passes (702/702)

Closes #409

### DIFF
--- a/.changeset/budget-headroom-ratio.md
+++ b/.changeset/budget-headroom-ratio.md
@@ -1,0 +1,7 @@
+---
+"@anthropic/lossless-claw": minor
+---
+
+feat(engine): add `budgetHeadroomRatio` to defer incremental compaction when under budget
+
+New opt-in config parameter `budgetHeadroomRatio` (env: `LCM_BUDGET_HEADROOM_RATIO`, default `0`) defers incremental leaf compaction while the current token count stays below `tokenBudget × (1 - ratio)`, regardless of cache state. Useful for non-Anthropic providers where cache telemetry is never reported and for large-context-window models. The hot-cache-specific `hotCacheBudgetHeadroomRatio` takes precedence when cache state is hot.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -60,7 +60,8 @@ Most installations only need to override a handful of keys. If you want a comple
   "dynamicLeafChunkTokens": {
     "enabled": true,
     "max": 40000
-  }
+  },
+  "budgetHeadroomRatio": 0
 }
 ```
 
@@ -134,6 +135,7 @@ openclaw plugins install --link /path/to/lossless-claw
 | `largeFileTokenThreshold` | `integer` | alias of `largeFileThresholdTokens` | `LCM_LARGE_FILE_TOKEN_THRESHOLD` | Legacy alias accepted by the runtime. Prefer `largeFileThresholdTokens` in new config. |
 | `maxAssemblyTokenBudget` | `integer` | unset | `LCM_MAX_ASSEMBLY_TOKEN_BUDGET` | Optional hard cap for assembly and threshold evaluation, useful with smaller-context models. |
 | `maxExpandTokens` | `integer` | `4000` | `LCM_MAX_EXPAND_TOKENS` | Default token cap for `lcm_expand_query` responses. |
+| `budgetHeadroomRatio` | `number` | `0` | `LCM_BUDGET_HEADROOM_RATIO` | Fraction of token budget that must be consumed before incremental compaction proceeds, regardless of cache state. `0` disables (default, preserving eager compaction). The hot-cache-specific `hotCacheBudgetHeadroomRatio` takes precedence when cache state is hot. Clamped to [0, 0.95]. |
 
 ### Model selection, execution, and prompts
 
@@ -185,6 +187,10 @@ When cache-aware compaction is enabled:
 - cold cache still allows bounded catch-up passes via `cacheAwareCompaction.maxColdCacheCatchupPasses`
 
 When incremental leaf compaction still runs on a hot cache, follow-on condensed passes are suppressed so the maintenance cycle only pays for the leaf pass that was explicitly justified.
+
+#### `budgetHeadroomRatio`
+
+When set to a value greater than `0`, incremental compaction is also deferred for `"unknown"` and `"cold"` cache states as long as the current token count stays below `tokenBudget × (1 - budgetHeadroomRatio)`. This is useful for non-Anthropic providers where cache telemetry is never reported (`cacheState` stays `"unknown"` permanently) and for large-context-window models where eager leaf creation wastes summarization LLM calls while ample headroom remains. The hot-cache-specific `hotCacheBudgetHeadroomRatio` takes precedence when cache state is `"hot"`. Default is `0` (disabled).
 
 ## Behavior notes
 
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -188,6 +188,10 @@
     "fallbackProviders": {
       "label": "Fallback Providers",
       "help": "Explicit fallback provider/model pairs for compaction summarization (e.g., [{\"provider\": \"anthropic\", \"model\": \"claude-haiku-4-5\"}])"
+    },
+    "budgetHeadroomRatio": {
+      "label": "Budget Headroom Ratio",
+      "help": "Fraction of token budget that must be consumed before incremental compaction runs, regardless of cache state (0 disables, 0.5 means defer until 50% used)"
     }
   },
   "configSchema": {
@@ -387,6 +391,11 @@
           "required": ["provider", "model"],
           "additionalProperties": false
         }
+      },
+      "budgetHeadroomRatio": {
+        "type": "number",
+        "minimum": 0,
+        "maximum": 0.95
       }
     }
   }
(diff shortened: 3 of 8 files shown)


100yenadmin:
First: is the general issue real? Yes. If `tokenBudget` is missing or stale, compaction policy can get skewed badly. A too-small budget causes early/eager compaction and unnecessary summarization; a too-large budget defers compaction too long.

One code-level concern: `budgetHeadroomRatio` currently appears to short-circuit before the hot-cache-specific branch. That means it looks like it can override `hotCacheBudgetHeadroomRatio` on hot caches even though the docs say the hot-cache-specific setting should win. I would either:
- gate the new branch to non-hot cache states, or
- move it after the hot-cache check,
then add a test with both ratios set.

So my read is: the knob may still have value for unknown/cold cache states and for providers without cache telemetry, however telemetry should be pulling native from config

liu51115:
Heads up — this builds on `evaluateIncrementalCompaction` in `afterTurn()`, which has a fundamental timing problem: **afterTurn fires immediately after the API call, when the provider prompt cache is always hot.** Any compaction at that point rewrites the conversation prefix and destroys a cache the provider just wrote. The next turn pays full cache-write cost again.

We documented this in #385 and the discussion on #407. The afterTurn decision structure is time-inverted — it makes compaction decisions at the one moment when compacting is guaranteed to be the most expensive.

The budget-headroom check itself is a good idea, but layering it into afterTurn means the decision is still made at the wrong time. When it *does* fire (headroom exceeded), it will destroy a hot cache.

Our approach in #385: move all proactive compaction to **pre-assembly** in `assemble()`, gated by `cacheTTLSeconds`. Compaction only runs when the cache has already expired (idle > TTL). This eliminates both the cache destruction problem and the foreground blocking problem (compaction finishes before the LLM call starts).

The headroom ratio could work well as a pre-assembly gate too — skip cache-TTL compaction when budget headroom is sufficient. That would combine both insights without the afterTurn timing inversion.

jetd1:
Closing this in favor of the pre-assembly approach in #385, which addresses the same budget-headroom concern more fundamentally.

**Why:** The `evaluateIncrementalCompaction` decision tree this PR modifies runs in `afterTurn()`, which fires immediately after the API call — when the provider prompt cache is always hot. Any compaction at that point rewrites the conversation prefix and destroys a freshly-written cache (see @liu51115's analysis in #410 comment and the detailed write-up in #385). The budget-headroom gate added here correctly suppresses unnecessary compaction, but when it *doesn't* suppress (headroom exceeded), the compaction still fires at the worst possible time.

#385 solves this by moving all proactive compaction to `assemble()` pre-assembly, gated by `cacheTTLSeconds` — compaction only runs when the cache has genuinely expired. It also already includes `leafBudgetHeadroomFactor`, which serves the same purpose as `budgetHeadroomRatio` but positioned where the decision is actually useful.

**References:**
- #407 — root issue (proactive compaction stalls foreground turns)
- #385 — pre-assembly with cache-TTL (includes `leafBudgetHeadroomFactor`)
- #408 — deferred compaction debt approach (complementary maintenance-state visibility)
- #409 — original feature discussion for this PR (keeping open for context)

PR #411, opened 2026-04-12
feat(engine): periodic background compaction sweep

ryanngit:
## Problem
LCM compaction currently advances on active turns (`afterTurn`) only. When a session goes idle, unsummarized raw turns can sit indefinitely, so dormant but important conversations never get background maintenance.

Operationally this shows up as active+unarchived conversations that keep growing until another foreground turn happens.

## What changed
- Added `LcmContextEngine.sweepActiveConversations()` in `src/engine.ts`.
  - Queries active conversations:
    - `SELECT conversation_id, session_id, session_key FROM conversations WHERE active = 1 AND archived_at IS NULL`
  - Skips ignored/stateless sessions.
  - Uses existing compaction evaluation (`evaluateLeafTrigger`) to detect eligible conversations.
  - Runs compaction via existing `compaction.compact(...)` flow.
  - Reuses circuit-breaker handling and returns sweep stats `{ swept, compacted, errors }`.
- Added periodic sweep loop in `src/plugin/index.ts`.
  - 5-minute interval.
  - Singleton guard per shared engine init.
  - Non-blocking overlap guard (`running` flag).
  - Best-effort error handling with warning logs.
  - Cleans up timer on `gateway_stop`.

## Design rationale
- Keeps compaction behavior aligned with existing engine policy rather than introducing a separate path.
- Uses per-session queue serialization to avoid ingest/compaction races.
- Keeps sweep best-effort and isolated from foreground request lifecycle.

## Backward compatibility
- No config migration required.
- Existing behavior remains unchanged except idle sessions now receive periodic maintenance.

## Configuration example
Current interval is fixed at 5 minutes in plugin wiring:

```ts
const SWEEP_INTERVAL_MS = 5 * 60 * 1000;
```

This can be made configurable later if needed, but ships as a safe default to avoid extra surface area in this PR.

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -62,7 +62,11 @@ import {
   type MessagePartType,
 } from "./store/conversation-store.js";
 import { SummaryStore } from "./store/summary-store.js";
-import { createLcmSummarizeFromLegacyParams, LcmProviderAuthError } from "./summarize.js";
+import {
+  createLcmSummarizeFromLegacyParams,
+  LcmProviderAuthError,
+  type LcmSummarizeFn,
+} from "./summarize.js";
 import type { LcmDependencies } from "./types.js";
 import { estimateTokens } from "./estimate-tokens.js";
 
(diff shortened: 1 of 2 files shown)


100yenadmin:
@ryanngit triage pass update: I marked this priority:P2 enhancement/stale-check and grouped it with the background compaction/drain queue. The branch is dirty; can you confirm whether it still overlaps with #746 or should remain a separate periodic sweep proposal?

100yenadmin:

**Summary**
@ryanngit, this PR is still potentially useful, but the current head `37e8d89d19da377331e04067e42cb6d26de50584` is `CONFLICTING` / `DIRTY` against `main`, so it cannot be reviewed for merge-readiness yet.

**Next step**

<details>

- PR title: feat(engine): periodic background compaction sweep
- Current labels: enhancement, priority:P2, stale-check

</details>

PR #413, opened 2026-04-12
fix(db): increase SQLite busy_timeout to 30 s for concurrent writers

ryanngit:
With 10+ concurrent agent lanes all writing to the same WAL-mode SQLite database, the previous 5 s busy_timeout caused frequent SQLITE_BUSY errors under burst contention. 30 s accommodates realistic worst-case write serialization without risking indefinite hangs.


jalehman:
Thank you!

PR #414, opened 2026-04-12
feat(summarize): round-robin profile rotation for concurrent summarizer calls

ryanngit:
## Problem

When multiple concurrent summarization calls iterate the same candidate list in deterministic order, they all hit the first auth profile simultaneously — causing 429 rate-limit cascading that propagates through the fallback chain.

## Fix

Add a module-level rotation counter so each invocation of the summarizer starts at a different offset in the candidate array. This distributes load across all available auth profiles while preserving the full fallback chain for each individual call.

- Counter increments atomically per call (single-threaded Node, so no mutex needed)
- Original candidate array is not mutated
- Error reporting / fallback model selection still references the original unrotated list

### DIFF
--- a/src/summarize.ts
+++ b/src/summarize.ts
@@ -2,6 +2,13 @@ import { describeLogError } from "./lcm-log.js";
 import type { LcmDependencies } from "./types.js";
 import { estimateTokens } from "./estimate-tokens.js";
 
+/**
+ * Module-level atomic counter for round-robin profile rotation.
+ * Each concurrent summarizer invocation starts at a different offset
+ * to distribute load across auth profiles and avoid thundering-herd 429s.
+ */
+let _profileRotationCounter = 0;
+
 export type LcmSummarizeOptions = {
   previousSummary?: string;
   isCondensed?: boolean;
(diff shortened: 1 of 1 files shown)


100yenadmin:
@ryanngit triage pass update: I marked this priority:P2 enhancement/stale-check because summarizer profile rotation could help concurrent summarizer reliability. The branch is dirty; can you confirm whether this is still current and rebase it if so?

100yenadmin:

**Summary**
@ryanngit, this PR is still potentially useful, but the current head `6bac0145d66032b8436207a137e0734463492dff` is `CONFLICTING` / `DIRTY` against `main`, so it cannot be reviewed for merge-readiness yet.

**Next step**

<details>

- PR title: feat(summarize): round-robin profile rotation for concurrent summarizer calls
- Current labels: enhancement, priority:P2, stale-check

</details>

ryanngit:
Closing because current OpenClaw runtime now owns authentication-profile selection, while this change rotates provider/model fallback targets instead. The original objective is therefore no longer correctly addressed by this patch. Thanks for the review.

PR #415, opened 2026-04-12
fix(store): handle UNIQUE constraint races in createConversation

ryanngit:
## Problem

When multiple agent lanes simultaneously call `getOrCreateConversation` for the same `session_key`, the `INSERT` in `createConversation` can race against the unique partial index `conversations_active_session_key_idx ON conversations (session_key) WHERE active = 1 AND session_key IS NOT NULL`.

The loser of the race gets `SQLITE_CONSTRAINT_UNIQUE`, which propagates as an unhandled error and can crash the lane or leave stale state.

## Fix

Wrap the INSERT in a try/catch. On `UNIQUE constraint failed` / `SQLITE_CONSTRAINT_UNIQUE`, fall back to selecting the winning row by `sessionKey` (preferred) or `sessionId`. The caller gets a valid `ConversationRecord` either way.

## Impact

- Eliminates intermittent crashes under concurrent multi-agent workloads
- No behavior change for single-writer or non-racing scenarios
- `getOrCreateConversation` callers are unaffected — they already expect either path to return a record

### DIFF
--- a/.changeset/neat-rockets-repair.md
+++ b/.changeset/neat-rockets-repair.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Handle conversation creation races on active session keys without crashing the caller.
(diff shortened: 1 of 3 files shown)


jalehman:
Thank you!

PR #395, opened 2026-04-11
feat: add /lcm rotate and /lcm backup commands

100yenadmin:
## Summary

Closes #394.

This adds two small plugin commands aimed at the large-session operational gap in lossless-claw:

- `/lcm backup` / `/lossless backup`
- `/lcm rotate` / `/lossless rotate`

The core user problem is that a long-lived OpenClaw session can accumulate into one very large active LCM conversation row because LCM correctly reuses the row for the stable `sessionKey`.

## Why not `/new` or `/reset`?

Today the existing commands solve adjacent problems, but not this one:

- `/new` prunes context, but keeps writing into the same active LCM row
- `/reset` changes OpenClaw session flow, which is heavier than users often want when the real issue is just LCM row size

This patch adds the missing middle ground: keep the live OpenClaw session identity, but split LCM storage.

## What changed

- adds `/lcm backup` and `/lossless backup` to create a timestamped SQLite snapshot on demand
- adds `/lcm rotate` and `/lossless rotate` to archive the current active LCM row and start a fresh row for the same live session
- reuses shared backup plumbing rather than inventing a second backup path
- fixes bootstrap so a freshly rotated empty conversation with checkpoint state starts **from now forward** instead of re-importing older transcript history
- prefers the active conversation when multiple rows share the same `sessionKey`
- hardens command and engine failure handling so backup / rotate return structured failure output instead of bubbling raw runtime errors
- documents why `/new`, `/reset`, and `/lcm rotate` solve different problems

## Why this is useful now

Cross-session search and retrieval change the tradeoff here in a good way: once archived rows remain searchable, splitting an oversized active row is usually a performance win.

The active row stays smaller and cheaper to bootstrap, assemble, and maintain, while archived history remains available for search and retrieval across conversations.

## Validation

- `npm test` ✅ (`40` files / `699` tests)
- `npm run build` ✅

New coverage includes:

- standalone `/lcm backup`
- `/lcm backup` structured failure output
- `/lcm rotate` backup-first behavior and command UX
- `/lcm rotate` structured failure output for backup/engine exceptions
- active-row preference when multiple rows share a `sessionKey`
- engine-level rotate path that archives the old row, creates a fresh row, and imports only newly appended transcript messages after the rotate checkpoint
- engine-level unreadable transcript handling for rotate

## Manual validation

- create or reuse a large active LCM conversation
- run `/lcm backup` and verify a timestamped SQLite backup is created
[...]

PR #416, opened 2026-04-12
Defer optional LCM maintenance off the immediate session lane

100yenadmin:
## Summary
- keep reply-critical LCM work on the immediate session lane
- defer optional bootstrap, maintain, and threshold-compaction follow-up work when the runtime budget is already tight
- add step-level bootstrap diagnostics and architecture docs so reviewers can see exactly where the lane is spending time

## Scope after split
The verified message-identity hash/index work moved into the narrower companion PR:
- #417 `Index message identity by verified hash`

This PR now focuses only on **hot-path scheduling and lifecycle observability**.

## Problem
Large sessions can feel like the gateway or provider timed out even when the real issue is that LCM is still holding the session lane after the model turn.

The important distinction is:
- some work is required before the next turn can proceed safely
- some work is valuable, but can happen a tick later without changing correctness

### 1. Defer optional follow-up work off the immediate lane
`afterTurn()` still ingests the new turn inline, but proactive threshold compaction can be scheduled onto the next event-loop tick instead of blocking the immediate reply path.

`maintain()` and post-bootstrap cleanup use the same rule when the runtime already reports very little time remaining.

```mermaid
flowchart LR
    A["turn completes"] --> B["persist new messages"]
    B --> C["decide whether maintenance is required"]
    C --> D{"enough time left?"}
    D -- yes --> E["run maintenance inline"]
    D -- no --> F["schedule background follow-up"]
    E --> G["release session lane"]
    F --> G
```

### 2. Make bootstrap slow paths attributable
Bootstrap now logs separate timings for:
- checkpoint check
- append-only probe
- full transcript parse
- reconcile

It also logs explicit fast-path miss reasons before falling back to the expensive read path.

```mermaid
flowchart TD
    A["bootstrap"] --> B["checkpoint check"]
    B --> C{"checkpoint hit?"}
    C -- yes --> D["done"]
    C -- no --> E["append-only probe"]
    E --> F{"append-only hit?"}
    F -- yes --> G["import new tail only"]
    F -- no --> H["full transcript parse"]
    H --> I["reconcile"]
    G --> J["done"]
    I --> J
```

### 3. Tighten lifecycle tracing without mixing in storage changes
Lifecycle tracing now carries a stable trace label through assemble/bootstrap/maintain/afterTurn so one slow turn can be correlated across hooks.

The trace helper was consolidated so the branch no longer carries duplicate implementations.

## Why this is separate from #417
#417 is about making identity lookups cheaper.
This PR is about making the lane spend less time on optional follow-up work.

They complement each other, but they answer different review questions:
- #417: can we make reconcile lookups cheaper without changing lifecycle semantics?
- #416: can we reduce lane occupancy without weakening correctness?

## Relationship to #395
#395 (`/lcm rotate`) reduces the size of the active conversation row.
This PR reduces how much optional work blocks the immediate lane.

If you want the smallest staged rollout, the order I would recommend is:
1. #395 rotate
2. #417 verified hash/index
3. #416 deferred hot-path maintenance

## Verification
- `git diff --check`
- `npx vitest run test/engine.test.ts -t "afterTurn|bootstrap|maintain"`

## Docs
- adds a new hot-path guardrails section with diagrams to `docs/architecture.md`

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -58,6 +58,43 @@ When OpenClaw processes a turn, it calls the context engine's lifecycle hooks:
 2. **ingest** / **ingestBatch** — Persists new messages to the database and appends them to context_items.
 3. **afterTurn** — After the model responds, ingests new messages, then evaluates whether compaction should run.
 
+### Hot-path guardrails
+
+The session lane is the part of LCM that users feel directly: if it stays busy, the next live turn looks hung even when the model provider is healthy.
+
+The hot-path changes in this branch keep the **minimum correctness work** inline and push **optional cleanup work** into best-effort follow-up tasks when the turn is already close to its deadline.
+
+```mermaid
+flowchart LR
+    A["turn starts"] --> B["bootstrap / assemble / ingest"]
+    B --> C["reply-critical work stays inline"]
+    C --> D["optional maintenance evaluated"]
+    D --> E{"enough time left?"}
+    E -- yes --> F["run maintenance now"]
+    E -- no --> G["schedule best-effort background work"]
+    F --> H["session lane released"]
+    G --> H
+```
+
+In practice that means:
+- `bootstrap()` logs step-by-step timings for checkpoint checks, append-only probes, transcript parsing, and reconcile, so slow starts are attributable instead of opaque.
+- `maintain()` can defer bootstrap checkpoint refresh when transcript cleanup already consumed the remaining budget.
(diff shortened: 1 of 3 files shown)


100yenadmin:
Split update: the verified message-identity hash/index work has moved out into #417 so this PR can stay focused on hot-path scheduling and lifecycle observability. Recommended review order if you want the smallest staged operational experiment is #395 -> #417 -> #416.

jalehman:
Closing in favor of #507 that was inspired by this but updated.

PR #417, opened 2026-04-12
Index message identity by verified hash

100yenadmin:
## Summary
- add a stored `identity_hash` column to `messages`
- backfill existing rows and add an index on `(conversation_id, identity_hash)`
- switch message-identity lookups to use the indexed hash path, with raw `role` + `content` verification after the indexed hit

## Why split this out
This PR is the narrow performance/correctness slice from #416.

It does **not** change `afterTurn()` scheduling, bootstrap time budgets, or deferred maintenance behavior.
That makes it easier to answer one focused question in review:

> Does a verified hash index make large-conversation identity lookups cheaper without changing lifecycle semantics?

### Before
Message identity lookups scanned by full `role` + `content` equality inside a conversation.
That is correct, but it gets more expensive as conversations grow and reconcile has to ask repeated questions like:
- have we already seen this exact message?
- how many times has this exact message identity occurred?

```mermaid
flowchart LR
    A["reconcile / dedup asks for message identity"] --> B["scan messages in conversation"]
    B --> C["compare full role + content"]
    C --> D["answer hasMessage / countMessagesByIdentity"]
```

### After
We store a deterministic SHA-256 fingerprint for the same identity tuple (`role`, `content`).
Lookups first narrow candidates through the indexed hash, then verify the original `role` and `content` before returning a match or count.

```mermaid
flowchart LR
    A["reconcile / dedup asks for message identity"] --> B["compute identity_hash from role + content"]
    B --> C["indexed lookup by conversation_id + identity_hash"]
    C --> D{"candidate rows?"}
    D -- no --> E["not found"]
    D -- yes --> F["verify raw role + content"]
    F --> G["exact match / exact count"]
```

## Safety model
- The hash is a **performance hint**, not the source of truth.
- `hasMessage()` still verifies the original `role` and `content` after the indexed lookup.
- `countMessagesByIdentity()` counts only rows whose raw content still matches after the hash prefilter.
- Existing databases migrate in place: add column, backfill missing hashes in batches, create index.

## Why this matters operationally
This specifically targets the large-session reconcile/identity-lookup path.
It should help when bootstrap or replay needs to anchor against repeated identical messages such as empty tool outputs.

It is **not** the full lane-stall fix by itself. It does not move post-turn compaction off the hot path.

## Verification
- `git diff --check`
- `npx vitest run test/migration.test.ts`

## Related work
- Split out of #416
- Pairs well with #395 (`/lcm rotate`) for very large long-lived conversations

### DIFF
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -1,5 +1,6 @@
 import type { DatabaseSync } from "node:sqlite";
 import { getLcmDbFeatures } from "./features.js";
+import { buildMessageIdentityHash } from "../store/message-identity.js";
 import { parseUtcTimestampOrNull } from "../store/parse-utc-timestamp.js";
 
 type MigrationLogger = {
@@ -33,6 +34,12 @@ type SummaryParentEdgeRow = {
 
 type TableNameRow = {
   name?: string;
+};
+
+type MessageIdentityBackfillRow = {
+  message_id: number;
+  role: string;
+  content: string;
 };
 
 type FtsTableSpec = {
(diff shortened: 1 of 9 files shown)


100yenadmin:
Companion split from #416. This PR is intentionally narrow: schema/backfill/index plus verified hash-based identity lookups only. It pairs well with #395 if you want to test large-session reconcile improvements without taking the deferred-maintenance changes yet.

100yenadmin:
@jalehman I didnt see the code bot responses above so I'd recommend debugging or treating this one as potentially volatile and ask agent to solve the issues above post merge.

jalehman:
> @jalehman I didnt see the code bot responses above so I'd recommend debugging or treating this one as potentially volatile and ask agent to solve the issues above post merge.

Addressed in #422

PR #418, opened 2026-04-13
Fix manual compaction under-target result

gitchrisqueen:
## Summary
- treat full-sweep/manual compaction as successful when persisted context is already under target
- prefer `tokensAfter` over stale observed token counts when deciding the result and reason
- add a regression test for the stale observed-token/manual compaction path

## Validation
- `npx vitest run test/engine.test.ts -t "treats full-sweep compaction as already under target when tokensAfter is below budget"`
- `npm test` *(current fork baseline still has unrelated timeouts in `test/plugin-config-registration.test.ts` in this environment)*


gitchrisqueen:
@copilot apply changes based on the comments in [this thread](https://github.com/Martian-Engineering/lossless-claw/pull/418#pullrequestreview-4099490153)

jalehman:
Thank you!

PR #420, opened 2026-04-13
fix: back up /lcm rotate from a secondary sqlite connection

### LINKED ISSUE #419: fix(rotate): avoid VACUUM backup failure on live /lcm rotate
## Summary

`/lcm rotate` can fail before the archive step starts with:

```text
status: failed
reason: cannot VACUUM from within a transaction
```

The failure happens during the pre-rotate backup, not during the actual conversation archive/create-fresh-row logic.

## What we saw in production

On April 13, 2026 we attempted `/lcm rotate` on a large active conversation (`conversation id: 1628`, `messages: 103,932`) to get the hot row out of the foreground path.

The command reached Lossless Claw correctly and returned:

```text
🦀 Lossless Claw v0.8.0
🪓 Lossless Claw Rotate
📍 Current conversation
conversation id: 1,628
session key: agent:main:main
messages: 103,932
💾 Backup
status: failed
reason: cannot VACUUM from within a transaction
```

That meant:
- the command handler was wired correctly
- rotate never reached `rotateSessionStorage(...)`
- the failure was entirely in the backup-first step

After patching the live install to back up from a second SQLite connection instead of running `VACUUM INTO` on the active command connection, the next `/lcm rotate` succeeded and archived the old row cleanly.

## Root cause

Today `buildRotateText()` calls:

- `createLcmDatabaseBackup(params.db, { label: "rotate", replaceLatest: true })`

That backup helper eventually does:

- `db.exec("VACUUM INTO ...")`

SQLite rejects `VACUUM` when the same connection is already inside a transaction. In practice, `rotate` runs on a live runtime-backed connection where that can happen.

So the issue is not "the active conversation is too large to rotate". The issue is that the rotate backup implementation uses a SQLite operation that is illegal on the active connection state it can encounter in real usage.

## Why this matters

`/lcm rotate` is specifically the escape hatch for giant hot conversations. If the backup-first step fails on the live connection, the operator cannot use the feature when they need it most.

## Expected

`/lcm rotate` should:
1. create the safety backup
2. archive the current active conversation row
3. create the fresh active row
4. checkpoint the transcript frontier

## Actual

It can fail at step 1 before the actual rotate begins.

## Proposed fix

Use a second SQLite connection for the pre-rotate backup path.

That preserves backup-first safety while avoiding `VACUUM INTO` on the active runtime connection.

One implementation that worked in production:
- open a fresh `DatabaseSync(fileBackedDatabasePath)`
- use the Node `node:sqlite` `backup()` API to copy into the backup path
- keep the existing `replaceLatest` semantics by backing up to a temp path and renaming it into the `*-latest.bak` target

## Regression test idea

Add a rotate command test that keeps a transaction open on the live command DB connection and verifies `/lcm rotate` still succeeds.

That would have caught this exact failure mode.

100yenadmin:
## Summary

Fix `/lcm rotate` so its pre-rotate backup does not fail with:

```text
cannot VACUUM from within a transaction
```

Closes #419.

## What happened in production

On April 13, 2026 we used `/lcm rotate` against a large hot conversation row to get the active row out of the foreground path.

The command reached Lossless Claw correctly, resolved the current conversation, and then failed before rotation started:

```text
🪓 Lossless Claw Rotate
📍 Current conversation
conversation id: 1,628
session key: agent:main:main
messages: 103,932
💾 Backup
status: failed
reason: cannot VACUUM from within a transaction
```

That meant the archive/create-fresh-row logic never ran. The failure was entirely in the backup-first step.

## Root cause

`buildRotateText()` currently calls `createLcmDatabaseBackup(... replaceLatest: true)` on the live runtime DB connection.

That helper ultimately runs `VACUUM INTO ...` on that same connection. SQLite rejects `VACUUM` when the connection is already inside a transaction, which is a real state the live runtime can encounter.

So the operator-facing symptom looked like “rotate failed on a large conversation”, but the actual bug was much narrower: rotate's backup implementation was using a SQLite operation that is illegal on the active connection state.

## Fix

Keep the existing rotate behavior and `*-latest.bak` semantics, but move the backup copy itself onto a second SQLite connection:

- add `createLcmDatabaseBackupFromSecondaryConnection(...)`
- use Node's `node:sqlite` `backup()` API on a fresh `DatabaseSync(fileBackedDatabasePath)`
- preserve `replaceLatest` behavior by copying to a temp path and renaming into the `*-latest.bak` target
- wire `/lcm rotate` to use that secondary-connection helper for the pre-rotate backup step

This keeps the safety property of “backup first” without depending on `VACUUM INTO` on the live command connection.

## Why this scope is intentionally narrow

This PR only fixes the rotate backup path.

It does **not** change:
- standalone `/lcm backup`
- doctor clean backup behavior
- rotate archive/checkpoint semantics
- any non-rotate storage logic

The goal is to land the production blocker fix with the smallest review surface.

## Tests

Added/updated focused command coverage:

- `/lcm rotate` still works and replaces the latest rotate backup
- `/lcm rotate` succeeds even when the live DB connection already has an open transaction
- `/lcm rotate` still reports a structured failure if the new secondary-connection helper itself throws

## Validation

- `git diff --check`
- `npx vitest run test/lcm-command.test.ts`

## Notes

The production confirmation for this approach was straightforward:
- before this patch, rotate failed at the backup step with `cannot VACUUM from within a transaction`
- after switching the backup to a second SQLite connection, the next `/lcm rotate` succeeded, archived the hot conversation, and created the fresh active row as expected

### DIFF
--- a/.changeset/green-lamps-trade.md
+++ b/.changeset/green-lamps-trade.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix `/lcm rotate` so it waits for the live database connection to become idle, takes a faithful pre-rotate backup on that connection, and then compacts the current session transcript without replacing the active LCM conversation. Rotation now preserves the existing conversation id, summaries, and context items while refreshing bootstrap state so dropped transcript history is not replayed.
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The plugin now ships a bundled `lossless-claw` skill plus a small plugin command
 
 - `/lcm` shows version, enablement/selection state, DB path and size, summary counts, and summary-health status
 - `/lcm backup` creates a timestamped backup of the current LCM SQLite database
-- `/lcm rotate` archives the active LCM row for the current session and starts a fresh row without changing the live OpenClaw session identity
+- `/lcm rotate` rewrites the active session transcript into a compact tail-preserving form without changing the live OpenClaw session identity or current LCM conversation
 - `/lcm doctor` scans for broken or truncated summaries
 - `/lcm doctor clean` shows read-only high-confidence junk diagnostics for archived subagents, cron sessions, and NULL-key orphaned subagent runs
 - `/lcm status` shows plugin, conversation, and maintenance state including deferred compaction debt
@@ -281,7 +281,7 @@ For large sessions, neither command is a perfect “keep my live agent context, 
 - `/new` keeps writing into the same active LCM conversation row.
 - `/reset` changes OpenClaw session flow, which is heavier than users often want when their real problem is just LCM row size.
 
-`/lcm rotate` fills that gap. It replaces one rolling `rotate-latest` SQLite backup, archives the current active LCM row, creates a fresh active row for the same `sessionKey`, and checkpoints the current transcript frontier so the new row starts from now forward instead of replaying old history. Older rows remain searchable, which pairs well with cross-session search and usually improves hot-path bootstrap and assemble performance for long-lived sessions. If you want timestamped snapshots instead, run `/lcm backup`.
+`/lcm rotate` fills that gap. It replaces one rolling `rotate-latest` SQLite backup, rewrites the current session transcript down to the preserved live tail plus current session settings, and refreshes the bootstrap frontier on the same active LCM conversation so dropped transcript history is not replayed. Existing summaries, context items, and conversation identity stay in place; only the transcript backing is compacted. If you want additional timestamped snapshots instead, run `/lcm backup`.
 
 `newSessionRetainDepth` (or `LCM_NEW_SESSION_RETAIN_DEPTH`) controls how much summary structure survives `/new`:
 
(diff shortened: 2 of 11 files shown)


jalehman:
This PR ended up taking a different direction than the original implementation.

The first pass made `/lossless rotate` back up from a secondary SQLite connection so rotate could succeed while the primary connection was inside an open transaction. That solved the immediate SQLite error, but it was not lossless: a second connection only sees committed state, so the rotate backup could silently omit in-transaction rows.

The final implementation changes the design in two places:

- Rotate now waits behind same-session work and the shared DB transaction mutex, takes the pre-rotate backup on the live connection, and only then performs the rotate work. That preserves a faithful backup without racing an active transaction.
- Rotate itself is now transcript maintenance rather than conversation replacement. Instead of archiving the current conversation row and creating a fresh one, it keeps the same conversation id, summaries, and context items in place, rewrites the backing transcript down to a preserved live tail, and refreshes bootstrap state so the dropped transcript history is not replayed.

That change in direction came out of local validation against a real long-lived session. The archive-and-fresh-row model turned out to be too close to `/new`/`/reset` semantics and reintroduced exactly the replay/recompaction behavior rotate is supposed to avoid. The current version is narrower: compact the transcript, preserve LCM state.

PR #444, opened 2026-04-15
fix: restore runtime dep and valid context engine registration

andyylin:
## Summary
- add the missing `@mariozechner/pi-coding-agent` runtime dependency back to `package.json`
- stop registering the same engine instance under the extra `default` context engine id
- add a regression test that asserts the plugin only registers `lossless-claw`

## Why
Installing `@martian-engineering/lossless-claw@0.9.1` in OpenClaw exposed two independent runtime failures:

1. the published bundle imports `@mariozechner/pi-coding-agent`, but `package.json` does not declare it as a dependency
2. the plugin registers a context engine as both `lossless-claw` and `default`, while the engine reports `info.id = "lossless-claw"`

That second issue trips OpenClaw's context-engine contract with:
`info.id must match registered id "default"`

## Validation
- `npm run build`
- `npx vitest run test/plugin-config-registration.test.ts -t "registers only the lossless-claw context engine id"`

## Notes
I also reproduced both issues from a fresh `npm pack @martian-engineering/lossless-claw@0.9.1`, so this is not specific to one local install.

### DIFF
--- a/.changeset/bright-rats-sneeze.md
+++ b/.changeset/bright-rats-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix context-engine registration so the plugin only registers its canonical `lossless-claw` id, align runtime Pi package versions with the current OpenClaw stack, and tighten selection helpers to stop treating the old `default` alias as equivalent to the plugin id.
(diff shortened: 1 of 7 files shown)


jalehman:
Thank you!

PR #446, opened 2026-04-15
fix(cli): register native lcm/lossless commands

coolmanns:
## Summary
- add the missing `api.registerCli(...)` bridge for native `openclaw lcm` and `openclaw lossless`
- reuse the existing `createLcmCommand()` implementation so runtime/slash and native CLI stay aligned
- add a regression assertion that CLI registration metadata is present

## Why
`createLcmCommand()` already defines the runtime command surface:
- `name: "lcm"`
- `nativeNames.default: "lossless"`

But in current OpenClaw, native Commander commands are populated from `registerCli(...)`, not from runtime `registerCommand(...)` metadata alone.

That means the plugin can load and the runtime/slash command can work, while native:
- `openclaw lcm`
- `openclaw lossless`

remain unavailable.

This patch adds the missing CLI bridge and keeps it wired to the same command handler.

## Validation
```bash
npm test -- --run test/plugin-config-registration.test.ts test/lcm-command.test.ts
```

Passed locally:
- `test/lcm-command.test.ts`
- `test/plugin-config-registration.test.ts`
- 64 tests passed

## Related
- Separate from #444, which fixes the context-engine registration/runtime dependency issue.
- This PR addresses the missing native CLI registration path only.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,17 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.8.0",
+  "version": "0.9.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.8.0",
+      "version": "0.9.1",
       "license": "MIT",
       "dependencies": {
         "@mariozechner/pi-agent-core": "*",
         "@mariozechner/pi-ai": "*",
+        "@mariozechner/pi-coding-agent": "^0.67.2",
         "@sinclair/typebox": "0.34.48"
       },
       "devDependencies": {
@@ -862,7 +863,6 @@
       "resolved": "https://registry.npmjs.org/@borewit/text-codec/-/text-codec-0.2.1.tgz",
       "integrity": "sha512-k7vvKPbf7J2fZ5klGRD9AeKfUvojuZIQ3BT5u7Jfv+puwXkUBUT5PVyMDfJZpy30CBDXGMgw7fguK/lpOMBvgw==",
       "license": "MIT",
-      "peer": true,
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/Borewit"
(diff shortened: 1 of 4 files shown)


100yenadmin:
@coolmanns I marked this as a possible linked candidate for the lcm command/tool registration failures, especially issue #615, but the branch is currently dirty and too old to treat as a verified fix. If you still want this path considered, please refresh against main and link the exact issue it resolves.

100yenadmin:
@coolmanns maintainer triage update: I rechecked this after the current P0/P1 pass. I no longer think #446 should sit in the P1 merge lane.

Why:
- The P1 issue this was loosely linked to, #615, is now closed after #740 established the supported-host behavior: lossless-claw requires OpenClaw hosts that assemble context before prompt construction, and unsupported CLI-style hosts now fail explicitly instead of silently degrading.
- Current `main` already has the `lossless-claw`-only context-engine registration shape from the first commit in this PR.
- The branch is still `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, has no current CI, and carries old `package.json` / `package-lock.json` dependency churn that should not be merged as-is.

The remaining useful part appears to be the native `api.registerCli` bridge for `openclaw lcm` / `openclaw lossless`, but that should be treated as a fresh, focused P2/P3 compatibility/admin-command improvement rather than a P1 fix for #615.

Next action: please rebase onto current `main` and rescope to the minimal `registerCli` patch only, with current tests for native CLI registration. If you want, link the exact issue this now resolves; otherwise I recommend replacing this with a clean new PR.

100yenadmin:

**Summary**
@coolmanns, this PR is still potentially useful, but the current head `116b1fc5f44cc3a732b0db767ca19a72094cf9e5` is `CONFLICTING` / `DIRTY` against `main`, so it cannot be reviewed for merge-readiness yet.

**Next step**

<details>

- PR title: fix(cli): register native lcm/lossless commands
- Current labels: bug, priority:P2, linked-pr, stale-check

</details>

PR #450, opened 2026-04-16
fix: use resolved summary plugin config fallback

coryscook:
## Summary
- fall back to the resolved LCM plugin config from deps.config when full runtime config is unavailable
- preserve existing runtime config precedence when plugins.entries.lossless-claw.config is present
- add a regression test for plugin summaryModel/summaryProvider resolved through deps.config
- add a patch changeset

## Why
Some OpenClaw compaction paths call createLcmSummarizeFromLegacyParams without legacyParams.config even though the plugin config has already been resolved into deps.config. In that shape, summaryModel/summaryProvider were ignored and LCM could resolve no summary candidates, causing emergency truncation despite valid plugin config.

## Verification
- docker container Node runtime: node v24.14.0
- npm ci --include=dev
- ./node_modules/.bin/vitest run test/summarize.test.ts: 48 passed
- npm run build

### DIFF
--- a/.changeset/quiet-crabs-resolve.md
+++ b/.changeset/quiet-crabs-resolve.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Use the resolved plugin summary config when runtime config is unavailable so compaction keeps the configured summary model instead of falling back to emergency truncation.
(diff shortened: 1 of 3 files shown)


jalehman:
Thank you!

PR #458, opened 2026-04-17
fix: wrap runLcmMigrations in BEGIN EXCLUSIVE transaction

JanPlessow:
## Problem

When multiple agents start simultaneously (or a gateway restarts quickly), concurrent callers race inside `runLcmMigrations`. The DDL and backfill steps are individually idempotent, but without a top-level lock, two callers can collide mid-migration and produce:

- `database is locked` errors
- Partial schema state if one caller crashes after some DDL but before others
- An open transaction leak when fts5 is unavailable (the early-return path was missing a `COMMIT`, leaving `BEGIN EXCLUSIVE` open until the connection closed)

## Fix

Wrap the entire `runLcmMigrations` function in `BEGIN EXCLUSIVE` / `COMMIT` / `ROLLBACK`:

- The **first** caller acquires the exclusive lock and runs all migrations normally
- Any **concurrent** caller blocks at `BEGIN EXCLUSIVE` until the first commits, then finds every step already applied and finishes in <5 ms
- On any error, `ROLLBACK` releases the lock cleanly
- The early-exit path (when fts5 is unavailable) now correctly `COMMIT`s before returning

## Changes

- `src/db/migration.ts` — add `BEGIN EXCLUSIVE` / `COMMIT` / `ROLLBACK` wrapper around `runLcmMigrations`; add missing `COMMIT` on the `!fts5Available` early-return path

## Notes

The existing per-step SAVEPOINTs inside `runVersionedBackfillStep` are complementary and remain unchanged — they handle per-conversation rollback within the outer transaction.

### DIFF
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -724,6 +724,14 @@ export function runLcmMigrations(
   options?: { fts5Available?: boolean; log?: MigrationLogger },
 ): void {
   const log = options?.log;
+  // Serialize concurrent callers at the SQLite level. When multiple agents
+  // (or a restart race) call runLcmMigrations simultaneously, the second
+  // caller blocks on BEGIN EXCLUSIVE until the first commits. Every step is
+  // idempotent, so the second caller completes in <5 ms after acquiring the
+  // lock. Without this guard, concurrent DDL/backfill writes can produce
+  // "database is locked" errors or partial schema state.
+  db.exec(`BEGIN EXCLUSIVE`);
+  try {
   db.exec(`
     CREATE TABLE IF NOT EXISTS conversations (
       conversation_id INTEGER PRIMARY KEY AUTOINCREMENT,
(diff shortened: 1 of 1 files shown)


jalehman:
Thank you for this, but closing as superseded by #455 .

100yenadmin:
Merge gate: **NOT ready-to-merge**. Outstanding work blocking merge:

1. **Merge conflicts** — `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. Rebase onto `main` and resolve.
2. **CI test failure** — [CI run 24545368813](https://github.com/Martian-Engineering/lossless-claw/actions/runs/24545368813/job/71759611712) reports:
   - `Error: cannot start a transaction within a transaction` originating at `backfillMessageIdentityHashes` (`src/db/migration.ts:190`), called via `runMigrationStep` → `runLcmMigrations` (`src/db/migration.ts:971-972`).
   - Downstream assertion failure in `test/migration.test.ts:1044` — expected `'simulated state write failure'` but received the nested-transaction error.

The new `BEGIN EXCLUSIVE` wrapper around `runLcmMigrations` collides with inner steps (like `backfillMessageIdentityHashes`) that open their own transactions. Either:
- strip the inner `BEGIN`/`COMMIT` from any migration step that runs inside the wrapper, or
- detect an active transaction in the inner step and skip its own BEGIN, or
- use SQLite `SAVEPOINT` for the inner step.

Re-run CI after the fix and after resolving the rebase conflict. Holding merge until both go green.

PR #465, opened 2026-04-18
fix: add type guards for .trim() calls on potentially non-string DB/config values

liu51115:
# Add type guards on `.trim()` calls in DB-result and config paths

## Problem

`SQLite` (via `node:sqlite`) returns `null`, `number`, or `Buffer` for column values that are typed as `string` in TypeScript. When the engine's bootstrap/maintain path encounters such a value and calls `.trim()` on it, it throws:

```
TypeError: value.trim is not a function
```

This crashes the per-conversation context-engine bootstrap, leaving the agent unable to respond. In our deployment it surfaced as a recurring `[agent/embedded] context engine maintain failed (bootstrap): TypeError: value.trim is not a function` on every assistant turn until the affected conversation was rolled.

The thrown variable is sometimes named `value`, sometimes a SQL row column (e.g. `summary.content`, `target.createdAt`, `row.content`, `dbPath`), and is reached from at least these entry points:

- `LcmContextEngine.maintain()` / `bootstrap()` in `src/engine.ts`
- `CompactionEngine.compact*()` in `src/compaction.ts`
- doctor/repair flows in `src/plugin/lcm-doctor-apply.ts`
- timestamp parsing in `src/store/parse-utc-timestamp.ts`
- summary prompt building in `src/summarize.ts`

The actual TS types claim `string` everywhere, so the static signature gives no warning — only the runtime SQLite return values can violate it (typically when the underlying column was inserted with a non-string, or a migration left a row with a NULL where `COALESCE` was expected).

## Fix

Add `typeof x === "string"` guards (or `String(x ?? "")` fallbacks) immediately before every `.trim()` call on values that originate from SQLite rows, plugin/runtime config, or loosely-typed parameters. 28 guards across 12 files. Behaviour on the happy path (string value) is unchanged.

## Files touched

```
src/assembler.ts                 |  2 +-
src/compaction.ts                |  9 ++++++---
src/db/connection.ts             |  1 +
src/engine.ts                    |  2 ++
src/expansion-policy.ts          |  2 ++
src/plugin/lcm-command.ts        |  2 ++
src/plugin/lcm-doctor-apply.ts   | 15 ++++++++-------
src/prune.ts                     |  1 +
src/store/conversation-store.ts  |  1 +
src/store/full-text-fallback.ts  |  1 +
src/store/parse-utc-timestamp.ts |  1 +
src/summarize.ts                 |  2 ++
12 files changed, 28 insertions(+), 11 deletions(-)
```

## Risk / Compatibility

- No public API change.
- No schema or migration change.
- Defensive only — when the value is the expected `string`, behaviour is identical.
- When the value is non-string (the failing case today), the function now returns the same "empty" sentinel it would have returned for an empty string instead of throwing. Existing fall-through branches handle empty correctly in every site touched.

## Reproduction

In a non-trivial production conversation (multiple bootstrap rounds, a doctor pass, several leaf compactions), ~5–10% of agent turns hit this once an offending row exists. After applying the patch, the same conversation runs `bootstrap`/`maintain` cleanly with zero `value.trim` errors over thousands of turns.

## Test status

- All existing tests pass locally (`pnpm test`).
- No new tests added — each guard is a one-line shape check before an existing trim, with no new branches that aren't already covered by empty-string tests in the same files. Happy to add explicit non-string regression tests if you'd prefer.

### DIFF
--- a/.changeset/fuzzy-bulldogs-applaud.md
+++ b/.changeset/fuzzy-bulldogs-applaud.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Harden defensive handling for non-string database path and timestamp values so malformed runtime data does not trigger `.trim()` crashes or silently skew stored chronology.
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -1141,7 +1141,7 @@ export class ContextAssembler {
     // content text AND the message_parts table are empty — assistant
     // messages that contain tool calls have empty text content but
     // non-empty parts and must be preserved.
-    if (msg.role === "assistant" && !msg.content.trim() && parts.length === 0) {
+    if (msg.role === "assistant" && !(typeof msg.content === "string" ? msg.content.trim() : "") && parts.length === 0) {
       return null;
     }
     const roleFromStore = toRuntimeRole(msg.role, parts);
--- a/src/compaction.ts
+++ b/src/compaction.ts
@@ -222,6 +222,7 @@ function parseMessagePartMetadata(part: CreateMessagePartInput | { metadata: str
 
 /** Detect whether a string is mostly binary/base64 payload and not meaningful prose. */
 function looksLikeBinaryPayload(value: string): boolean {
+  if (typeof value !== "string") return false;
   const trimmed = value.trim();
   if (!trimmed) {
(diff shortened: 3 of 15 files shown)


jalehman:
Thank you!

PR #477, opened 2026-04-20
fix: drop trailing empty-content assistant from assembled context

liu51115:
## Problem

When an Anthropic-backed agent session triggers LCM compaction mid-session, subsequent turns can fail with:

```
400 {"type":"invalid_request_error","message":"This model does not support assistant message prefill. The conversation must end with a user message."}
```

`assemble()` occasionally returns a message list whose final entry is `{ role: "assistant", content: [] }`. Anthropic models that don't support assistant prefill (e.g. `claude-sonnet-4-6`) reject this shape.

The existing `sanitizeToolUseResultPairing` pass strips dangling `tool_use` pairings but does not remove an empty-content assistant tail, so the bad shape reaches the provider.

## Repro

1. Run an agent backed by `anthropic/claude-sonnet-4-6` with LCM enabled.
2. Accumulate enough context to breach the compaction threshold.
3. Trigger compaction (proactive or manual).
4. Send another user turn.

Observed: the next prompt posted to the Anthropic API ends with `{"role":"assistant","content":[]}` and the turn fails. Each retry appends another empty-assistant entry to the session transcript, so the failure is sticky until the session is reset.

## Fix

In `src/transcript-repair.ts`, add a helper `stripTrailingEmptyAssistantPrefill` that pops a trailing assistant message whose content is undefined, null, empty string, or empty array. `ContextAssembler.assemble()` chains this helper after `sanitizeToolUseResultPairing` before returning.

No-op in the common case; prevents the 400 in the post-compaction edge case.

## Tests

Regression tests included in `test/transcript-repair.test.ts`:

- drops empty-array content
- drops undefined content
- drops whitespace-only string content
- keeps a real assistant tail
- keeps a trailing empty user message (not affected)
- no-op on empty list
- does not mutate input

All 10 tests in `transcript-repair.test.ts` pass locally (3 existing + 7 new).

## Risk

- Surgical: a small guard localised to the final step of `assemble()`.
- Harmless in the non-repro case.
- If a legitimate empty-content assistant were part of a valid conversation, dropping it just means the next turn re-emits the assistant output.

### DIFF
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -1056,6 +1056,22 @@ export class ContextAssembler {
       .filter((entry) => entry.segment === "freshTail")
       .map((entry) => entry.message);
     const repaired = sanitizeToolUseResultPairing(cleaned) as AgentMessage[];
+    // Guard: Anthropic rejects a prompt ending with an empty-content assistant
+    // message ("does not support assistant message prefill"). Post-compaction
+    // assembly can produce this shape. Drop it so the conversation ends on the
+    // last real user/assistant turn.
+    if (repaired.length > 0) {
+      const last = repaired[repaired.length - 1];
+      const content = (last as { content?: unknown }).content;
+      const isEmpty =
+        content === undefined ||
+        content === null ||
+        (typeof content === "string" && content.trim() === "") ||
+        (Array.isArray(content) && content.length === 0);
+      if (last.role === "assistant" && isEmpty) {
+        repaired.pop();
+      }
+    }
     return {
       messages: repaired,
       estimatedTokens,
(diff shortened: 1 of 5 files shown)


jalehman:
Thanks for the careful report and targeted fix here. We have since merged the assistant-prefill/cold-cache fixes on current main, so this PR is now superseded:

- #502, commit 74004a4ef486ba2e351a4143acfb8cb4a7573b6c, covers the cold-cache/no-user-turn fallback case.
- #504, commit 7063a1f17b7be4cd60fc87563d5dbf9ce125b1c4, covers the assemble() early-return assistant prefill case and preserves the runtime hook's required new-array fallback behavior.

Because those merged changes overlap this PR and current main now conflicts in the same engine fallback path, I'm closing this as superseded rather than asking you to rework it. Thanks again for tracking this down and contributing the repro and tests.

Issue #461, opened 2026-04-17
fix: /lcm status shows stale cache state (does not account for TTL expiry)


liu51115:
## Bug

The `/lcm` status command displays the stored `cacheState` value from the DB, which reflects the last-observed state at telemetry update time. It does not recompute whether the cache TTL has actually expired.

**Example:** If the last cache touch was at 23:27, retention is `short` (5-min TTL), and the user runs `/lcm` at 23:40, it still shows `cache state: hot` — even though the Anthropic prompt cache expired 8 minutes ago.

## Root cause

`lcm-command.ts` line ~711:
```ts
buildStatLine("cache state", telemetry?.cacheState ?? "unknown")
```

This reads the stored DB field directly. The engine's `isPromptCacheStillHot()` correctly computes live TTL state for compaction decisions, but the display path does not use the same logic.

## Expected behavior

`/lcm` status should compute the live cache state by checking `(now - lastCacheTouchAt) < ttlMs`, mirroring the engine's `isPromptCacheStillHot()` + `resolvePromptCacheTtlMs()`. If the stored state is `hot` but TTL has expired, display `cold (ttl expired)` or similar.

## Impact

Display-only — compaction decisions already use the live TTL check via `shouldDelayPromptMutatingDeferredCompaction()`. But the stale display is confusing when debugging cache behavior.

## Suggested fix

Add a local helper in `lcm-command.ts` that:
1. Reads `telemetry.retention` to resolve TTL (`long`/`1h` → 3600s, `none` → null, default → `config.cacheAwareCompaction.cacheTTLSeconds`)
2. Gets the last touch time from `lastCacheTouchAt` → `lastObservedCacheHitAt` → `lastApiCallAt` (same fallback chain as the engine)
3. Compares `(now - lastTouch) < ttlMs`
4. If stored state is `hot` but TTL expired → display `cold (ttl expired)`

We have a working patch if helpful.

100yenadmin:
@liu51115 triage pass update: I marked this as priority:P3 because stale /lcm status cache display is confusing but not usually data-loss. Can you confirm whether current /lcm status still ignores TTL expiry?

Issue #436, opened 2026-04-14
[Cross-link]: Ollama cloud compaction summaries can return empty normalized output and degrade to truncation fallback


BeliefanX:
## Summary

Opening this as a cross-link from OpenClaw issue `openclaw/openclaw#66620`.

I do **not** currently think the primary bug lives inside `lossless-claw` itself. The stronger evidence points to an upstream OpenClaw / `@mariozechner/pi-ai` compatibility gap in the compaction summary completion path for Ollama cloud models.

Still, I think it is worth tracking here because from an operator perspective the symptom shows up as:

- `lossless-claw` compaction summary generation
- empty normalized summary
- fallback to deterministic truncation
- degraded summary graph quality

## What I observed

Environment:

- OpenClaw 2026.4.5 (stable)
- `lossless-claw` 0.8.2
- macOS arm64

Models tested as `summaryModel`:

- `ollama/devstral-small-2:24b-cloud`
- `ollama/qwen3-coder-next:cloud`
- `ollama/qwen3.5:cloud`

Normal chat with the same Ollama cloud models works.

Delegated expansion can also work.

But when LCM compaction actually tries to generate summaries, the logs show repeated empty extraction and truncation fallback:

```text
[lcm] empty normalized summary on first attempt; provider=ollama; model=qwen3-coder-next:cloud; block_types=(none); response_blocks=0; keys=content; content_kind=array; content_len=0; content_preview=[]; retrying with conservative settings
[lcm] retry also returned empty summary; provider=ollama; model=qwen3-coder-next:cloud; block_types=(none); response_blocks=0; keys=content; content_kind=array; content_len=0; content_preview=[]; falling back to truncation
[lcm] all extraction attempts exhausted; provider=ollama; model=qwen3-coder-next:cloud; source=fallback
[lcm] LCM compaction leaf pass (normal): 137628 -> 125865 conversation=1 summary=sum_7418e7965a4083d2
```

I reproduced the same pattern with `qwen3.5:cloud` as well.

## Why I think this is upstream

After digging through the installed code locally, two things stood out:

1. `lossless-claw` is using a `completeSimple`-style summary path and then normalizing text from fields like:
   - `text`
   - `output_text`
   - nested `content / summary / output / message / response`

[...]

Issue #439, opened 2026-04-14
Cache-aware compaction telemetry leaks hot-state across provider/model switches


jalehman:
## Summary

Lossless Claw currently stores cache-aware compaction telemetry as a single per-conversation state (`provider`, `model`, `cacheState`, `lastObservedCacheHitAt`, etc.). That lets hot-cache hysteresis leak across provider/model switches even though prompt cache validity is provider/model specific.

## Why This Matters

OpenClaw's prompt-cache observability treats provider/model changes as descriptive `promptCache.observation.changes`, not as an unconditional cache break. That is correct at the provider-cache level: if a conversation switches from provider A to provider B and then back to provider A within TTL, both providers may still have valid caches.

The problem is that Lossless Claw collapses that reality into one shared conversation-level cache state. A conversation that recently had a hot cache on one provider can continue to look effectively `hot` after switching providers, even though the hot-state evidence came from a different provider/model family.

## Current Behavior

- OpenClaw records provider/model changes as prompt-cache observation changes (`code: "model"`), but `observation.broke` is driven by observed `cacheRead` drops, not by the switch itself.
- Lossless Claw persists one telemetry row per conversation, including `provider`, `model`, `cacheState`, `lastObservedCacheHitAt`, `lastObservedCacheBreakAt`, and cold-observation streaks.
- Cache-aware incremental compaction uses that single row to decide whether to preserve cache (`hot-cache-budget-headroom`, `hot-cache-defer`) or catch up (`cold-cache-catchup`).

## Consequence

Cache-aware compaction policy can make decisions using hot-state telemetry that belongs to a different provider/model than the current run. This makes the cache-aware tuning harder to reason about and likely contributed to the recent deferred-compaction failures where large raw backlogs were preserved too aggressively.

## Suggested Direction

Simplify the model:

1. Scope cache-aware telemetry by provider/model family instead of by conversation alone.
2. Or, at minimum, treat a provider/model family mismatch as non-authoritative for hot-cache hysteresis.
3. Keep OpenClaw's provider-cache semantics as-is; do not force every provider switch to be a hard cache break.

## Related

- #362
- #408
- #434

100yenadmin:
@jalehman I triaged this as priority:P2 because provider/model cache telemetry leaking across contexts can make compaction decisions misleading. I did not find a direct merged fix. Can you confirm whether current main still exposes the stale cache telemetry state described here?

PR #479, opened 2026-04-21
feat: move condensed summaries to systemPromptAddition for prompt cache stability

rivetnode:
## Problem

When lossless-claw compacts context, it generates condensed summaries and injects them as **user-role messages** in the assembled context. This means:

1. Every compaction cycle changes the prefix of the conversation (new summary replaces old one)
2. The summary content — which is **semantically stable** (same information, just compressed) — breaks the **prompt cache** at providers like Anthropic and OpenAI
3. The entire conversation prefix must be re-processed each turn, even though the summary content is logically equivalent to what came before

In our production instance:
- System prompt: ~10,741 tokens (stable, cacheable)
- Condensed summaries injected as user messages: ~6,046 tokens (unstable, re-processed every turn)
- **Cacheable prefix**: only ~61% of conversation prefix → 39% re-processed unnecessarily

## Solution

Use the existing `systemPromptAddition` field in `AssembleResult` (already supported by OpenClaw's context engine interface and integration) to move condensed summaries from the user message stream into the **system prompt prefix**.

### Before
```
[System prompt - 10K tokens - CACHE STABLE]
[User: <summary kind="condensed">...6K tokens...</summary> - CACHE UNSTABLE]
[User: Hey, how are you?]
[Assistant: ...]
```

### After
```
[System prompt + condensed summary - 16K tokens - CACHE STABLE]
[User: Hey, how are you?]
[Assistant: ...]
```

## Why this is architecturally correct

1. **Summaries are system context, not user input.** They represent the agent's memory of past conversation — logically the same category as system prompts, workspace files, and SOUL.md
2. **The summary content is semantically stable.** Each compaction replaces the old summary with a new one covering the same information. The content changes in compression ratio, not in semantic meaning. This makes it ideal for the cacheable prefix.
3. **The ContextEngine interface already supports this.** `AssembleResult.systemPromptAddition?: string` exists and OpenClaw already handles it by prepending to the system prompt. We just need to populate it.

## Impact

For our production instance:
- **Cacheable prefix**: ~10,741 → ~16,787 tokens (**56% more stable prefix**)
- **Re-processed tokens per turn**: ~6,046 fewer tokens
- **Cache hit improvement**: From ~61% to ~85%+ of conversation prefix

For any user of prompt-caching providers (Anthropic, OpenAI, Google):
- Fewer tokens re-processed per turn → lower cost
- Faster time-to-first-token on cache hits
- Better cache utilization across the entire conversation lifecycle

## Implementation

The change is minimal — 40 lines in `engine.ts`:

1. After `assemble()`, split the assembled messages into condensed summaries and conversation messages
2. Render condensed summary messages as `systemPromptAddition`
3. Return only non-condensed messages in the `messages` array

**Only condensed summaries are moved.** Leaf summaries remain as user messages because they represent recent, detailed conversation chunks that change frequently and are not suitable for the stable prefix.

## Testing

- All 183 existing tests pass
- New test: `moves condensed summaries to systemPromptAddition for prompt cache stability`
  - Verifies `systemPromptAddition` contains condensed summary XML
  - Verifies condensed summaries no longer appear as user messages
- Existing test for leaf summaries still passes (they remain as user messages)

## Breaking change assessment

This is **not a breaking change**:
- The `AssembleResult` type already has `systemPromptAddition` as optional
- OpenClaw already handles it by prepending to system prompt
- Models receive the same content, just in a different position (system prompt vs user message)
- The semantic meaning is preserved — summaries are context, not conversation turns

## Related issues

- #436 (Ollama cloud compaction summaries can return empty normalized output) — this change does not affect the summary generation path, only the assembly output format
- #461 (`/lcm status` shows stale cache state) — related to cache state tracking, but orthogonal to this change

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -5348,9 +5348,50 @@ export class LcmContextEngine implements ContextEngine {
         );
       }
 
+      // Separate condensed summary messages from conversation messages.
+      // Condensed summaries represent stable context (the agent's memory of
+      // past turns) and are better placed in systemPromptAddition so they
+      // become part of the cacheable prompt prefix at providers like Anthropic
+      // and OpenAI.  This dramatically improves prompt cache hit rates because
+      // summaries change content on every compaction cycle but remain
+      // semantically equivalent — a cache-stable prefix means the entire
+      // conversation history up to the summary is only processed once.
+      //
+      // Only condensed summaries are moved; leaf summaries remain as user
+      // messages because they represent recent conversation chunks that are
+      // more specific and detailed.
+      const condensedSummaryMessages: AgentMessage[] = [];
+      const conversationMessages: AgentMessage[] = [];
+      for (const msg of assembled.messages) {
+        if (
+          msg.role === "user" &&
+          typeof msg.content === "string" &&
+          msg.content.startsWith("<summary ") &&
+          /kind="condensed"/.test(msg.content)
+        ) {
+          condensedSummaryMessages.push(msg);
+        } else {
+          conversationMessages.push(msg);
+        }
+      }
+
+      const systemPromptAddition =
+        condensedSummaryMessages.length > 0
+          ? condensedSummaryMessages
+              .map((msg) => (typeof msg.content === "string" ? msg.content : ""))
+              .join("\n\n")
+          : undefined;
+
+      if (condensedSummaryMessages.length > 0) {
+        this.deps.log.info(
+          `[lcm] assemble: moved ${condensedSummaryMessages.length} condensed summary message(s) to systemPromptAddition (${condensedSummaryMessages.reduce((sum, m) => sum + (typeof m.content === "string" ? m.content.length : 0), 0)} chars) conversation=${conversation.conversationId} ${sessionLabel}`,
+        );
+      }
+
       const result: AssembleResult = {
-        messages: assembled.messages,
+        messages: conversationMessages,
         estimatedTokens: assembled.estimatedTokens,
+        ...(systemPromptAddition ? { systemPromptAddition } : {}),
       };
       return result;
     } catch (err) {
--- a/test/engine.test.ts
+++ b/test/engine.test.ts
@@ -4675,7 +4675,7 @@ describe("LcmContextEngine.assemble canonical path", () => {
     expect(promptAddition).toBeUndefined();
   });
 
-  it("does not emit assembly-specific system prompt guidance when summaries are present", async () => {
+  it("does not emit assembly-specific system prompt guidance when leaf summaries are present", async () => {
     const engine = createEngine();
     const sessionId = "session-summary-guidance";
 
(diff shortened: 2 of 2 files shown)


rivetnode:
## Workaround for Ollama / self-hosted models (cache state)

For models that **do** support KV-cache reuse (like GLM-5 via Ollama) but where lossless-claw cannot detect cache state (no API header support), set `forceCacheState: hot` in the plugin config:

```json
{
  "plugins": {
    "entries": {
      "lossless-claw": {
        "config": {
          "cacheAwareCompaction": {
            "forceCacheState": "hot"
          }
        }
      }
    }
  }
}
```

This tells lossless-claw to treat the model as if prompt caching is always active, which:
- Prevents unnecessary re-compaction on every turn
- Keeps summaries stable (no flip-flopping between hot/cold)
- Is safe because self-hosted models with KV-cache reuse behave like "always hot"

**Note:** `cacheEligibleModelPatterns` (e.g. `[":cloud"]`) can also be used to scope which models are considered cache-eligible, but `forceCacheState: hot` is the simplest override for known-good models.

This is related to:
- #439 (cache-aware compaction leaks hot-state across provider switches)
- #367 (timing inversion based on last-call vs cache expiry)
- #409 (budget-aware deferral for unknown cache state)

A proper fix would be to detect KV-cache reuse via Ollama's `/api/generate` response headers, but that requires upstream changes in both Ollama and lossless-claw.

rivetnode:
## Updated: Models that benefit from `forceCacheState: hot"`

The `forceCacheState: hot` workaround applies to **any model that supports prefix/KV-cache reuse but where lossless-claw cannot detect it via API response headers**. Here are the ones we verified:

**Self-hosted (Ollama/vLLM):**
- **GLM-5 / GLM-5.1** via Ollama — KV-cache reuse between sequential requests ✅
- **DeepSeek V3** via vLLM — vLLM Automatic Prefix Caching (APC) reuses KV-cache blocks ✅
- **Qwen 3** via vLLM — same APC mechanism ✅
- **Llama 3/4** via vLLM/Ollama — same mechanism ✅

**Cloud APIs with automatic caching (no cache headers exposed):**
- **MiniMax M2.5/M2.7** — [automatic prefix caching](https://platform.minimax.io/docs/api-reference/text-prompt-caching) enabled by default, 512+ token threshold, prefix matching: tools → system → messages ✅
- **DeepSeek V3/R1** — [context caching on disk](https://api-docs.deepseek.com/guides/kv_cache/) enabled by default, prefix overlap reuse ✅
- **Kimi K2.5** via Fireworks — [prompt caching](https://docs.fireworks.ai/guides/prompt-caching) supported on Fireworks, exact prefix match ✅
- **Kimi K2.5** via Azure AI Foundry — [prompt caching enabled by default](https://learn.microsoft.com/en-us/answers/questions/5764725/) ✅

**Cloud APIs with detectable caching (no workaround needed):**
- **Anthropic Claude** — `cache_creation_input_tokens` / `cache_read_input_tokens` in response
- **OpenAI GPT-4o/o1** — `cached_tokens` in `prompt_tokens_details`
- **Google Gemini** — `cachedContentTokenCount` in usage metadata

**No caching (forceCacheState not applicable):**
- **Groq** — no prompt caching
- **Cerebras** — no prompt caching

### Config example for multi-model setups

```json
{
  "plugins": {
    "entries": {
      "lossless-claw": {
        "config": {
          "cacheAwareCompaction": {
            "forceCacheState": "hot",
            "cacheEligibleModelPatterns": [":cloud", "minimax", "deepseek", "kimi"]
          }
        }
      }
    }
  }
}
```

`cacheEligibleModelPatterns` scopes which models are considered cache-eligible. Combined with `forceCacheState: hot`, it prevents lossless-claw from re-compacting on every turn for these models.

### Why this matters for our PR

Moving condensed summaries to `systemPromptAddition` creates a larger **stable prefix**. When prompt caching is active (detected or forced), the entire system prompt + condensed summaries stay cached — no re-processing. Without `forceCacheState: hot`, models like MiniMax M2.5 or DeepSeek V3 would be treated as "unknown cache state" and lossless-claw may over-compact or flip-flop, negating the cache benefit of this PR.

The combination of this PR + `forceCacheState: hot` for non-header-detectable models = maximum cache efficiency.

PR #513, opened 2026-04-28
fix(lcm-expand-query): account for leaf summary tokens in totalSourceTokens

### LINKED ISSUE #487: bug: lcm_expand_query reports zero source tokens for explicit leaf summaries
## What

`lcm_expand_query` can return a correct answer with `expandedSummaryCount > 0` while reporting `totalSourceTokens: 0` when the selected evidence is an explicit leaf summary.

## Repro

In the live OpenClaw session after switching the gateway to native Codex harness, this smoke test returned a correct answer but misleading accounting:

```json
{
  "summaryIds": ["sum_78585f222b7ad42e"],
  "prompt": "What was the immediate next action for PR #70809?"
}
```

Observed shape:

```json
{
  "answer": "Have Adele rebase onto the current remote PR #70809 branch and push her stale-history fix again.",
  "citedIds": ["sum_78585f222b7ad42e"],
  "expandedSummaryCount": 1,
  "totalSourceTokens": 0,
  "truncated": false
}
```

`lcm_describe` for that same summary showed it was a leaf with its own summary tokens:

```text
kind=leaf tok=1063 descTok=0 srcTok=39933 desc=0
```

A comparison against a non-leaf summary returned nonzero accounting:

```json
{
  "summaryIds": ["sum_473a179b8410363e"],
  "expandedSummaryCount": 3,
  "totalSourceTokens": 3591
}
```

## Why this happens

This appears to be lossless-claw accounting/semantics, not a Codex harness failure:

- `lcm_expand_query` delegates to a child agent and asks it to return `totalSourceTokens` in JSON.
- The parent parser trusts the child-provided number.
- The child prompt defines `totalSourceTokens` as tokens consumed from `lcm_expand` calls.
- Explicit summary IDs are added with `requiresMessageExpansion: false`.
- `lcm_expand` on a leaf summary with `includeMessages=false` returns no child summaries/messages, so its `totalTokens` can be `0`.
- The child can still answer from `lcm_describe`, which includes the leaf summary content, but those tokens are not counted in `totalSourceTokens`.

Relevant files:

- `src/tools/lcm-expand-query-tool.ts`
- `src/tools/lcm-expand-tool.ts`
- `src/retrieval.ts`

## Expected

The accounting should not imply zero source context was used when the answer came from a leaf summary. Possible fixes:

- Rename/clarify the field as expansion-only tokens.
- Include `lcm_describe` summary token counts in `totalSourceTokens`.
- Treat explicit leaf summaries as requiring message expansion or otherwise account for the leaf summary token count.

## Notes

A non-leaf `lcm_expand_query` returned nonzero `totalSourceTokens` under the current Codex harness, so this does not look like a blanket delegated-tool or Codex harness wiring problem.

mvanhorn:
## Summary

Closes #487.

When an answer was derived from an explicit leaf summary, the delegated child agent in `lcm_expand_query` reported `totalSourceTokens: 0` even though the leaf's content was actually consumed (via `lcm_describe`). Reproducer from the issue:

```json
{
  "answer": "Have Adele rebase onto the current remote PR #70809 branch...",
  "citedIds": ["sum_78585f222b7ad42e"],
  "expandedSummaryCount": 1,
  "totalSourceTokens": 0,
  "truncated": false
}
```

The leaf had `tok=1063` per `lcm_describe`, but neither the explicit-leaf path (`requiresMessageExpansion: false`) nor `lcm_expand` (returns 0 tokens for childless leaves) surface those tokens to the accounting.

## Fix

Per the issue's "Possible fixes" list (option 1 + option 2 combined), update the delegated child agent's prompt rule for `totalSourceTokens` to explicitly include leaf-summary tokens from `lcm_describe` in the count:

> totalSourceTokens should estimate the total source tokens consumed for retrieval. Include both: (a) the `totalTokens` returned by each `lcm_expand` call you made, AND (b) for any explicit leaf summary used as evidence whose `lcm_expand` returned 0 tokens because the leaf has no children to expand, the leaf summary's own `tok` value from `lcm_describe`. This avoids reporting `totalSourceTokens: 0` when the answer was actually derived from a leaf summary's content.

This is a delegated-prompt-only change in `src/tools/lcm-expand-query-tool.ts`. No retrieval logic moves, no public API change, no SDK or schema change. The accounting becomes honest about what was actually consumed when the evidence is a leaf summary.

## Files

- `src/tools/lcm-expand-query-tool.ts` - one prompt rule replaced (1 line changed).

## Verification

- `npm run build` clean.
- `npm test` 44 test files / 795 tests passing.
- Diff is +1/-1 (one prompt rule rewritten).

## Out of scope

The other two options from the issue (renaming the field outright, or treating explicit leaves as `requiresMessageExpansion: true`) are bigger semantic changes and would touch consumers/output shape. The prompt-only fix here addresses the reporter's "misleading accounting" symptom without breaking the API contract. Happy to follow up with a deeper refactor if the maintainer prefers a different option.

### DIFF
--- a/.changeset/count-leaf-summary-tokens.md
+++ b/.changeset/count-leaf-summary-tokens.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Correct `lcm_expand_query` source-token accounting guidance for explicit leaf summaries.
--- a/src/tools/lcm-expand-query-tool.ts
+++ b/src/tools/lcm-expand-query-tool.ts
@@ -300,7 +300,7 @@ function buildDelegatedExpandQueryTask(params: {
     `- Keep answer concise and focused (target <= ${params.maxTokens} tokens).`,
     "- citedIds must be unique summary IDs.",
     "- expandedSummaryCount should reflect how many summaries were expanded/used.",
-    "- totalSourceTokens should estimate total tokens consumed from expansion calls.",
+    "- totalSourceTokens should estimate the total source tokens consumed for retrieval. Include both: (a) the `totalTokens` returned by each `lcm_expand` call you made, AND (b) for any explicit leaf summary used as evidence whose `lcm_expand` returned 0 tokens because the leaf has no children to expand, the leaf summary's own `tok` value from `lcm_describe`. This avoids reporting `totalSourceTokens: 0` when the answer was actually derived from a leaf summary's content.",
     "- truncated should indicate whether source expansion appears truncated.",
   ]
     .filter((line): line is string => typeof line === "string")
--- a/src/tools/lcm-expand-query-tool.ts
+++ b/src/tools/lcm-expand-query-tool.ts
@@ -300,7 +300,7 @@ function buildDelegatedExpandQueryTask(params: {
     `- Keep answer concise and focused (target <= ${params.maxTokens} tokens).`,
     "- citedIds must be unique summary IDs.",
     "- expandedSummaryCount should reflect how many summaries were expanded/used.",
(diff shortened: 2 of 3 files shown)


jalehman:
Thank you!

Issue #534, opened 2026-04-29
Expose assembly provenance/cache policy and treat openai-codex cache windows as mutation-sensitive


100yenadmin:
## Summary

`lossless-claw`/LCM appears to assemble prompt-ready compact context correctly, but the current context-engine result surface is too coarse for the host to reason safely about assembly provenance, fallback mode, raw-history debt, and prompt-cache sensitivity. In the observed OpenClaw integration, this contributed to ambiguity around whether the assembled context should be prompt-authoritative and whether Codex prompt-cache windows should delay prompt-mutating deferred compaction.

This is a companion hardening issue to the OpenClaw host-side bug filed as openclaw/openclaw#74229. The primary live reset loop appears OpenClaw-side, but LCM can make the contract harder to misuse.

## Environment observed

- Installed plugin: `@martian-engineering/lossless-claw@0.9.2`
- Active OpenClaw model: `openai-codex/gpt-5.5`
- Active context window: `258000`
- LCM config highlights:
  - `freshTailCount=64`
  - `freshTailMaxTokens=16000`
  - `proactiveThresholdCompactionMode=deferred`
  - `cacheAwareCompaction.enabled=true`
  - `cacheAwareCompaction.cacheTTLSeconds=300`
  - `promptAwareEviction=false`

## Confidence matrix

- **0.88 - issue:** LCM assembly is intended to produce prompt-ready compact context, but the returned SDK surface does not expose enough metadata for hosts to distinguish authoritative assembly from fallback live context or emergency recovery.
- **0.84 - issue:** prompt-cache mutation sensitivity is still Anthropic/Claude-shaped in the current source lineage, while live telemetry shows `openai-codex/gpt-5.5` has active hot cache signals.
- **0.84 - live evidence:** DB telemetry shows active `openai-codex/gpt-5.5`, `cache_state=hot`, `retention=long`, `last_observed_cache_read=59904`, and prompt token telemetry.
- **0.75 - contributing cause:** session reset/rotation in the host makes exact at-failure LCM attribution harder because the DB tracks the active `sessionKey` through reset rather than preserving a clean per-session assembly snapshot.
- **0.70 - possible cause:** the installed plugin is bundled/minified and does not preserve source-level provenance, making live debugging harder when multiple local PR worktrees exist.
[...]

PR #535, opened 2026-04-29
fix(cache): treat Codex prompt cache as mutation-sensitive

100yenadmin:
## Summary

Refs Martian-Engineering/lossless-claw#534 and openclaw/openclaw#74233. Companion host-side PR: openclaw/openclaw#74255.

This PR fixes the LCM cache-policy side of the OpenClaw/LCM context instability investigation. LCM already delayed some prompt mutations around Anthropic prompt-cache windows, but the policy was Anthropic/Claude-specific. OpenClaw's Codex Responses route can also have cache-sensitive prompt windows, especially with provider identifiers such as `openai-codex`, `openai-codex-responses`, and `github-copilot`.

The practical failure mode: LCM could see a cache-write-only turn, or a still-recent cache touch, and still allow prompt-leaf compaction to rewrite the prompt too soon. That is exactly the kind of prompt mutation that can break cache reuse and make subsequent context/cost behavior look erratic.

## Three-PR Map

```mermaid
flowchart LR
  A["OpenClaw #74255\nPrompt authority"] --> D["Stop raw history from\nfalse-failing fitted prompts"]
  B["LCM #535\nCache mutation policy"] --> E["Delay deferred prompt rewrites\nwhile cache is still hot"]
  C["LCM #536\nAssembly provenance"] --> F["Expose assembly/fallback facts\nto the host"]
  B --> C
```

- openclaw/openclaw#74255 is the main production fix: the host should hard-check the assembled prompt, not the full raw transcript, unless explicitly opted in.
- #535 is cache hardening: Codex/OpenClaw prompt-cache providers now get the same mutation-sensitive protection Anthropic had.
- #536 is observability: LCM returns compact provenance so future cache/context blow-ups can be diagnosed without log archaeology.

## Root Cause

The old policy asked a provider-family question that was too narrow: "is this Anthropic/Claude?" That missed Codex/OpenClaw provider shapes even though they can be prompt-cache sensitive in the same operational way.

There were also hot/cold classification edge cases. A turn with only cache writes is not cold; it is evidence that the cache was just touched. A recent `lastCacheTouchAt` inside the retention window should also be treated as cache-sensitive unless LCM has observed a fresh explicit cache break. If a newer cache touch arrives after an older break observation, the old break must not remain sticky forever.

## What This Changes

- Replaces the Anthropic-only family check with a mutation-sensitive prompt-cache family check.
- Covers Anthropic/Claude plus OpenClaw Codex provider identifiers, including `openai-codex`, `openai-codex-responses`, `github-copilot`, and Codex CLI spelling variants.
- Treats cache-write-only usage as hot/cache-sensitive.
- Keeps cache-write-only telemetry hot in the inline incremental-compaction path, rather than letting a zero cache-read share force cold-cache catch-up.
- Treats recent `lastCacheTouchAt` within the retention TTL as hot for mutation-delay purposes.
- Preserves fresh explicit cache-break observations as cold so known breaks still allow catch-up compaction.
- Clears stale break pressure when a newer cache touch has been observed.
- Applies the generalized delay and post-TTL catch-up behavior anywhere the old policy was Anthropic-only.
- Adds a patch changeset.

## What The Policy Does

This does not disable compaction. It changes when deferred prompt-leaf mutations are allowed to run:

- cache read or cache write observed: delay prompt mutation while the cache window is hot
- cache-write-only turn: still delay, because the prompt cache was just touched
- recent `lastCacheTouchAt`: delay until retention expires
- fresh explicit cache break: treat as cold and allow catch-up work
- newer cache touch after an old break: treat as hot again
- TTL expired: allow deferred work to run

That gives Codex/OpenClaw cache windows the same protection shape already intended for Anthropic.

## Validation

- `npx vitest run test/engine.test.ts -t "low cache-read share|cache-write-only telemetry|Codex cache-write-only|explicit Codex cache breaks|breaks as hot again|prompt cache TTL"`
- `npm test`
- `npm run build`
- `npm run release:verify`

## Review Notes

The key review question is whether the generalized provider-family predicate is broad enough to catch the OpenClaw Codex paths without making explicit cache-break observations sticky. The tests cover cache read, cache write only, inline cache-write-only state, explicit break, break-then-newer-touch, hot retention TTL, expired TTL, Anthropic, and Codex provider cases.

### DIFF
--- a/.changeset/codex-cache-mutation-policy.md
+++ b/.changeset/codex-cache-mutation-policy.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Treat Codex prompt-cache writes and recent cache touches as mutation-sensitive so deferred compaction does not rewrite hot cached prompts before the cache TTL expires.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -2089,13 +2089,23 @@ export class LcmContextEngine implements ContextEngine {
     return Math.max(1, this.config.cacheAwareCompaction.cacheTTLSeconds) * 1000;
   }
 
-  /** Detect Anthropic-family sessions where local prompt rewrites can invalidate a hot prefix cache. */
-  private isAnthropicPromptCacheFamily(
+  /** Detect prompt-cache families where local prompt rewrites can invalidate a hot prefix cache. */
+  private isPromptCacheMutationSensitiveFamily(
     telemetry: ConversationCompactionTelemetryRecord | null,
   ): boolean {
     const provider = telemetry?.provider?.trim().toLowerCase() ?? "";
     const model = telemetry?.model?.trim().toLowerCase() ?? "";
-    return provider.includes("anthropic") || model.includes("claude");
+    const identifiers = [provider, model];
+    return identifiers.some((identifier) =>
+      identifier.includes("anthropic")
+      || identifier.includes("claude")
+      || identifier.includes("openai-codex")
+      || identifier.includes("openai_codex")
+      || identifier.includes("github-copilot")
+      || identifier.includes("github_copilot")
+      || identifier.includes("codex-cli")
+      || identifier.includes("codex_cli")
+    );
   }
 
   /** Determine whether the last prompt-cache touch is still within the active TTL window. */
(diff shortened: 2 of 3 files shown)


100yenadmin:
@jalehman

PR #536, opened 2026-04-29
feat(assemble): expose LCM assembly provenance

100yenadmin:
## Summary

Refs Martian-Engineering/lossless-claw#534 and openclaw/openclaw#74233. Companion host-side PR: openclaw/openclaw#74255.

This PR makes LCM assembly provenance host-readable. LCM already had useful debug facts in logs, but OpenClaw hosts generally received only the returned `messages` and `estimatedTokens`. That meant future context blow-ups, cache breaks, or fallback paths could be visible in local logs but not in the host-side result that made the prompt decision.

The change is backward-compatible: older hosts can ignore the extra fields, while newer hosts can consume `promptAuthority: "assembled"` and optionally read namespaced `lcmAssembly` diagnostics to understand exactly what kind of prompt LCM returned.

## Three-PR Map

```mermaid
flowchart LR
  A["OpenClaw #74255\nConsumes prompt authority"] --> D["Host trusts assembled prompt\nby default"]
  B["LCM #535\nProtects hot cache windows"] --> E["Fewer accidental\nprompt-cache breaks"]
  C["LCM #536\nEmits assembly provenance"] --> F["Host can see mode, fallback,\ncounts, tokens, cache, hashes"]
  C --> A
```

- openclaw/openclaw#74255 fixes the host behavior that rejected fitted assembled prompts because raw history was large. It consumes the `promptAuthority` contract.
- #535 hardens when LCM is allowed to mutate prompt leaves around Codex/OpenClaw cache windows.
- #536 gives hosts compact proof of what LCM assembled and why, so the next incident report can point to metadata instead of reconstructing behavior from logs. The `lcmAssembly` object is namespaced diagnostic metadata and remains safe for older hosts to ignore.

## Root Cause

Without host-visible provenance, three very different outcomes looked too similar from OpenClaw's side:

- LCM assembled a normal fitted prompt.
- LCM intentionally fell back to live messages because there was no usable context record.
- LCM hit an edge fallback, such as DB-trails-live, empty output, no user turn, or an assembler error.

Those distinctions matter when diagnosing context growth, cache behavior, and apparent forgetfulness. The host needs to know whether `messages` came from normal assembly or from a fallback path, and whether those returned messages are the actual prompt authority.

## What This Changes

- Returns `promptAuthority: "assembled"` for normal LCM assembly and live fallback results, because returned `messages` are the actual prompt the host should evaluate.
- Adds namespaced `lcmAssembly` provenance using an intersection type so older OpenClaw hosts ignore it safely.
- Includes compact diagnostics: assembly mode, fallback reason, input/output message counts, context item counts, estimated tokens, cache state, selection mode, and stable hashes when available.
- Estimates fallback tokens from the actual fallback messages, so non-empty live fallbacks no longer report zero-token diagnostics.
- Preserves existing log diagnostics; this PR exposes the same facts to the host instead of making logs the only evidence source.
- Covers normal assembly plus fallback paths for ignored sessions, missing conversations, no context items, DB-trails-live, empty assembled output, no user turn, and assembler errors.
- Adds a patch changeset.

## What The Provenance Fields Do

`promptAuthority` tells the host which prompt surface should be authoritative for hard overflow decisions. In normal LCM behavior this is `"assembled"`, meaning the returned `messages` should be checked, not the raw pre-assembly transcript.

`lcmAssembly.mode` tells the host whether LCM returned a normal assembled prompt or a live fallback.

`lcmAssembly.fallbackReason` explains why a fallback happened, when there is one.

The count/token/cache/hash fields give enough stable evidence to compare incidents without exposing full prompt contents. They are diagnostics, not behavior changes.

## Validation

- `npx vitest run test/engine.test.ts -t "assemble canonical path"`
- `npm test`
- `npm run build`
- `npm run release:verify`

## Review Notes

This PR can merge independently from #535. Its host-side prompt-authority consumer is openclaw/openclaw#74255, but the full provenance payload remains compatible with older hosts because the new metadata is optional and namespaced.

### DIFF
--- a/.changeset/lcm-assembly-provenance.md
+++ b/.changeset/lcm-assembly-provenance.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Expose assembly provenance and prompt-authority metadata from LCM assemble results so hosts can distinguish assembled prompts from live fallback paths.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -6121,9 +6121,12 @@ export class LcmContextEngine implements ContextEngine {
       }
 
       const contextItems = await this.summaryStore.getContextItems(conversation.conversationId);
-      const summaryContextItemCount = contextItems.filter(
-        (item) => item.itemType === "summary",
-      ).length;
+      let summaryContextItemCount = 0;
+      for (const item of contextItems) {
+        if (item.itemType === "summary") {
+          summaryContextItemCount += 1;
+        }
+      }
       const rawContextItemCount = contextItems.length - summaryContextItemCount;
       const contextDiagnostics = {
         contextItemCount: contextItems.length,
@@ -6184,13 +6187,7 @@ export class LcmContextEngine implements ContextEngine {
         this.deps.log.info(
           `[lcm] assemble: assembled context has no user turns, falling back to live context to prevent prefill errors conversation=${conversation.conversationId} ${sessionLabel} assembledMessages=${assembled.messages.length} duration=${formatDurationMs(Date.now() - startedAt)}`,
         );
-        return buildAssembleResult({
-          messages: params.messages,
-          estimatedTokens: 0,
-          mode: "live_fallback",
-          fallbackReason: "no_user_turn",
-          ...contextDiagnostics,
-        });
+        return safeFallback("no_user_turn", contextDiagnostics);
       }
 
       this.deps.log.info(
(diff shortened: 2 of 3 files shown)


100yenadmin:
@jalehman

100yenadmin:

**Summary**
@100yenadmin, this PR is still potentially useful, but the current head `63244cf2ace72c0b0fb4521fcc8f9fb1a60f5751` is `CONFLICTING` / `DIRTY` against `main`, so it cannot be reviewed for merge-readiness yet.

**Next step**

<details>

- PR title: feat(assemble): expose LCM assembly provenance
- Current labels: bug, enhancement, priority:P2

</details>

PR #546, opened 2026-05-01
fix: Resolve DeepSeek (and other providers) API routing when model catalog lacks baseUrl/api

baghvn:
## Problem

The lossless-claw plugin relies on OpenClaw's model catalog to resolve two critical routing fields when building `completeSimple` options:

1. **`api`** — the API protocol family (e.g. `"openai-completions"`, `"anthropic-messages"`)
2. **`baseUrl`** — the HTTP endpoint for the provider

When a provider is configured via `openclaw.json` `models.providers` without declaring these fields explicitly, and the installed model catalog does not contain entries for those providers, the plugin silently falls back to:

- `api`: `undefined` (which pi-ai resolves to a default, often mismatched)
- `baseUrl`: `""` (empty string, which routes to `https://api.openai.com/v1`)

This causes silent **401 Unauthorized** errors for providers like DeepSeek, Groq, Together, and Mistral that have their own API endpoints but use the OpenAI-compatible protocol.

## Root Cause

- The `inferApiFromProvider` map only covered `openai`, `openai-codex`, `google`, and `anthropic`. Any other provider returned `undefined`, leading pi-ai to use its own default API resolution.
- The `baseUrl` fallback was always `""`, meaning requests for providers without an explicit model catalog URL were routed to OpenAI's endpoint.

### 1. `inferApiFromProvider`: Add `deepseek` (60a8f8e)

Extended the existing `inferApiFromProvider` map:

```typescript
deepseek: "openai-completions",
```

This tells the plugin to use the OpenAI-compatible completions API for DeepSeek models, matching how DeepSeek's API surface works.

### 2. `inferBaseUrlFromProvider`: Known provider base URL map (this PR)

Added a new `inferBaseUrlFromProvider(provider)` function that maps well-known providers to their correct API base URLs:

| Provider     | Base URL                                         |
|--------------|--------------------------------------------------|
| deepseek     | `https://api.deepseek.com`                       |
| anthropic    | `https://api.anthropic.com`                      |
| openai       | `https://api.openai.com/v1`                      |
| openai-codex | `https://api.openai.com/v1`                      |
| google       | `https://generativelanguage.googleapis.com/v1beta`|
| groq         | `https://api.groq.com/openai/v1`                 |
| mistral      | `https://api.mistral.ai`                         |
| together     | `https://api.together.xyz`                       |
| openrouter   | `https://openrouter.ai/api/v1`                   |
| ollama       | `http://localhost:11434`                         |

This is used as the final fallback in two code paths where empty string was previously the default:

- **Known-model route** (lines ~1664-1667): when `providerLevelConfig.baseUrl` is not a string and `knownModel.baseUrl` is not a string
- **Fallback route** (lines ~1693-1696): when `providerLevelConfig.baseUrl` is not a string and the model was resolved from config without a URL

## Verification

Build: **813/813 tests passing**.

In production at 99% cache hit rate, the fix was verified by observing successful DeepSeek compaction:

```
[lossless-claw] Compaction summarization model: deepseek/deepseek-v4-flash (default)
[lossless-claw] LCM compaction leaf pass (normal): 66363 -> 46884
```

## Notes

- The `baseUrl` map stores the endpoint without trailing slashes where possible, consistent with how OpenClaw's model catalog formats URLs.
- Providers using OpenAI-compatible endpoints share the same `"openai-completions"` api value.
- The previous commit (60a8f8e) added the api fix; this commit adds the parallel baseUrl fix. Both are needed together for a provider like DeepSeek that is not in the default model catalog.

## Related

- Closes gap for any OpenAI-compatible provider not in the model catalog
- Follows the same pattern as the existing `inferApiFromProvider` function
- Together with 60a8f8e, fully resolves DeepSeek integration

### DIFF
--- a/.changeset/deepseek-provider-routing.md
+++ b/.changeset/deepseek-provider-routing.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Resolve uncataloged DeepSeek and other known provider models to their expected API family and base URL defaults when OpenClaw model metadata is unavailable.
--- a/src/plugin/index.ts
+++ b/src/plugin/index.ts
@@ -747,6 +747,7 @@ function inferApiFromProvider(provider: string): string | undefined {
   const normalized = normalizeProviderId(provider);
   const map: Record<string, string> = {
     anthropic: "anthropic-messages",
+    deepseek: "openai-completions",
     openai: "openai-responses",
     [OPENAI_CODEX_PROVIDER_ID]: OPENAI_CODEX_RESPONSES_API,
     "github-copilot": OPENAI_CODEX_RESPONSES_API,
--- a/src/plugin/index.ts
+++ b/src/plugin/index.ts
@@ -749,9 +749,13 @@ function inferApiFromProvider(provider: string): string | undefined {
   const map: Record<string, string> = {
     anthropic: "anthropic-messages",
     deepseek: "openai-completions",
+    groq: "openai-completions",
+    mistral: "openai-completions",
     openai: "openai-responses",
     [OPENAI_CODEX_PROVIDER_ID]: OPENAI_CODEX_RESPONSES_API,
     "github-copilot": OPENAI_CODEX_RESPONSES_API,
+    openrouter: "openai-completions",
+    together: "openai-completions",
     google: "google-generative-ai",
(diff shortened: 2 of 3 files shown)


jalehman:
Thank you!

PR #555, opened 2026-05-03
fix(manifest): declare contracts.tools so the plugin loads on OpenClaw 2026.5.2

jeremyheslop:
OpenClaw 2026.5.2 tightened plugin validation in `loader-CiaemmFD.js`'s `registerTool`: every tool name a plugin registers must appear in `contracts.tools` of `openclaw.plugin.json`. Without it the loader emits

    plugin must declare contracts.tools before registering agent tools

and rejects all four `lcm_*` registrations. The engine itself still initializes (logs `[lcm] Plugin loaded ... Engine initialized`), but because none of the agent tools wire up, compaction never runs end to end — sessions silently grow until they hit context limits.

Declaring the four tools the plugin actually registers (`lcm_grep`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`) restores normal loading. After applying this on a 2026.5.2 install:

    $ openclaw plugins doctor
    No plugin issues detected.

### DIFF
--- a/.changeset/declare-contracts-tools.md
+++ b/.changeset/declare-contracts-tools.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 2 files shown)


jalehman:
Thank you!

PR #599, opened 2026-05-04
fix: provider routing fallbacks, inferBaseUrlFromProvider, and sessions_yield dedup

baghvn:
## Problem

When the agent produces a message and calls `sessions_yield()`, the OpenClaw runtime re-ingests the same assistant message into LCM. This causes duplicate Telegram messages — every `sessions_yield` call triggers a replay of the last assistant message with a 0-second timestamp gap.

## Root Cause

The `ingestSingle` method in `LcmContextEngine` had no dedup check. The yield-abort cycle causes the runtime to re-send the same content to LCM's `ingest` path, which faithfully persists it as a new message.

## Fix

Added a **60-second time-window content-hash dedup** in `ingestSingle()`: before persisting, check if the last message in the same conversation has the exact same `role + content` and was written within the last 60 seconds. If so, silently skip (`return { ingested: false }`).

Key decisions:
- **60s window:** generous enough to cover any yield/abort replay, short enough to not suppress genuine duplicates across separate turns
- **Role + content check:** uses existing `messageIdentity()` function (SHA256-based in `message-identity.ts`), same dedup strategy already used in `reconcileSessionTail()`
- **Applies to both assistant and user messages** (user messages from the same input storm could theoretically hit the same race)
- **No new dependencies** — uses existing `getLastMessage()`, `messageIdentity()`, and `createdAt` on `MessageRecord`

Also includes:
- `inferBaseUrlFromProvider()` — resolves known provider base URLs when model catalog lookup misses (DeepSeek, Groq, Mistral, OpenRouter, Together, etc.)
- Complete provider routing fallbacks — covers Groq, Mistral, and other providers missing from `inferApiFromProvider()`
- Ollama cloud baseUrl test — verifies user-configured `https://ollama.com` is preserved and not silently overridden
- `isExplicitlyConfigured` guard in `shouldUseNativeCodexBaseUrl` — prevents paid OpenAI API-key Codex users from having their explicitly-set `https://api.openai.com/v1` baseUrl silently overridden with the native Codex URL (fixes CI regression)
- `contracts.tools` declaration in `package.json` — fixes a startup error on OpenClaw 2026.5.4 where the plugin loader rejects tool registrations from plugins that haven't declared `contracts.tools`. Without this, all four LCM recall tools (`lcm_grep`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`) log an error on every gateway startup. Neither `upstream/main` nor the published `0.9.4` package includes this declaration yet.

## Testing

- Build: passes (esbuild, no type errors)
- Tests: all pass locally
- Verified dedup logic: `getLastMessage()` pattern, `messageIdentity()` comparison, and `createdAt.getTime()` time delta
- Verified `contracts.tools` fix: startup error absent from gateway logs after restart

## Evidence

6 duplicate pairs confirmed in LCM DB with 0-second gaps, all following the yield pattern.

## Related

Fixes CHE-249

---

*— Submitted by Baghavan 🇲🇾 Fun fact: Malaysia's total highway network is longer than the circumference of the Earth.*

### DIFF
--- a/.changeset/deepseek-provider-routing.md
+++ b/.changeset/deepseek-provider-routing.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Resolve uncataloged DeepSeek and other known provider models to their expected API family and base URL defaults when OpenClaw model metadata is unavailable.
--- a/package.json
+++ b/package.json
@@ -67,6 +67,9 @@
     "extensions": [
       "./dist/index.js"
     ],
+    "contracts": {
+      "tools": ["lcm_grep", "lcm_describe", "lcm_expand", "lcm_expand_query"]
+    },
     "compat": {
       "pluginApi": ">=2026.2.17 <2026.6.0",
       "minGatewayVersion": "2026.2.17",
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -6014,6 +6014,25 @@ export class LcmContextEngine implements ContextEngine {
       stored = rawPayloadIntercepted.stored;
     }
 
+    // ── Time-window content-hash dedup ────────────────────────────────────────
+    // Prevent duplicate assistant messages caused by sessions_yield() replay.
+    // When the agent produces a message and then yields, the runtime re-persists
+    // the same content into LCM. Skip if the last message in this conversation
+    // has the exact same role + content and was written within the last 60s.
+    if (stored.role === "assistant" || stored.role === "user") {
+      const lastMsg = await this.conversationStore.getLastMessage(conversationId);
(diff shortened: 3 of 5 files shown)


jalehman:
Thanks for putting this together, Vishnu. We reviewed this again, and main has moved substantially since the PR was originally opened. At this point the branch is carrying several stale changes that overlap with newer behavior on main, so we are going to close this PR rather than continue trying to rebase it.

If there are still specific issues here that need attention, please open separate, targeted PRs for each fix. Smaller PRs will make it much easier for us to review the intended behavior, preserve the current main-branch semantics, and give focused feedback on each issue.

jalehman:
Closing per the note above. Thanks again for the contribution.

PR #626, opened 2026-05-07
(ignore) feat(v4.2): stub-tier stratification — externalize old tool results, agent drills down via lcm_describe(file_xxx)

100yenadmin:
> **Stacked on top of #613** (LCM v4.1 omnibus). PR #626 has 9 commits + ~2,140 LOC of v4.2 delta over `electricsheephq:feat/lcm-v4.1-omnibus` head (`536784c`). Reviewers can either wait for #613 to merge then this rebases onto `main` cleanly, or review the delta directly via `git diff 536784c...feat/lcm-stub-tier-stratification`.

## The problem this solves

When a long session pushes against the token budget at assemble time, v4.1's only lever for evictable items is "drop the whole row." Heavy tool results (12K+ tokens for a verbose `Read`/`Bash`/`Grep`) force the budget into a bad choice:

- Keep the heavy old tool result → lose 3 older summaries
- Drop the tool result → lose its assistant pairing and meaning too

Measured on a real DB (live snapshot, 2.6 GB, 315k messages), session `0cb8928b` at 258k budget: chronological eviction kept **333 items**.

## What this PR does

Adds a per-row sidecar (`messages.large_content`) that stores a `file_xxx` id pointing to the externalized payload in `large_files` (existing v4.1 storage table). At assemble time, evictable tool-result rows with the sidecar populated are replaced with the v4.1 `[LCM Tool Output: file_xxx | tool=… | N bytes]` reference format that's been in production for months. Drilldown uses the existing `lcm_describe(id="file_xxx")` path.

```
[LCM Tool Output: file_85b322f5fda14187ab641331 | tool=exec | 34,975 bytes]

Exploration Summary:
Tool: exec | Command: bash -lc 'cd /Users/lume/.openclaw/workspace && echo "== selfheal ==" && sed -n "1,260p" scripts/evaos-support/selfheal.sh && echo "== config-guard ==" …

Use lcm_describe with the file id to inspect the full output.
```

The `Exploration Summary` line carries a one-line preview of the originating `tool_input` (path / command / pattern / sessionId) so an agent reading the conversation can match a user reference like "the selfheal.sh script you read earlier" to the right fileId, then drill down.

## Architecture

| Layer | File | What |
|---|---|---|
| Schema | `src/db/migration.ts` | `messages.large_content TEXT` (idempotent ALTER); `PRAGMA busy_timeout=30000` before `BEGIN EXCLUSIVE` to coexist with running gateway |
[...]

PR #613, opened 2026-05-05
feat(lcm): v4.1 —LCM V2 (replaces #516; companion #616 deferred)

100yenadmin:
# LCM v2 (iteration experiments v4.1) — Lossless Agent Memory

**77 commits · 1502 tests passing · 10 audit waves closed · live-DB verified against the user's 2.6 GB / 4187-leaf corpus**

> Replaces the rollup approach from #516. Companion draft #616 preserves cut features.

This PR rebuilds LCM the way a person actually remembers: keep the raw conversation forever, embed it for similarity search, and synthesize fresh views on demand. The v3 `lcm_recent` rollup approach (summaries-of-summaries-of-summaries) is removed because it produced repetitive, lossy output that got worse the further back you looked.

After merge, the agent can answer every one of these without operator intervention:

| Question | Answer path |
|---|---|
| "What did we work on yesterday?" | `lcm_synthesize_around` with `period: 'yesterday'` (timezone-aware) |
| "Have we ever discussed X?" | `lcm_grep --mode hybrid` (FTS + Voyage rerank, +52.5pp recall on paraphrases) |
| "What did Eva *exactly* say about Y?" | `lcm_grep --mode verbatim` (full message rows, no summary paraphrase) |
| "What's the history with the operator-VM customer?" | `lcm_get_entity('operator-VM customer')` |
| "Where did this synthesis claim come from?" | `lcm_describe(id, expandChildren: true)` then `lcm_expand_query` |

The operator gets `/lcm health`, `/lcm purge` (soft-suppression cascading through 10+ read paths), `/lcm reconcile-session-keys`, `/lcm eval` (recall + drift), `/lcm worker` lifecycle, and a real lossless raw bedrock.

---

## Table of contents

1. [TL;DR — what merges and why](#tldr--what-merges-and-why)
2. [The problem this solves](#the-problem-this-solves)
3. [Architecture (with diagrams)](#architecture)
4. [The 8 agent tools](#the-8-agent-tools)
5. [Why Voyage embeddings (Phase A spike)](#why-voyage-embeddings)
6. [Worker auto-ticks](#worker-auto-ticks)
7. [Operator commands](#operator-commands)
8. [Cost discipline](#cost-discipline)
9. [What was CUT and why](#what-was-cut-and-why)
10. [Test infrastructure](#test-infrastructure)
11. [Audit history (10 waves, ~140 bugs closed)](#audit-history)
12. [Verification](#verification)
13. [Migration safety](#migration-safety)
[...]

PR #628, opened 2026-05-07
feat(v4.2): stub-tier stratification — externalize old tool results (rebased on main, independent of #613)

100yenadmin:
> **Independent v4.2 PR rebased directly onto `main`.** Same feature as #626 but without the #613 dependency, so maintainers can review and test v4.2 in isolation before either PR lands. We'll eventually merge with #613 anyway, but this version lets you exercise v4.2 against the current v3.x baseline.

Companion: #626 (same v4.2 feature, stacked on top of #613).

## The problem this solves

When a long session pushes against the token budget at assemble time, v4.1's only lever for evictable items is "drop the whole row." Heavy tool results (12K+ tokens for a verbose `Read`/`Bash`/`Grep`) force the budget into a bad choice:

- Keep the heavy old tool result → lose 3 older summaries
- Drop the tool result → lose its assistant pairing and meaning too

Measured on a real DB (live snapshot, 2.6 GB, 315k messages), session `0cb8928b` at 258k budget: chronological eviction kept **333 items**.

## What this PR does

Adds a per-row sidecar (`messages.large_content`) that stores a `file_xxx` id pointing to the externalized payload in `large_files` (existing v4.1 storage table). At assemble time, evictable tool-result rows with the sidecar populated are replaced with the v4.1 `[LCM Tool Output: file_xxx | tool=… | N bytes]` reference format that's been in production for months. Drilldown uses the existing `lcm_describe(id="file_xxx")` path.

```
[LCM Tool Output: file_85b322f5fda14187ab641331 | tool=exec | 34,975 bytes]

Exploration Summary:
Tool: exec | Command: bash -lc 'cd /Users/lume/.openclaw/workspace && echo "== selfheal ==" && sed -n "1,260p" scripts/evaos-support/selfheal.sh && echo "== config-guard ==" …

Use lcm_describe with the file id to inspect the full output.
```

The `Exploration Summary` line carries a one-line preview of the originating `tool_input` so an agent reading the conversation can match a user reference like "the selfheal.sh script you read earlier" to the right fileId, then drill down.

## Architecture

| Layer | File | What |
|---|---|---|
| Schema | `src/db/migration.ts` | `messages.large_content TEXT` (idempotent ALTER); `PRAGMA busy_timeout=30000` before `BEGIN EXCLUSIVE` to coexist with running gateway |
| Store | `src/store/conversation-store.ts` | Project + map row → `MessageRecord.largeContent` |
| Assembler | `src/assembler.ts` | New `applyStubSubstitution()` runs before budget pass on evictable items only |
| Engine | `src/engine.ts` | Forwards `config.stubLargeToolPayloads` (default `false`) |
| Tool description | `src/tools/lcm-describe-tool.ts` | Strengthened to mention `[LCM Tool Output: file_xxx]` references |
| Migration script | `scripts/lcm-blob-migrate.mjs` | Idempotent (only touches `large_content IS NULL` rows); 200-row chunked transactions; `PRAGMA busy_timeout`; `wal_checkpoint(TRUNCATE)` after large UPDATE; populates `exploration_summary` with `tool_input`-derived disambiguator |

## Empirical bench

Session `0cb8928b`, 6,804 messages, 258k token budget:

| Variant | Items | Tokens | Stubbed | Tokens saved |
|---|---:|---:|---:|---:|
| v4.1 baseline | 333 | 252,288 | 0 | 0 |
| **v4.2 (full migration)** | **689** | **257,849** | **86** | **412,373** |

Tool-result count is **identical** in both (101 each). v4.2 stubs heavy ones and reuses budget for older history. Same token budget → ~2× wall-clock context (~74 min → ~130 min).

## Drilldown validation (Opus 4.1 subagents)

| Test | Result |
|---|---|
| Conversational summary ("what did we work on?") | Substantive answer from assistant turns. Zero tool calls. **No confabulation.** |
| Specific probe of elided content | **Found correct fileId, wrote `lcm_describe(id="file_xxx")` call, refused to fabricate.** |

Opus on the disambiguator format:

> "The Exploration Summary `Tool: … | Command: …` line was extremely helpful. The command string contained `sed -n \"1,260p\" scripts/evaos-support/selfheal.sh` literally — that's an unambiguous keyword match. With it, the mapping was one grep away."

## What's NOT stubbed

- **Fresh tail** (last ~64 turns / 24K tokens) — agent's working memory preserved verbatim
- **Assistant turns** — never stubbed; narrative of what was done always intact
- **User turns** — never stubbed
- **Tool messages without `large_content`** — legacy / unmigrated rows untouched
- **Tool messages with degraded role** (DB role 'tool' but no `toolCallId`) — phantom drilldown risk avoided

## Default off

Behind `config.stubLargeToolPayloads` (default `false`). Flag off → byte-identical to v3.x main behavior.

## Mitigation evaluation

Four mitigations recommended by the Opus comparative analysis went through `first-principles-architectural-decision` skill (research / run-the-system / where-it-lives / adversarial debate at ≥95% confidence). Verdict: **REJECT ALL FOUR.** Decision record at `audit/v42-bench/DECISION-mitigations.md`.

| Mitigation | Decision | Why |
|---|---|---|
| Recency cue `[t-NNm]` | REJECT | Cache thrashing — clock-based string changes prefix every assemble. User timestamps already exist inline. |
| `<lcm-stub>` XML wrapping | REJECT | Existing `[LCM Tool Output:]` format works in live test. Novel format = unproven regression risk. |
| Empty-assistant collapsing | REJECT | "Empty" assistant turns contain `tool_use` blocks (Anthropic/OpenAI wire contract). Collapsing breaks pairing. |
| Resolution markers | REJECT | No reliable signal for "work completed". False positives strictly worse. |

## Tests

**868/868 pass on main** (added 5 new v4.2 tests in `test/v42-stub-tier.test.ts`):

- `emits stubs only for evictable externalized tool messages` (boundary)
- `preserves tool_use ↔ tool_result pairing when stubbing`
- `never stubs tool messages without externalized files (legacy rows)`
- `preserves multi-block tool_result content shape (image + text)`
- **`drilldown round-trip: agent can recover the full payload via the file_xxx referenced in the stub`**

## How to download and test

```bash
gh pr checkout <this-pr-number> --repo Martian-Engineering/lossless-claw
git log --oneline -1   # → feat(v4.2): stub-tier stratification …
npm ci

# Full test suite
VOYAGE_API_KEY=$(cat ~/.openclaw/credentials/voyage-api-key) \
LCM_TEST_VEC0_PATH=$HOME/.openclaw/extensions/node_modules/sqlite-vec-darwin-arm64/vec0.dylib \
  npx vitest run --reporter=dot
# expected: 868 / 868 pass

# Build deployable tarball
npm run build && npm pack

# Snapshot a DB before testing migration
mkdir -p /tmp/v42-test
cp ~/.openclaw/lcm.db /tmp/v42-test/lcm-test.db

# Schema migration runs automatically on plugin load; the bench triggers it once.
VOYAGE_API_KEY=... LCM_TEST_VEC0_PATH=... \
  npx tsx scripts/v42-assemble-bench.mjs --db /tmp/v42-test/lcm-test.db --variant baseline

# Dry-run blob migration (size the eligibility set)
node scripts/lcm-blob-migrate.mjs --db /tmp/v42-test/lcm-test.db --threshold-bytes 8000 --dry-run

# Apply for real
mkdir -p /tmp/v42-test/files
node scripts/lcm-blob-migrate.mjs --db /tmp/v42-test/lcm-test.db \
    --threshold-bytes 8000 --storage-dir /tmp/v42-test/files

# Bench v4.2 against the migrated DB
VOYAGE_API_KEY=... LCM_TEST_VEC0_PATH=... \
  npx tsx scripts/v42-assemble-bench.mjs \
    --db /tmp/v42-test/lcm-test.db --variant v42-stubs \
    --session-id <your-session-id>
# → reports estimatedTokens, contextItems, stubStats
```

To deploy live and watch real-runtime drilldown behavior, follow the same recipe as #626 (stop gateway → install tarball → migrate live DB → flip flag → restart → tail logs).

## Reversibility

| Step | How to reverse |
|---|---|
| Schema ALTER | Idempotent. Column can stay; readers ignore it. |
| Blob migration | `UPDATE messages SET large_content = NULL` |
| Storage files | `rm -rf <storage-dir>` |
| Flag enable | `stubLargeToolPayloads = false` + restart |

## Test plan

- [ ] **Live runtime drilldown** (post-merge): enable flag, run a tool-heavy session, observe drilldown invocation rate
- [ ] **JOIN cost on a large DB**: confirm `getLargeFile()` lookups in `resolveMessageItem` don't regress assembler latency
- [ ] **Reversibility drill**: enable → migrate → disable → confirm assembly returns to baseline

## LOC

| | Source LOC |
|---|---:|
| Schema | ~10 |
| Store | ~10 |
| Assembler | ~75 |
| Engine | 5 |
| Tools | ~10 |
| **Total source** | **~110** |
| Tests | ~380 |
| Migration script | ~190 |
| Harnesses | ~600 |

PR diff: 9 files, +2,081 LOC, 0 deletions to existing code.

### DIFF
--- a/.changeset/stub-tier-followups.md
+++ b/.changeset/stub-tier-followups.md
@@ -1,0 +1,5 @@
+---
+"lossless-claw": patch
+---
+
+Fix v4.2 stub-tier drilldown fallback for migrated tool payloads, align the migration script's default storage directory with runtime state-dir config, and repair `--revert --dry-run`.
--- a/architecture/v4.2-stub-tier-stratification.md
+++ b/architecture/v4.2-stub-tier-stratification.md
@@ -1,0 +1,438 @@
+# LCM v4.2 — Stub-Tier Stratification
+
+**Status:** Architecture proposal · pre-adversarial review
+**Base:** `pr-613-newest` (= PR #613 v4.1 omnibus)
+**Branch:** `feat/lcm-stub-tier-stratification`
+**Author:** drafted with operator
+**Last updated:** 2026-05-07
+
+---
+
+## 1. TL;DR
+
+**Problem:** v4.1's stored conversations are dominated (53%) by tool-result payloads. Most tool results are large (>5k tokens), repeat across turns, and the model rarely needs to look at the full byte-stream of a 4-day-old `find` result. But every assemble walks through them, inflating the per-turn prompt and burning context.
+
+**Proposal:** Separate the message *thread* (the semantic spine of the conversation) from the *payloads* (tool result bodies). The thread stays small. Payloads move to a content-addressable blob tier with stub references on the thread. Drill-down is by message-id via existing v4.1 tools (`lcm_describe expandMessages`, `lcm_grep --mode verbatim`).
+
+```
+BEFORE (v4.1):
+  messages.content holds:
+    - text messages (user, assistant) — naturally small
+    - tool result bodies — often 10k-50k tokens each
+  assemble() inlines all of these.
+
+AFTER (v4.2 proposal):
+  messages.content holds:
+    - text messages (user, assistant) — unchanged
+    - tool calls (the request) — unchanged, naturally small
+    - tool results: STUB ("[tool_result blob_ref=blob_abc123 size=26585t kind=vm_status]")
+      ← original body lives in lcm_blobs(blob_id, content, content_type, byte_size, token_count, ref_count)
+  assemble() inlines stubs for tool results outside the inline window.
+  Drill-down: lcm_describe(message_id, expandMessages=true) joins messages.blob_ref → lcm_blobs.content
+```
+
+**Expected outcome on operator's live DB:**
+- Per-turn assembled token cost: ~155k → ~75-90k (–50%)
+- DB row count unchanged; storage size unchanged (the bytes still live somewhere)
+- No information loss — every byte still queryable by id
+- No runtime API change — drill-down via existing v4.1 tool surface
+- One-shot migration tool to convert existing tool messages to stubs (operator runs once after deploy)
+
+---
+
+## 2. Goals & non-goals
+
+### 2.1 Goals
+
+1. **G1 — Reduce per-turn assembled context cost by ≥40% on a real LCM with mature history**, measured against the v4.1 agent-harness DB (3,855 leaves, ~13M token corpus), without losing information.
+2. **G2 — Lossless storage invariant preserved.** Every byte that v4.1 stores is still recoverable in v4.2; only its location and access pattern differ.
+3. **G3 — Backward-compatible schema migration.** Existing rows continue to work without conversion. Operator can run a one-shot migration tool to convert historical tool messages to stubs at their convenience, on a schedule, or never.
+4. **G4 — No new agent tools required.** Drill-down works via existing `lcm_describe(message_id, expandMessages=true)` and `lcm_grep --mode verbatim`. Both already exist in v4.1.
+5. **G5 — Cache-stable.** Stub representation does not change between consecutive turns for the same message — prompt cache invalidations only happen when a message itself changes.
+6. **G6 — Refcount-correct.** Blobs are deduplicated by content hash. Refcounts are updated atomically. Orphaned blobs (refcount=0) are GC'd by an explicit operator command, never automatically (lossless invariant).
+7. **G7 — Migration is operator-controlled and idempotent.** No automatic background mutation of existing data. The operator triggers conversion via CLI, can dry-run, can scope by date range or by token threshold.
+8. **G8 — Test parity with v4.1's agent-harness baseline.** `scripts/v41-qa-runner.mjs --suite full` must continue to return 30/30 against the migrated DB.
+
+### 2.2 Non-goals
+
+- **NG1 — Not changing the host (OpenClaw) at all.** This is a plugin-only change. Host's prompt assembly is unchanged.
+- **NG2 — Not changing summary semantics.** Summaries continue to compress raw messages into condensed leaves. The blob tier is orthogonal: it just changes where the bytes live.
+- **NG3 — Not deleting any data.** The blob tier is content-addressable storage with refcount; rows in `messages` and `summaries` continue to exist. The migration tool transforms `messages.content` from "full body" to "stub", but doesn't drop rows.
+- **NG4 — Not making the model smarter or change its tool-call behavior.** This is purely about how the assembler emits the conversation thread.
+- **NG5 — Not changing the public TypeScript API of `LcmContextEngine`.** New blob accessors are added; nothing existing is renamed or removed.
+
+---
+
+## 3. The problem in numbers (operator's live data)
+
+From the audit of conversation 1872:
+
+| Layer | Count | Tokens | % of stored |
+|---|---|---|---|
+| User messages | 1,890 | 630K | 5% |
+| Assistant messages | 12,444 | 5.65M | 42% |
+| **Tool result messages** | **12,212** | **7.04M** | **53%** |
+| Total messages | 26,546 | 13.3M | 100% |
+| Summaries (depth 0–3) | 624 | 655K | (separate) |
+
+**Distribution of tool message sizes:**
+| Size bucket | Count | Tokens |
+|---|---|---|
+| 0–1k | 9,500 | ~3M |
+| 1k–5k | 2,200 | ~5M |
+| 5k–10k | 250 | ~1.7M |
+| 10k–20k | 46 | ~550K |
+| 20k–50k | 16 | ~382K |
+
+**Top single-message offenders** (sample): VM status outputs at 26.5k each (29 copies of similar pattern), image-read base64 at 26k each, session-jsonl reads at 23k each, debug-trace dumps repeated 4x in 2 minutes (clear bug).
+
+**Stub-tier savings target:**
+- For tool messages outside the inline window (last ~30 turns), replace body with ~30-token stub.
+- Conservative: 12,000 of 12,212 tool messages stubbed × ~26 tokens average stub size ≈ 312K stub tokens (vs 7.04M body tokens) = **6.7M tokens saved on assemble** (when those 12k messages are in the assembled window).
+- Per-turn savings depends on which tool messages happen to be in the assembled window. From recent assemble events: ~155k assembled, of which ~120k is tool/assistant content. Stubbing tool results outside the inline window cuts that in half on average.
+
+---
+
+## 4. Design
+
+### 4.1 Schema additions (additive, no destructive changes)
+
+```sql
+-- New: blob tier (content-addressable storage)
+CREATE TABLE IF NOT EXISTS lcm_blobs (
+  blob_id        TEXT PRIMARY KEY,           -- 'blob_' + sha256(content)[0..32]
+  content        TEXT NOT NULL,
+  content_type   TEXT NOT NULL,               -- e.g. 'tool_result/bash', 'tool_result/image', 'tool_result/file_read'
+  byte_size      INTEGER NOT NULL,
+  token_count    INTEGER NOT NULL,            -- precomputed via estimateTokens at insert
+  ref_count      INTEGER NOT NULL DEFAULT 1,  -- number of messages referencing this blob
+  created_at     TEXT NOT NULL,               -- ISO 8601
+  source_role    TEXT,                        -- 'tool' | 'assistant_thinking' | etc. (informational)
+  CHECK (ref_count >= 0)
+);
+
+CREATE INDEX IF NOT EXISTS lcm_blobs_content_type_idx ON lcm_blobs(content_type);
+CREATE INDEX IF NOT EXISTS lcm_blobs_created_at_idx ON lcm_blobs(created_at);
+CREATE INDEX IF NOT EXISTS lcm_blobs_size_idx ON lcm_blobs(token_count DESC);
+
+-- Modify: messages table gets two new nullable columns
+ALTER TABLE messages ADD COLUMN blob_ref TEXT;  -- when set, content is the stub representation
+ALTER TABLE messages ADD COLUMN inline_window_pin INTEGER NOT NULL DEFAULT 0;  -- 1 = always inline regardless of age (operator-pinnable)
+
+CREATE INDEX IF NOT EXISTS messages_blob_ref_idx ON messages(blob_ref) WHERE blob_ref IS NOT NULL;
+
+-- New: blob GC log (when operator runs purge-orphans, log what was removed for auditability)
+CREATE TABLE IF NOT EXISTS lcm_blob_gc_log (
+  log_id        INTEGER PRIMARY KEY AUTOINCREMENT,
+  blob_id       TEXT NOT NULL,
+  byte_size     INTEGER NOT NULL,
+  token_count   INTEGER NOT NULL,
+  reason        TEXT NOT NULL,             -- 'refcount_zero' | 'manual_force' | etc.
+  removed_at    TEXT NOT NULL
+);
+```
+
+**Migration order (idempotent):**
+1. CREATE TABLE IF NOT EXISTS lcm_blobs
+2. ALTER TABLE messages ADD COLUMN blob_ref (NULLable, default NULL)
+3. ALTER TABLE messages ADD COLUMN inline_window_pin (default 0)
+4. CREATE INDEXES IF NOT EXISTS
+
+These run on every plugin load through the existing `runLcmMigrations` flow; existing rows are unaffected (blob_ref stays NULL → assemble inlines content as before).
+
+### 4.2 Stub format (thread-side)
+
+When a message is converted to use the blob tier, its `content` becomes a **stable, deterministic stub**:
+
+```
+[tool_result blob_ref=blob_<sha256-prefix> size_tokens=<N> kind=<content_type> drilldown=lcm_describe(messageId=<id>,expandMessages=true)]
+```
+
+Concrete example (a 26,585-token VM status output):
+```
+[tool_result blob_ref=blob_a1b2c3d4e5f6789012345678 size_tokens=26585 kind=tool_result/bash drilldown=lcm_describe(messageId=531655,expandMessages=true)]
+```
+
+**Stub size:** ~150–250 chars (~30–60 tokens). Constant across turns regardless of underlying body size.
+
+**Stub stability:** the stub is fully derivable from `(blob_id, token_count, content_type, message_id)`. None of these change between consecutive assembles → cache-stable.
+
+**Why include the drill-down hint inline?** So the model sees the exact tool call needed to recover the body without extra inference. Mirrors the v4.1 design philosophy of putting actionable hints in tool results (cf. `catalogStatus`, `confidenceBand` from PR #613's bug fixes).
+
+### 4.3 Inline window (which tool results stay inline vs stubbed)
+
+The assembler decides per-message whether to inline the full body or emit the stub. Three rules, in priority order:
+
+1. **`inline_window_pin = 1`** → always inline (operator can pin specific messages, e.g., critical instructions).
+2. **Within fresh tail** (the last `freshTailCount` messages or `freshTailMaxTokens`, same as today) → always inline. The model needs immediate-history fidelity.
+3. **Outside fresh tail** → if `blob_ref IS NOT NULL`, emit stub. Else inline (legacy uncoverted messages).
+
+Configuration knobs (added to `LcmConfig`):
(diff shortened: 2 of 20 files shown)


100yenadmin:
## Companion PR (stacked on #613)

This PR is the v4.2 stub-tier feature rebased onto **`main` directly**, independent of #613.

The original v4.2 work was developed on top of #613 (the v4.1 omnibus PR), where it stacks cleanly. That stacked version is at **#626**. Both PRs contain the same v4.2 logic — same architecture, same Opus-validated drilldown behavior, same decision record. The difference is only the base:

| PR | Base | Diff size | When to use |
|---|---|---|---|
| **#628 (this PR)** | `main` | ~2,080 LOC | Review v4.2 in isolation against v3.x main |
| **#626** | `main` (stacked on #613) | ~53K LOC (includes #613) | After #613 merges, this trivially rebases |

Pick whichever PR fits the review path you want — only one will need to merge. Once one lands, the other gets rebased/closed.

100yenadmin:
## Wave 1 + Wave 2 adversarial review — 3 P0 blockers found, returning to draft

7 parallel Opus 4.1 adversarial agents (5 in Wave 1 covering different functional areas, 2 in Wave 2 chasing consequences and deployment/security). Each had its own focus, no cross-pollination during the runs. Two independent agents found each P0 below.

### P0-A — Config flag is silently inert in production

**Files:** `src/db/config.ts:82-153,334-568`, `src/engine.ts:6722-6723`

`engine.ts` reads the flag via cast: \`(this.config as { stubLargeToolPayloads?: boolean }).stubLargeToolPayloads === true\`. But \`LcmConfig\` type does NOT declare the field, AND \`resolveLcmConfigWithDiagnostics\` does NOT copy it from \`pc\` (plugin config). An operator who sets \`\"stubLargeToolPayloads\": true\` in \`~/.openclaw/openclaw.json\` is silently ignored. The cast lies; the field is always undefined; the assembler never stubs.

Tests pass because every test (\`test/v42-stub-tier.test.ts\`, \`scripts/v42-assemble-bench.mjs\`, \`scripts/v42-drilldown-harness.mjs\`) calls \`assembler.assemble({ stubLargeToolPayloads: true, … })\` directly, bypassing both \`LcmConfig\` and \`resolveLcmConfigWithDiagnostics\`. The engine→assembler seam is uncovered.

### P0-B — \`lcm_describe(file_xxx)\` returns no file content; drilldown structurally broken

**Files:** \`src/retrieval.ts:199-218\`, \`src/tools/lcm-describe-tool.ts:225-251\`, \`src/large-files.ts:530\`

\`describeFile()\` returns only metadata: \`fileName\`, \`mimeType\`, \`byteSize\`, \`storageUri\` (a path string), \`createdAt\`, \`explorationSummary\`. No code in the entire repo reads from \`storageUri\` — verified via \`grep -rn 'readFile.*storageUri\|storage_uri' src/\`.

The PR's stub instructs the agent to \"Use lcm_describe with the file id to inspect the full output\" (\`large-files.ts:530\`), and the new \`lcm_describe\` description (Option D in this PR) explicitly says \"Call lcm_describe(id=file_xxx) to fetch the original output.\" This is a false promise.

Worse: combined with Option F (this PR's choice), \`lcm-blob-migrate\` writes the \`tool_input\` disambiguator into \`exploration_summary\`. That same string is ALREADY in the stub block. So a v4.2-migrated row that the agent drills down on returns LITERALLY THE SAME INFO already visible in the stub. **The drilldown loop is a structural no-op; the feature is net-negative for migrated DBs** (loses payload tokens for a round-trip that returns nothing).

**Why the test passed:** \`test/v42-stub-tier.test.ts:373-377\` asserts \"agent can recover the full payload\" by calling \`readFileSync(fileRow.storageUri)\` directly — bypassing \`lcm_describe\` entirely. The \"5/5 PASS\" Opus harness in the original PR description observed that the model decides to call \`lcm_describe\` — never that the call returns useful content. (The harness only inspects \`response.tool_calls\`; it never simulates the tool response back to the model.)

I owe an honest correction here: my earlier claim that Opus drilldown was empirically validated was based on roleplay, not real tool execution. The model perfectly identified the fileId and wrote the call; it never observed that the call returns nothing. The structural gap was hidden by the test design.

### P0-C — Path layout mismatch + secret leakage in \`exploration_summary\`

**Files:** \`scripts/lcm-blob-migrate.mjs:110,154-155\`, \`src/engine.ts:3830-3831\`

**Path layout:** Runtime large-file writer uses \`<largeFilesDir>/<conversationId>/<fileId>.<ext>\` (\`engine.ts:3830-3831\`). Migration writes flat: \`<storageDir>/<fileId>.txt\`. With default config, both point at \`~/.openclaw/lcm-files/\`. Any tool that walks one shape mishandles the other (cleanup, backup, integrity scan, doctor).

**Secret leakage:** \`renderToolInputDisambiguator()\` writes \`Tool: exec | Command: ${oneLine(inp.command, 240)}\` into \`large_files.exploration_summary\`. Pre-v4.2, commands like \`ssh -i ~/.openclaw/secrets/cloud-deploy-key host\`, \`AWS_SECRET_ACCESS_KEY=AKIA... aws s3 cp\`, \`curl -H \"Authorization: Bearer …\"\`, \`psql \"postgres://user:pw@…\"\` lived in heavy tool bodies and got evicted under budget pressure. **After v4.2, those 240-char excerpts are PERMANENT in every stub line, on every assemble.** A privacy regression that was not previously possible.

### Other notable findings (P1, P2, P3)

- **P1: Multi-block test is fake.** \`test/v42-stub-tier.test.ts:279-319\` claims to test \"preserves multi-block tool_result content shape\" but \`seedConversation\` only ever inserts a single text part. A regression that breaks array→string shape would still pass.
- **P1: File write inside transaction → orphan leak.** \`scripts/lcm-blob-migrate.mjs:159-167\` does \`writeFileSync\` inside \`BEGIN…COMMIT\`. ROLLBACK leaves orphan files; re-runs create more orphans. No cleanup mechanism.
- **P1: Failure summary mis-reports.** Same script: if chunk 25 of 32 fails, chunks 1-24 are committed (4800 messages migrated) but \`summary.applied\` stays 0. Operator re-runs, doubles orphan count.
- **P1: \`byte_size\` column mislabeled.** SQLite's \`length(content)\` returns CHARACTERS, not bytes. UTF-8 stub labels say \"34,975 bytes\" when the on-disk file may be 100KB. \`--threshold-bytes 8000\` is actually \"8000 chars.\"
- **P1: File ACLs are world-readable.** \`writeFileSync\` defaults to 0644. On multi-user / shared-CI boxes, every local user can read tool outputs that may contain credentials. Fix: \`writeFileSync(..., {mode: 0o600})\`.
- **P1: No FK between \`messages.large_content\` and \`large_files.file_id\`.** Conversation deletion cascades \`large_files\` but not the on-disk file; pre-migration backup restore leaves orphans; post-migration backup restore on a fresh disk leaves dangling pointers.
- **P1: Reversibility incomplete.** PR description claims \`UPDATE messages SET large_content = NULL\` reverts the migration. It doesn't unlink files or delete \`large_files\` rows. Re-running migration after revert creates new files on top of orphans → disk fills.
- **P1: Compaction \`file_ids\` tracking blind to \`large_content\`.** \`src/compaction.ts:1512\` extracts file_xxx IDs only from \`message.content\`, not \`large_content\`. Future GC pass that prunes \"unreferenced files\" will mistakenly delete v4.2-externalized files still referenced via \`large_content\`.
- **P1: FTS / lcm_grep view inconsistency.** FTS indexes \`messages.content\` (the original), but agent sees stub. Operator searches for a credential, gets a hit; agent calls \`lcm_describe\` to fetch full match → gets nothing. UX bug.
- **P1: Hot-cache flip on flag-on triggers prefix break.** First assembly after enabling the flag substitutes evictable items → different prefix hash → cache miss across every active session. Documented nowhere; will surprise operators on rollout.
- **P2: Live-DB migration race.** \`busy_timeout\` covers SQLITE_BUSY but a concurrent gateway UPDATE during migration could clobber \`large_content\` to NULL last-writer-wins. No advisory lock, no row-version check.
- **P2: \`lcm_describe\` description risks false-positive invocation.** New sentence \"ALSO USE THIS when you see [LCM Tool Output: file_xxx]\" applies in non-v4.2 DBs (where no such pattern exists). Eager models could hallucinate matches.
- **P2: \`stubToolCallId\` field set but never read.** Dead field in \`ResolvedItem\`.
- **P2: \`getLargeFile()\` lookup unconditional.** Runs on every assemble even when \`stubLargeToolPayloads=false\`. N point-lookups per assemble; no cache.
- **P3: Bench token-estimation \`length / 4\` heuristic** — not provider tokenizer. The 333→689 numbers are estimates.
- **P3: Drilldown harness OpenRouter-only**, no fallback for reviewers without credits.
- **P3: \`shortenToolCallId\` collision risk** — astronomically low but undetectable if it happens.
- **P3: Migration \`mime_type\` hardcoded \"text/plain\"** for all tool outputs (wrong for JSON, base64-image, binary).
- **P3: Decision record narrow scope.** \`audit/v42-bench/DECISION-mitigations.md\` debated prompt-shape mitigations only. Never enumerated disk-lifecycle, ACLs, secret-leak, or FK-less sidecar — exactly the issues this review surfaced.

### Synthesis

Even fixing only P0-A leaves the feature broken (P0-B means drilldown returns nothing). Even fixing P0-A and P0-B leaves the privacy regression (P0-C secret leakage) and operator-facing breakage (path layout, FK, reversibility, ACLs).

The architectural design (per-row sidecar + on-disk file + drilldown via existing \`lcm_describe\`) is sound; the integration is incomplete. **PR is returning to DRAFT.** Not a rebuild — a fix-pass.

### Fix path

| P0 | Fix | LOC |
|---|---|---|
| **A** | Add \`stubLargeToolPayloads: boolean\` to \`LcmConfig\` type + resolver entry; drop the cast in engine.ts | ~10 |
| **B** | Modify \`describeFile()\` to read \`storageUri\` and return content (size-bounded, with conversation-scope check + path validation under \`largeFilesDir\` to prevent traversal). Combined with Option β: re-run blob-migrate to generate real LLM-summarized \`exploration_summary\` so the stub line itself is informative. | ~80 (read path) + LLM-summary integration |
| **C** | (1) Migration writes to \`<storageDir>/<conversationId>/<fileId>.txt\` to match runtime layout. (2) Redact common credential patterns in \`renderToolInputDisambiguator\` OR truncate to argv[0] only. | ~30 |

Plus the P1s, especially: file ACLs (\`mode: 0o600\`), FK on \`messages.large_content\`, fix the \"applied:0 on chunk failure\" reporting bug, redo the multi-block test for real, write an actual end-to-end drilldown test that calls \`lcm_describe\` and asserts content (not bytes-on-disk).

### Posture

This is exactly what adversarial review is for. The earlier Opus drilldown validation that I cited as "5/5 PASS" was design-correct (Opus identified the right fileId) but did not execute the tool, so the structural gap was invisible to it. Wave 1 + Wave 2 caught it because the agents read the actual code, not the test results.

Both PRs (#628 and #626) returning to DRAFT until P0s + critical P1s addressed.

100yenadmin:
## Wave-3 verification complete — all P0/P1s closed, returning to ready-for-review

**Summary of waves**

| Wave | Agents | Findings |
|---|---|---|
| 1 | 5 parallel Opus 4.1, focused by file/concern | 2 P0s (config flag inert; lcm_describe doesn't read content), multiple P1s |
| 2 | 2 parallel Opus 4.1, P0 verification + deployment/security | Both P0s confirmed independently. New finding: secret leak in exploration_summary disambiguator (originally classified as P0-C, downgraded to P1 after path-layout finding revisited). |
| 3 | 2 parallel Opus 4.1, verification + regression hunt | Found P0 in initial fix attempt: redactCredentials regex bypassed by ~half the patterns it claimed to catch. Plus P1s: symlink path validation (path.resolve is lexical), Windows separator portability, manifest schema gap, --revert against snapshot unlinks live files, UTF-8 truncation mojibake. |

### Final state — commits on this branch (vs `main` `36c47d1`)

| Commit | What |
|---|---|
| `0035d38` | **Wave-3 fixes**: drop fragile redaction → fail-closed disambiguator; realpathSync path validation; openSync+fstatSync TOCTOU close; UTF-8 boundary scan; manifest schema entry; --revert validates storage_uri is under --storage-dir; deferred unlink-after-commit |
| `97679b8` | Wave-1+2 fixes: P0-A config schema; P0-B describeFile reads disk; secret redaction (later replaced); file ACLs 0600/0700; byte_size CAST AS BLOB; applied:0 reporting; --revert mode; multi-block test rewrite; drilldown round-trip via real tool path |
| Earlier 9 commits | Original feature implementation across Options C/D/F |

### Why the redaction strategy changed

Wave 3 Agent 9 ran 17 adversarial inputs through the original `redactCredentials` regex set. **8 patterns leaked** — including `--ssh-key=`, `Authorization: Basic`, JSON-quoted forms, and lowercase variants. The function was giving false confidence ("we catch all the credential patterns") while letting half of them through.

The replacement design is fail-closed: only LOW-RISK shapes (path, pattern, sessionId) make it into `exploration_summary`. Commands and URLs are elided — the agent gets `Tool: exec | (command elided for security; use lcm_describe with expandFile=true)`. This loses some disambiguator richness but eliminates the leak class entirely. For specific exec calls, the agent can match by tool name + byte size, then drill down via `lcm_describe(id=file_xxx, expandFile=true)` to see the full input + output.

### Final fix matrix

| Original finding | Severity | Status |
|---|---|---|
| `stubLargeToolPayloads` config flag inert | P0 | ✅ Fixed: `LcmConfig` field + resolver entry + manifest schema |
| `lcm_describe(file_xxx)` returns no content | P0 | ✅ Fixed: `describeFile` reads `storageUri` (size-bounded, path-validated via `realpathSync`, fstatSync to close TOCTOU, UTF-8 boundary-scan on truncation) |
| Secret leak in `exploration_summary` | P1 | ✅ Fixed: dropped fragile redaction in favor of fail-closed disambiguator (commands/URLs elided entirely) |
| File ACLs world-readable | P1 | ✅ Fixed: writeFileSync with `mode: 0o600`, mkdirSync with `mode: 0o700` |
| `byte_size` mislabeled (chars vs bytes) | P1 | ✅ Fixed: `length(CAST(content AS BLOB))` |
| `applied:0` mis-reporting on partial failure | P1 | ✅ Fixed: catch block sets `applied = done` before printing |
| Reversibility incomplete | P1 | ✅ Fixed: `--revert` mode with `--storage-dir` validation, deferred unlink-after-commit, --dry-run support |
| Multi-block test was fake | P1 | ✅ Fixed: actually seeds multi-block parts; asserts `content[0].type === "text"` shape |
| Drilldown test bypassed `lcm_describe` | P1 | ✅ Fixed: test now goes through `RetrievalEngine.describe(fileId, { expandFile: true })` |
| Path validation lexical (symlink bypass) | P1 | ✅ Fixed: `realpathSync` + `path.sep` separator |
| Manifest configSchema missing field | P1 | ✅ Fixed: declared in both labels and configSchema.properties |
| Compaction `file_ids` blind to `large_content` | P1 | ⏸ Deferred: separate bug from v4.2 (existing v4.1 GC story is "no GC"; nothing prunes orphans today). To fix when GC is built. |
| FTS view inconsistency | P1 | ⏸ Acknowledged in decision record. Working as designed (FTS indexes content, not stub). |
| Hot-cache flip on flag-on prefix break | P1 | ⏸ Documented in operator runbook (deploy guide); won't fix in code (intrinsic to enabling stubbing). |
| TOCTOU between fs ops | P2 | ✅ Fixed: single openSync + fstatSync + readFileSync(fd) |
| Sync FS in async hot path | P3 | ⏸ Deferred: bench shows <1ms per drilldown on SSD; revisit if observed under load |
| Token cap arbitrary | P3 | ⏸ Deferred: 32K default + 500K hard cap is reasonable; `expandFileMaxBytes` lets caller tune |

### Tests

**868/868 pass.**

5 v4.2 unit tests, including:
- Boundary: stubs only on evictable items with `large_content` set, fresh tail untouched
- Pairing preserved: tool_use ↔ tool_result toolCallId still matches after substitution
- Legacy rows untouched (no `large_content`)
- Multi-block content shape preserved (text + image parts → array stub)
- **Drilldown round-trip via `RetrievalEngine.describe(fileId, { expandFile: true })`** — content === originalPayload, contentTruncated === false

### Posture

PR is ready for review.

**3 review waves, 7 total Opus 4.1 agents, 18 distinct findings, all closed at P0/P1 level.** This is genuinely the cleanest version of v4.2.

Companion stacked PR is at #626 (will rebase trivially after either lands).

100yenadmin:
Currently in experimental testing. Results so far is it works but needs soak.

PR #637, opened 2026-05-08
fix: move @mariozechner/pi-* from optional peer-deps to runtime dependencies

### LINKED ISSUE #636: [regression] lossless-claw@0.9.4 fails to load on fresh openclaw@2026.5.7 install: ERR_MODULE_NOT_FOUND for optional @mariozechner/* peer-deps
## Summary

After upgrading to `lossless-claw@0.9.4` on `openclaw@2026.5.7` (clean install via `openclaw update` on macOS), the plugin fails to load at runtime with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@mariozechner/pi-coding-agent' imported from /Users/<user>/.openclaw/npm/node_modules/@martian-engineering/lossless-claw/dist/index.js
```

The three `@mariozechner/pi-*` packages (`pi-coding-agent`, `pi-ai`, `pi-agent-core`) are declared as **optional peer-dependencies** in `package.json`:

```json
"peerDependencies": {
  "@mariozechner/pi-agent-core": ">=0.66 <1",
  "@mariozechner/pi-ai": ">=0.66 <1",
  "@mariozechner/pi-coding-agent": ">=0.66 <1"
},
"peerDependenciesMeta": {
  "@mariozechner/pi-agent-core": { "optional": true },
  "@mariozechner/pi-ai": { "optional": true },
  "@mariozechner/pi-coding-agent": { "optional": true }
}
```

The `optional: true` flag implies the runtime should degrade gracefully when these are missing, but `dist/index.js` imports them at top level (build script: `--external:"@mariozechner/*"`), so absence becomes a hard `ERR_MODULE_NOT_FOUND` at module-load time, not a runtime feature gate.

## Environment

- **OpenClaw**: `2026.5.7 (eeef486)`, launchd user service on macOS / Node 22.22.1
- **lossless-claw**: `@martian-engineering/lossless-claw@0.9.4` (installed via `openclaw update`'s plugin-sync step)
- **Install path**: `~/.openclaw/npm/node_modules/@martian-engineering/lossless-claw/dist/index.js`
- **Pre-upgrade**: was running `lossless-claw@0.9.2` from `~/.openclaw/extensions/lossless-claw/` (older API, peer-deps weren't required at runtime there)

## Steps to reproduce

1. Fresh OpenClaw 2026.5.7 install where `~/.openclaw/npm/node_modules/@mariozechner/` does **not** contain `pi-coding-agent` / `pi-ai` / `pi-agent-core` (only `openclaw`'s bundled copies exist at `/opt/homebrew/lib/node_modules/openclaw/node_modules/@mariozechner/`).
2. `openclaw plugins doctor` → reports `lossless-claw failed to load: Cannot find package '@mariozechner/pi-coding-agent'`.
3. `plugins.slots.contextEngine: "lossless-claw"` becomes inert; sessions fall back to legacy compaction.

## Local workaround in use

```bash
mkdir -p ~/.openclaw/npm/node_modules/@martian-engineering/lossless-claw/node_modules/@mariozechner
for dep in pi-coding-agent pi-ai pi-agent-core; do
  ln -s "/opt/homebrew/lib/node_modules/openclaw/node_modules/@mariozechner/$dep" \
        "$HOME/.openclaw/npm/node_modules/@martian-engineering/lossless-claw/node_modules/@mariozechner/$dep"
done
openclaw gateway restart
```

After symlinking + restart: plugin loads cleanly (`Status: loaded`), context engine registers, transcript GC + cache-aware compaction fire as configured.

The OpenClaw plugin installer auto-symlinks the `openclaw` peer-dep at install time (visible in install output: `Linked peerDependency "openclaw" -> /opt/homebrew/lib/node_modules/openclaw`) but does **not** auto-symlink `@mariozechner/*` peer-deps, even though they're bundled inside openclaw at `node_modules/@mariozechner/`.

## Suggested fix paths

**(a) Move the three `@mariozechner/*` from `peerDependencies` to `dependencies`** — simplest: makes the plugin self-contained, npm installs them on `npm i`, no host-bundling assumption. Trade-off: ~minor disk overlap with openclaw's bundled copies, but version skew is bounded by `>=0.66 <1` and openclaw's bundled version (0.66.x at time of writing) sits in that range.

**(b) Honor the `optional: true` contract in runtime code** — keep the peer-dep declarations but switch the imports in `index.ts` to dynamic imports with try/catch, so the plugin still loads when these are absent (and just disables the features that need them). Aligns with the existing `peerDependenciesMeta.optional: true` intent.

**(c) Document that openclaw must auto-link `@mariozechner/*` for plugins that declare them as peer-deps**, and request the upstream fix in OpenClaw's plugin installer.

(a) is the smallest patch and lands easily; (b) is more semantically correct given the existing `optional: true` flag. Happy to open a PR for either if you'd prefer one specifically.

## Cross-refs

- `package.json` build script externalizes `@mariozechner/*`: `--external:"@mariozechner/*"` — so they must be resolvable at runtime from the plugin's module-resolution chain.
- Issue #588 is a different but adjacent regression (silent context-engine register drop on 2026.5.2 + 0.9.3). 0.9.4 + 2026.5.7 hits this load-time failure first; engine-register reaches doesn't even get a chance.

NePav:
## Summary

Moves `@mariozechner/pi-agent-core`, `@mariozechner/pi-ai`, and `@mariozechner/pi-coding-agent` from `peerDependencies` (with `peerDependenciesMeta.optional: true`) to runtime `dependencies`.

The plugin's bundled `dist/index.js` imports these unconditionally — the build script has `--external:"@mariozechner/*"`, so they must be resolvable from the plugin's module-resolution chain at runtime. The `optional: true` flag is therefore not honored: when these packages aren't present, the plugin fails to load with `ERR_MODULE_NOT_FOUND` instead of degrading gracefully.

Treating them as runtime `dependencies` makes the plugin self-contained on `npm install` and removes the host-symlink workaround that fresh OpenClaw installs currently need.

Fixes #636.

## Why this approach (vs. honoring `optional` via dynamic imports)

The simplest and safest fix is to make the `package.json` reflect what the runtime already does (hard-require these). Honoring `optional: true` in code would require splitting the imports into dynamic `await import()` calls with try/catch and feature-gating the dependent code paths — a larger change with semantic implications about which features should still work without the pi-* deps. If you'd prefer that direction instead, happy to close this PR and rework — let me know.

## Trade-offs

- **Pro:** Self-contained `npm install`. No host-bundling assumption. Works on any `openclaw` install path or version, not just ones that ship `@mariozechner/*` bundled.
- **Pro:** Removes the now-orphan `peerDependenciesMeta` entries (only `@mariozechner/*` were listed; `openclaw` peer-dep stays).
- **Con:** ~minor disk overlap with OpenClaw's bundled copies of the same packages on installs where both are present. Bounded by `>=0.66 <1` range and the existing pinned bundled version.

## Verification

Locally on macOS / Node 22.22.1, after applying:

```bash
$ rm -rf node_modules/@mariozechner/  # simulate fresh install with no host-symlink
$ npm install --omit=dev
added 685 packages in 12s
$ ls node_modules/@mariozechner/
pi-agent-core  pi-ai  pi-coding-agent
```

After running the plugin under OpenClaw 2026.5.7:
- `openclaw plugins inspect lossless-claw` → `Status: loaded` (was: `failed to load`)
- Context engine slot binds correctly; sessions use lossless-claw, not legacy fallback
- No `ERR_MODULE_NOT_FOUND` in logs

## Diff scope

- `package.json`: 3 entries moved (deps), 3 entries removed (peerDeps), `peerDependenciesMeta` removed
- `package-lock.json`: regenerated via `npm install --omit=dev` (smaller, since pi-* are no longer duplicated as devDeps)
- `.changeset/move-pi-deps-to-dependencies.md`: patch-level changeset describing the fix

## Notes

- I dropped `@mariozechner/pi-*` from `devDependencies` since they're now installed via `dependencies` regardless. If you'd rather keep them pinned in devDeps for reproducible test/build (the original had them at `0.66.1` exact), happy to add back — just say the word.
- Used `>=0.66 <1` for the dependency range to mirror the original peer-dep range; tightening to e.g. `^0.66.1` would also be reasonable.

### DIFF
--- a/.changeset/move-pi-deps-to-dependencies.md
+++ b/.changeset/move-pi-deps-to-dependencies.md
@@ -1,0 +1,7 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Move `@mariozechner/pi-agent-core`, `@mariozechner/pi-ai`, and `@mariozechner/pi-coding-agent` from `peerDependencies` (with `peerDependenciesMeta.optional: true`) to runtime `dependencies`. The plugin's bundled `dist/index.js` imports these unconditionally (build externalizes `@mariozechner/*`), so absence becomes a hard `ERR_MODULE_NOT_FOUND` at module load — `optional: true` was not honored at runtime. Treating them as runtime dependencies makes the plugin self-contained on `npm install` and removes the host-symlink workaround currently required on fresh OpenClaw installs.
+
+Fixes #636.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,42 +1,28 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.9.2",
+  "version": "0.9.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.9.2",
+      "version": "0.9.4",
       "license": "MIT",
       "dependencies": {
+        "@mariozechner/pi-agent-core": ">=0.66 <1",
+        "@mariozechner/pi-ai": ">=0.66 <1",
+        "@mariozechner/pi-coding-agent": ">=0.66 <1",
         "@sinclair/typebox": "0.34.48"
       },
       "devDependencies": {
         "@changesets/changelog-github": "^0.6.0",
         "@changesets/cli": "^2.30.0",
-        "@mariozechner/pi-agent-core": "0.66.1",
-        "@mariozechner/pi-ai": "0.66.1",
-        "@mariozechner/pi-coding-agent": "0.66.1",
         "esbuild": "^0.28.0",
         "typescript": "^5.7.0",
         "vitest": "^3.0.0"
       },
       "peerDependencies": {
-        "@mariozechner/pi-agent-core": "*",
-        "@mariozechner/pi-ai": "*",
-        "@mariozechner/pi-coding-agent": "*",
-        "openclaw": "*"
-      },
-      "peerDependenciesMeta": {
-        "@mariozechner/pi-agent-core": {
-          "optional": true
-        },
-        "@mariozechner/pi-ai": {
-          "optional": true
-        },
-        "@mariozechner/pi-coding-agent": {
-          "optional": true
-        }
+        "openclaw": ">=2026.2.17 <2026.6.0"
       }
     },
     "node_modules/@agentclientprotocol/sdk": {
@@ -2992,7 +2978,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3006,7 +2991,6 @@
       "version": "0.3.2",
       "resolved": "https://registry.npmjs.org/@mariozechner/clipboard-darwin-universal/-/clipboard-darwin-universal-0.3.2.tgz",
       "integrity": "sha512-mxSheKTW2U9LsBdXy0SdmdCAE5HqNS9QUmpNHLnfJ+SsbFKALjEZc5oRrVMXxGQSirDvYf5bjmRyT0QYYonnlg==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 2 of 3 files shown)


jalehman:
Thank you!

PR #659, opened 2026-05-11
fix: recover same-path transcript shrink epochs

jetd1:
## Summary

- Detect same-path transcript shrink when the stored bootstrap checkpoint points past the current JSONL EOF.
- Recover those shrink cases as bounded transcript epochs instead of accepting an empty append-only read as covered.
- Avoid stale content anchors for same-path shrink epochs and add bootstrap/afterTurn regressions for externalized and duplicate-frontier cases.

## Tests

- `npm test -- --run test/engine.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`
- `codex review --commit HEAD`

Co-Authored-By: Cha <cha@jetd.one> via OpenClaw

### DIFF
--- a/.changeset/same-path-transcript-shrink.md
+++ b/.changeset/same-path-transcript-shrink.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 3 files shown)


jalehman:
Thank you!

PR #662, opened 2026-05-12
feat: add compaction start token floor

### LINKED ISSUE #409: feat: budget-aware compaction deferral for unknown cache state

jetd1:
## Summary

- Adds `compactionStartTokens` config (env: `LCM_COMPACTION_START_TOKENS`) as an absolute current-prompt token floor before optional LCM compaction may start.
- Applies the floor across incremental leaf compaction, deferred after-turn debt, inline proactive threshold compaction, maintenance debt cleanup, and the public `compact()` hook while preserving critical-pressure bypass behavior.
- Updates config schema, UI hints, README, configuration docs, bundled skill reference, tests, and Changesets release metadata.

This addresses the same wasteful early-compaction problem described in #409, using an explicit token floor instead of the ratio-based `budgetHeadroomRatio` approach from the closed #410. The absolute floor matches the operator intent more directly: keep fully raw context until the prompt reaches N tokens.

Closes #409
Refs #410

## Tests

- `npx vitest run test/engine.test.ts test/config.test.ts`
- `npm run build`
- `npm test`
- `codex review --base main`

### DIFF
--- a/.changeset/compaction-start-tokens.md
+++ b/.changeset/compaction-start-tokens.md
@@ -1,0 +1,7 @@
+---
+"@martian-engineering/lossless-claw": minor
+---
+
+feat(engine): add `compactionStartTokens` to defer optional compaction until the prompt reaches an explicit token floor
+
+`compactionStartTokens` and `LCM_COMPACTION_START_TOKENS` let operators preserve fully raw context below an absolute current-prompt token count, while still allowing critical budget pressure to compact before overflow.
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Add a `lossless-claw` entry under `plugins.entries` in your OpenClaw config:
         "config": {
           "freshTailCount": 64,
           "leafChunkTokens": 80000,
+          "compactionStartTokens": 0,
           "newSessionRetainDepth": 2,
           "contextThreshold": 0.75,
           "incrementalMaxDepth": 1,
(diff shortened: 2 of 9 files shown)


jetd1:
Closing in favor of the upstream threshold full-sweep refactor (PR #681, shipped in v0.9.x). The new architecture replaced the multi-path inline compaction model with a single `CompactionEngine.evaluate()` gate: compaction now only triggers when `currentTokens > contextThreshold × tokenBudget`, which already acts as a relative floor that scales with budget size. This addresses the core problem from #409 (early compaction on large-context models) for the common case. The remaining gap — an absolute token floor that survives `maxAssemblyTokenBudget`-induced threshold lowering — is narrow enough to revisit if it becomes a real pain point.

PR #664, opened 2026-05-12
fix: skip adjacent duplicate ingests

hannesrudolph:
## Summary
- skip non-empty adjacent duplicate messages with the same role/content during ingest
- preserve legitimate repeated content separated by other turns and empty tool-result repeats
- add regression coverage and changeset

## Verification
- npm test -- test/engine.test.ts
- npm run build


hannesrudolph:
Closing this in favor of installing and validating current main first; recent unreleased transcript/replay fixes likely address the root path more directly.

PR #666, opened 2026-05-12
fix(tui): avoid transplant context ordinal collisions

holgergruenhagen:
## Summary

Fixes `lcm-tui transplant` context merge failures when the target conversation already has context items and SQLite enforces `UNIQUE (conversation_id, ordinal)`.

## Root Cause

`mergeTransplantedContextItems()` uses temporary ordinal moves before it rebuilds the final context order. Two conflict windows were possible:

1. Existing target items were shifted upward, but transplanted summaries were inserted at `tempShift + i + 1`, which can overlap the shifted rows.
2. After inserting summaries into a free range, the summary reorder update could still collide when an existing summary should move to ordinal `0` while a temporary transplanted summary was already at ordinal `0`.

Both show up as:

```text
UNIQUE constraint failed: context_items.conversation_id, context_items.ordinal
```

This was reproduced during real `lcm-tui backfill ... --transplant-to ...` runs against an existing target conversation with summary and message context items.

## Fix

- Insert transplanted summaries into the low ordinal range freed by the first shift.
- Move all summary context items out of the final low-ordinal range before assigning final summary ordinals.
- Keep the final ordering behavior unchanged:
  - summaries first, ordered by depth and creation time
  - messages after summaries

## Validation

```bash
cd tui
go test ./...
```

Result:

```text
ok  	github.com/Martian-Engineering/lossless-claw/tui	0.139s
```

### DIFF
--- a/tui/transplant.go
+++ b/tui/transplant.go
@@ -1136,7 +1136,7 @@ func mergeTransplantedContextItems(ctx context.Context, q sqlQueryer, targetConv
 		return fmt.Errorf("temporarily shift context ordinals for conversation %d: %w", targetConversationID, err)
 	}
 
-	// Step 2: Insert transplanted summaries with temporary high ordinals.
+	// Step 2: Insert transplanted summaries into the free low-ordinal range.
 	// They'll be reordered in step 3.
 	for i, source := range sourceContext {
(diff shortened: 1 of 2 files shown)


jalehman:
Thank you!

PR #677, opened 2026-05-14
Fix PI package scope after owner rename

abnershang:
## Summary
- update PI imports and runtime dependencies from the old `@mariozechner/*` scope to `@earendil-works/*`
- externalize the new PI scope in the build output
- refresh the lockfile and existing changeset for the runtime dependency fix

## Real behavior proof
- Reproduced local OpenClaw plugin load failure before patch: `Cannot find module '@mariozechner/pi-coding-agent'` from the published npm install.
- Linked this working tree into OpenClaw with `openclaw plugins install --link ~/path/to/lossless-claw`.
- After gateway restart, `openclaw health --json` reports `plugins.errors: []` and `lossless-claw` in the loaded plugin list.
- Gateway log shows `[lcm] Engine initialized` and `[lcm] Plugin loaded` from the linked tree.

## Verification
- `npm run release:verify`
- `git diff --check`


jalehman:
Thank you!

PR #685, opened 2026-05-15
fix: seed placeholder bootstrap_state when afterTurn stat-fail fallback runs

jetd1:
## Summary

`#649` added a permissive stat-fail fallback in `reconcileTranscriptTailForAfterTurn`'s slow path that returns `hasOverlap:true` to allow live `afterTurn` ingest even when the transcript can't be `stat()`-ed. It assumed the afterTurn-tail `refreshAfterTurnBootstrapState` hook would persist the checkpoint, but that hook delegates to `refreshBootstrapState`, which **itself calls `stat(sessionFile)` and throws on failure**. The hook's `try/catch` swallows the error, logs a warn, and leaves `conversation_bootstrap_state` `NULL`.

Every subsequent `afterTurn` then re-enters the slow path with `reason="checkpoint-missing"` (intentionally excluded from `allowNoAnchorImport`), the slow path returns `hasOverlap:false` + zero imports, and the outer `transcriptReconcileUnsafeToAdvance` guard skips the ingest batch. The conversation gets stuck in a **transparent-passthrough** state where `assemble()` takes the safeFallback path and returns `params.messages` verbatim, and **compaction never runs**.

The observable harm is **not** message loss (the host runtime still sees the full live context via assemble's fallback) but a permanent compaction outage that grows the prompt unbounded, multiplies token cost, and will eventually hit the model's hard context window.

### Live incident

A long-running session got stuck in this state for ~11 hours before auto-healing. The host runtime eventually rewrote the JSONL small enough to flip the slow-path reason to `same-path-shrink`, which IS in the `allowNoAnchorImport` whitelist.

Symptoms while stuck: `current_token_count` climbed to ~425K against a 300K `token_budget`, `cache_state=hot` because the live prefix was getting reused, but `last_observed_prompt_token_count` kept growing past the budget every turn (matching the "compaction never runs" failure mode above).

## Fix

Keep every `#649` design decision intact:
- ✅ epoch rollover import (`allowNoAnchorImport` + cap)
- ✅ `hasOverlap` return value
- ✅ `transcriptReconcileUnsafeToAdvance` guard in afterTurn
- ✅ "no anchor / no import → leave checkpoint stale on purpose" (the deliberately stale slow-path branch)
- ✅ permissive ingest in the stat-fail branch (the implicit policy that this hotfix does NOT relitigate — see follow-up section)

**Only** seed a placeholder `conversation_bootstrap_state` row in the stat-fail branch, directly via `summaryStore.upsertConversationBootstrapState` (no `stat()` call), anchored to the current `sessionFile` with `lastProcessedOffset=0` and a null entry hash.

This closes the implicit contract gap `"hasOverlap:true ⟹ a bootstrap_state row exists for next turn"` without re-introducing the unconditional refresh that `#649` deliberately removed. The placeholder makes a deliberately weak claim (`"I know nothing about this file yet; next time it's readable, start from the top"`), so `#649`'s no-proof-no-advance principle still holds — we claim zero progress, not partial progress.

### Recovery semantics on the next turn

- **stat() succeeds:** the append-only fast path reads from `offset=0` and re-walks the full transcript. `identity_hash` dedup prevents re-ingestion of anything already in the DB. `refreshBootstrapState` then overwrites the placeholder with real `{size, mtime, offset, hash}` for that turn's frontier.
- **stat() still fails:** the same stat-fail fallback runs again, idempotently upserting the placeholder. No progress, but no further damage.

## Test

`seeds placeholder bootstrap_state when afterTurn stat-fail fallback runs (#649 follow-up)` in `test/engine.test.ts`:

1. Prime a conversation row with a single ingested message (no `bootstrap_state` row yet — matches the production incident shape).
2. Call `afterTurn` against a `sessionFile` that doesn't yet exist on disk → `stat()` throws ENOENT → stat-fail branch runs.
3. Assert the placeholder `bootstrap_state` row was created with the expected fields and the warn log mentions the seed path.
4. Write the real transcript to disk and call `afterTurn` again.
5. Assert all transcript messages are now in the DB and the checkpoint has advanced past offset=0.

Verified that the test **fails** before the fix (with `AssertionError: expected null not to be null` on the `bootstrap_state` lookup) and **passes** after.

Full suite: 902 / 902 pass.

## Follow-up

A separate issue will discuss whether the stat-fail branch's *permissive-ingest* decision should be aligned with `#649`'s "no-proof-no-advance" principle (i.e. flip `hasOverlap` to `false` so the ingest batch is also skipped for the stat-fail turn). That's a behavior change worth discussing on its own — out of scope for this hotfix.


jetd1:
Follow-up policy discussion opened as #686.

jalehman:
Thank you!

PR #700, opened 2026-05-17
fix: normalize provider-prefixed runtime llm model refs

sercada:
## Summary
- Treat provider-prefixed Lossless runtime LLM model refs as canonical before applying a separate provider hint.
- Prevent configs like summaryProvider=openai plus summaryModel=openai/gpt-5.5 from being checked as openai/openai/gpt-5.5.
- Add regression coverage for the startup policy warning path.

## Tests
- npm run build
- npm test -- doctor-contract-api.test.ts plugin-config-registration.test.ts index-runtime-llm-complete.test.ts


jalehman:
Thank you!

PR #665, opened 2026-05-12
feat: Codex OAuth profile + interceptCompaction — accordion cadence for autonomous loops

100yenadmin:
## Why

Pairs with **[openclaw/openclaw#81164](https://github.com/openclaw/openclaw/pull/81164)** — an upstream openclaw PR that adds the `ContextEngine.interceptCompaction` contract and wires it into pi-coding-agent's `session_before_compact` SDK event. Together they implement an **accordion compaction cadence** for Codex autonomous loops: agent fills to ~90% → openclaw routes the compaction event to LCM → LCM losslessly compacts to ~35% → agent resumes with an LCM-derived summary instead of codex's lossy GPT one.

Current state (without these PRs): codex's native GPT compaction fires at 90%, summarizes the whole history into a single short string, replaces history. Repeated mid-turn for 100-1000-call autonomous loops, this compounds lossiness until almost no original context remains.

## Architecture

```
┌──────────────────────────────────────────────────────────────────────┐
│                  Codex autonomous-agent turn                           │
│                                                                       │
│  ┌────────────┐                  ┌────────────┐                       │
│  │ Tool call  │── ~50 calls ─►   │ Context    │                       │
│  │ 1..100     │                  │ at 90%     │                       │
│  └────────────┘                  └─────┬──────┘                       │
│                                        │                              │
│        pi-coding-agent emits ──────────┴──► session_before_compact    │
│                                                  │                    │
│  ┌─────────────────────────────────────────────────────────────────┐ │
│  │ openclaw#81164: compactionInterceptExtension                     │ │
│  │   ▼                                                              │ │
│  │ engine.interceptCompaction(request)  ── this PR's implementation │ │
│  │   ▼                                                              │ │
│  │ 1. Guards: ignored / stateless session → handled:false           │ │
│  │ 2. Guard:  no compactionTargetFraction configured → handled:false│ │
│  │ 3. Guard:  pre-abort signal → handled:false                      │ │
│  │ 4. engine.compact(targetTokens = 0.35 × budget, force=true)      │ │
│  │ 5. Guard:  mid-abort signal → handled:false                      │ │
│  │ 6. engine.assemble() → AgentMessage[]                            │ │
│  │ 7. Guard:  empty assembled → handled:false (lcm-produced-no-ctx) │ │
│  │ 8. serializeAssembledMessagesForCompaction() → summary string    │ │
│  │ 9. return { handled: true, summary, firstKeptEntryId, ... }      │ │
[...]

Issue #584, opened 2026-05-03
fix(summarize): withTimeout doesn't abort underlying call — stack underway HTTP requests on flap


100yenadmin:
## Summary

`withTimeout` in `src/summarize.ts:146-154` rejects with `SummarizerTimeoutError` after the deadline, but the wrapped `deps.complete` provider call continues running. On a flaky provider, retries stack up: caller catches the timeout, retries with a different model, and a second request fires while the first is still in flight.

## Background

Raised by the chunk-G concurrency agent during the lossless-claw#516 audit (`audit/pr516/chunk-G-concurrency.md`, P3 finding). Not a correctness bug today — wasted bandwidth and provider quota, not data corruption — but cancellation propagation is uniformly absent across the engine and worth fixing here as a contained example.

## Current behavior

`src/summarize.ts:146-154` (approximate):
```ts
async function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
  return Promise.race([
    promise,
    new Promise<never>((_, reject) =>
      setTimeout(() => reject(new SummarizerTimeoutError()), ms)
    ),
  ]);
}
```

The wrapped `promise` (the `deps.complete` call) has no `AbortSignal` and continues to consume connection + bandwidth + provider quota until the upstream returns. Caller treats the timeout as a model failure and retries, doubling the work in flight.

## Proposed fix

1. Plumb an `AbortController` through `withTimeout`:

```ts
async function withTimeout<T>(
  callFn: (signal: AbortSignal) => Promise<T>,
  ms: number,
  options?: { onTimeout?: () => void }
): Promise<T> {
  const controller = new AbortController();
  const timer = setTimeout(() => {
    controller.abort();
    options?.onTimeout?.();
  }, ms);
  try {
    return await callFn(controller.signal);
  } finally {
    clearTimeout(timer);
  }
}
```

2. Update `deps.complete` signature to accept an optional `AbortSignal` and pass through to the underlying provider client (most LLM SDKs accept `signal` natively).

3. Each summarize call uses the new shape:
```ts
const result = await withTimeout(
  (signal) => deps.complete({ messages, model, signal }),
  timeoutMs,
);
```

## Alternative considered

- **Promise.race + ignore the in-flight call** — current approach; what we have today. Wasteful.
- **Plumb a `cancel` callback instead of AbortSignal** — non-standard, every provider SDK speaks AbortSignal natively.

## Test plan

[...]

PR #712, opened 2026-05-19
fix: bound compactFullSweep so a compaction sweep cannot hang the turn

### LINKED ISSUE #711: Compaction stall: LCM's compactFullSweep and compactUntilUnder are unbounded — a compaction can hang the agent for ~20 minutes
*Filed early covering `compactFullSweep` only; expanded with the full investigation — the larger unbounded path (`compactUntilUnder`) was found later, plus the dispatch logic, diagrams, and the fix mapping.*

## Summary

On OpenClaw 2026.5.18 + lossless-claw 0.10.0, a context compaction can leave the agent unresponsive for **tens of minutes**. The compaction eventually completes (or fails), but the agent turn is hung the whole time.

LCM's `compact()` has **two** internal compaction routines, and **both** were unbounded:

1. **`compactFullSweep`** — the leaf-pass sweep (used for `threshold` / manual compaction). Its Phase-1 `while (true)` loop had no max-iteration cap and no wall-clock deadline.
2. **`compactUntilUnder`** — the budget path (used for `budget` / codex *automatic* compaction — the common path). It wraps a `for round = 1..maxRounds` loop **around** `compactFullSweep`; with `maxRounds = 10` and each round's sweep re-arming its own deadline, the worst case is ≈ 10 × 120 s ≈ **20 minutes**.

Both run synchronously on the turn-critical path (inside `assemble()` under deferred compaction), and each leaf pass also does a synchronous `node:sqlite` scan that blocks the Node event loop.

## How `compact()` dispatches

```mermaid
flowchart TD
    C["contextEngine.compact(force, compactionTarget)"]
    C --> D{"useSweep ?"}
    D -->|"manual OR compactionTarget = threshold"| FS["compactFullSweep"]
    D -->|"budget — codex automatic compaction"| CU["compactUntilUnder"]
    CU -->|"for round = 1..maxRounds (10), each round runs"| FS
    FS --> LP["Phase 1: while(true) leaf passes<br/>each pass = sync node:sqlite scan + summarizer LLM call"]
    classDef bad fill:#ffe3e3,stroke:#d00000
    class LP bad
```

`useSweep` is decided in `executeCompactionCore` (`src/engine.ts`): `useSweep = manualCompactionRequested || compactionTarget === "threshold"`. `manualCompactionRequested` is set only by LCM's own `/lcm` command — never by a host harness — so a host that asks for `compactionTarget: "budget"` always lands on `compactUntilUnder`.

## Path 1 — `compactFullSweep` is unbounded

`src/compaction.ts` — `compactFullSweep`'s Phase-1 leaf-pass loop (`while (true)`, introduced by #681 "threshold full sweeps") terminates correctly (no chunks left / tokens stop dropping) but is **uncapped** — no max-iteration limit, no wall-clock deadline. One leaf pass runs per raw-message chunk; a large conversation drives many passes (observed: **16 passes** on a 308 K-token conversation). Each pass also issues a synchronous `node:sqlite` `DatabaseSync` query that blocks the event loop.

## Path 2 — `compactUntilUnder` is unbounded (the larger gap)

`compactUntilUnder` (`src/compaction.ts`) loops `for (round = 1; round <= maxRounds; round++)` and **each round calls `compactFullSweep`**. `maxRounds` is 10 (`src/engine.ts`). Crucially, each `compactFullSweep` invocation re-initializes its **own** `sweepStartedAt` / `sweepDeadlineAt` — so even once a per-sweep deadline is added, that deadline **resets every round**:

```mermaid
flowchart TD
    CU["compactUntilUnder — for round = 1..maxRounds (10)"]
    CU --> R1["round 1: compactFullSweep — fresh 120 s deadline"]
    R1 --> R2["round 2: compactFullSweep — fresh 120 s deadline"]
    R2 --> RN["round 3..10: compactFullSweep — fresh 120 s deadline each"]
    RN --> TOT["a per-sweep deadline RE-ARMS every round<br/>total worst case ≈ 10 × 120 s ≈ 20 minutes"]
    classDef bad fill:#ffe3e3,stroke:#d00000
    class TOT bad
```

This is the path the **codex automatic compaction** (`force: false, compactionTarget: "budget"`) takes — the most common compaction trigger — so it is the dominant real-world stall.

## Turn-critical path — it runs inside `assemble()`

With `proactiveThresholdCompactionMode: "deferred"`, recorded compaction debt is drained inside the next turn's `assemble()` (`src/engine.ts`), which runs the sweep **before** returning the assembled prompt:

```mermaid
sequenceDiagram
    participant Host as OpenClaw host
    participant A as LCM assemble
    participant SW as compactFullSweep / compactUntilUnder
    participant P as Summarizer provider
    Host->>A: assemble() — build the next-turn prompt
    Note over A: deferred compaction debt is pending
    A->>SW: drain the debt inline (turn-critical path)
    loop each leaf pass / round
        SW->>P: summarizer LLM call
        P--xSW: slow / rate-limited
    end
    Note over SW: unbounded — up to ~20 minutes
    A-->>Host: assemble() finally returns
    Note over Host: the turn was blocked the entire time
```

## The trigger

Each leaf pass makes a summarizer LLM call, `withTimeout`-wrapped per call (configurable `summaryTimeoutMs`). When the summarizer model is slow or rate-limited, each pass burns its full timeout before falling back — and there is no cap on the *number* of passes (`compactFullSweep`) or *rounds* (`compactUntilUnder`).

In the real-world report, LCM's summarizer was pointed at `openai-codex/gpt-5.4-mini` while the Codex subscription was usage-capped (`FailoverError: You've reached your Codex subscription usage limit`). OpenClaw `openclaw/openclaw#82008` already warns against OpenAI-Responses summarization with Codex OAuth tokens — but a valid-yet-slow provider produces the same unbounded behaviour. The structural bug is the missing bounds, independent of the provider.

## Evidence

A real threshold compaction (gateway log):

```
[lcm] compact: decision ... compactionTarget=threshold force=false storedTokens=308095 ...
... 16 sequential "LCM compaction leaf pass (normal)" steps, 308095 -> ... -> 41875 tokens ...
[lcm] compact: done ... ok=true tokensBefore=308095 tokensAfter=41875 duration=119502
```

~2 minutes with a healthy summarizer. With a rate-limited summarizer each of those 16 passes instead burns up to `summaryTimeoutMs`; via `compactUntilUnder` the round loop multiplies that further.

## Environment

- lossless-claw 0.10.0 (post-#681 "threshold full sweeps"), on OpenClaw 2026.5.18.
- `proactiveThresholdCompactionMode: "deferred"`, `contextThreshold: 0.9`.

## Fix

PR **Martian-Engineering/lossless-claw#712** — bounds **both** paths:

- **`compactFullSweep`** — a hard per-pass iteration cap (`maxSweepIterations`, default 12) and a wall-clock deadline (`sweepDeadlineMs`, default 120 s), shared across Phase 1 and Phase 2. The sweep also yields the Node event loop between synchronous `node:sqlite` scans so a long sweep cannot freeze the gateway for its whole duration.
- **`compactUntilUnder`** — a new operation-wide wall-clock deadline (`compactUntilUnderDeadlineMs`, default 300 s) computed once at the start and threaded into every round's `compactFullSweep` (each round's sweep stops at whichever is sooner — its own deadline or the operation deadline), plus a per-round check before starting the next round. This caps the whole operation at ~5 minutes instead of ~20.

On hitting any limit, compaction stops cleanly and returns the consistent partial result, logging a clear warning; remaining context pressure is picked up by the next compaction. All limits are configurable via plugin config / env vars.

The host-side half — OpenClaw `await`ing a plugin `compact()` with no timeout of its own — is tracked at **openclaw/openclaw#84077** (fix: openclaw/openclaw#84083).

### Known follow-up

LCM does not read `abortSignal` anywhere (`engine.ts` / `compaction.ts` / `summarize.ts`). With #712's internal deadlines a timed-out compaction now self-terminates at LCM's own deadline, so this is no longer a hang risk — but making LCM honor `abortSignal` would allow *prompt* cancellation on a user interrupt instead of deadline-bounded termination. Worth a separate hardening change.

## Related

- #605 — *Runaway recursive compaction causes permanent context loss when summarize provider fails* — same trigger family.
- #584 — *`withTimeout` doesn't abort underlying call — stack underway HTTP requests on flap.*
- #639 — *Dedup Fail-Closed and Deferred Compaction Loop.*
- #604 — *`proactiveThresholdCompactionMode "inline"` never triggers threshold sweep.*
- openclaw/openclaw#84077 — the OpenClaw host has no safety timeout on the plugin-owned (`ownsCompaction`) compaction lane.
- PR #665 (`interceptCompaction`) and PR #619 (`threshold-hard-floor`) are related in-flight compaction work; #712 is complementary — it bounds the sweep routines themselves.

100yenadmin:
## Problem

On OpenClaw 2026.5.18 + lossless-claw 0.10.0, a context compaction can leave the
agent unresponsive for tens of minutes. Three compounding LCM-side causes:

1. **`compactFullSweep` was unbounded.** The Phase 1 leaf-pass loop in
   `src/compaction.ts` (`while (true)`) had no max-iteration cap and no
   wall-clock deadline — one leaf pass per raw-message chunk. Large
   conversations produce many passes (observed: 16 passes on a 308K-token
   conversation).
2. **The sweep ran synchronously on the turn-critical path.** With
   `proactiveThresholdCompactionMode: "deferred"`, recorded compaction debt is
   drained inside the next turn's `assemble()`, which runs `compactFullSweep`
   before returning the assembled prompt. Each leaf pass also does a synchronous
   `node:sqlite` scan that blocks the Node event loop.
3. **`compactUntilUnder` was unbounded across rounds.** Bounding one sweep is
   not enough: `compactUntilUnder` (`src/compaction.ts`) wraps a
   `for (round = 1; round <= maxRounds; round++)` loop around sweeps, calling
   `compact()` → `compactFullSweep` once per round with `maxRounds = 10`. Each
   `compactFullSweep` invocation re-initializes its own `sweepStartedAt` /
   `sweepDeadlineAt`, so the per-sweep `sweepDeadlineMs` **resets every round**.
   Worst case is `maxRounds × sweepDeadlineMs ≈ 10 × 120s ≈ 20 minutes`. This is
   the codex *automatic* compaction path (`force:false,
   compactionTarget:"budget"`) — the common one — so the per-sweep bound alone
   still leaves a real ~20-minute stall.

When the summarizer is slow or rate-limited, each pass burns its full
`summaryTimeoutMs` (default 60s, often configured 180s) before falling back;
16 passes × that = 15–48 minutes of a hung turn.

## Fix

- **Bound the sweep.** `compactFullSweep` now enforces both a hard per-pass
  iteration cap (`maxSweepIterations`, default 12) and a wall-clock deadline
  (`sweepDeadlineMs`, default 120000). The counter and deadline are shared
  across Phase 1 (leaf) and Phase 2 (condensed), so the *total* sweep stays
  bounded. On hitting either limit the sweep stops before starting another
  pass and returns the consistent partial result, logging a clear
  `compactFullSweep stopped at …` warning. Remaining context pressure is
  picked up by the next sweep.
- **Bound the whole `compactUntilUnder` operation.** `compactUntilUnder` now
  computes one operation-wide wall-clock deadline at its start and (a) threads
  it into every round's `compactFullSweep` via a new optional
  `operationDeadlineAt` input, so a sweep stops at whichever is sooner — its
  own `sweepDeadlineMs` or the operation deadline — and (b) checks it in the
  `for round` loop, bailing before the next round with the consistent partial
  result and a `compactUntilUnder stopped at …` warning. The total budget is a
  **separate knob**, `compactUntilUnderDeadlineMs` (default 300000), *not* a
  reuse of `sweepDeadlineMs`: a single legitimate round's sweep can use most of
  `sweepDeadlineMs`, so reusing it as the operation-wide budget would let the
  first round alone exhaust it and break multi-round compaction. 5 minutes
  leaves room for a few full-deadline sweeps while capping the worst case well
  below 20 minutes.
- **Yield the event loop.** The sweep now `await`s a macrotask
  (`setImmediate`) between the synchronous `node:sqlite` scans of consecutive
  passes, so a long sweep cannot freeze the gateway for its whole duration.
- **Time-box the inline drain.** The `assemble()` deferred-debt drain reaches
  `compactFullSweep` through `compact()`, so the deadline above also bounds
  what `assemble()` can do inline — the turn is no longer held hostage by a
  full sweep.

All three limits are configurable via plugin config (`maxSweepIterations`,
`sweepDeadlineMs`, `compactUntilUnderDeadlineMs`) or the
`LCM_MAX_SWEEP_ITERATIONS` / `LCM_SWEEP_DEADLINE_MS` /
`LCM_COMPACT_UNTIL_UNDER_DEADLINE_MS` environment variables, following the
existing config conventions.

This is complementary to the `interceptCompaction` handoff work (#665) and the
threshold hard floor (#619): it bounds the sweep itself, orthogonal to those.

The **host-side half** of this fix — giving the OpenClaw host a safety timeout
so it does not `await` a slow plugin-owned compaction forever — is in
**openclaw/openclaw#84083**. The two PRs together close the stall: this one
keeps an individual sweep bounded; that one bounds the host's wait on the
plugin regardless.

Related: #584 (`withTimeout` doesn't abort the underlying summarizer call). This
PR does not change `withTimeout` — bounding the sweep already caps total time
even when a single pass burns its full timeout. Aborting the underlying call is
a separate, larger change.

## Tests

- New `compactFullSweep bounds` suite in `test/lcm-integration.test.ts`: a sweep
  that would exceed the iteration cap stops cleanly with a consistent partial
  result; raising the cap genuinely runs more passes (probe-verified: 14 passes
  uncapped vs 2 capped — directly reproduces the 16-pass bug); a sweep that
  exceeds the wall-clock deadline stops cleanly; a bounded sweep returns within
  a small multiple of the deadline.
- New `compactUntilUnder bounds` suite in `test/lcm-integration.test.ts`: a
  multi-round `compactUntilUnder` that would otherwise run `maxRounds ×
  sweepDeadlineMs` stops at the operation deadline instead (total wall-clock
  bounded to a small multiple of `compactUntilUnderDeadlineMs`, not
  `maxRounds × sweepDeadlineMs`); it returns a consistent partial result on the
  deadline; and a generous deadline does not cut a legitimate fast multi-round
  run short.
- `test/config.test.ts`: default, plugin-config, and env-override coverage for
  all three new settings.
- `test/circuit-breaker.test.ts`: the cooldown test now fakes only `Date`
  (its cooldown is `Date.now()`-based) so the new in-sweep `setImmediate` yield
  still runs under that test.
- `npm run build` (the CI build gate) passes; the local vitest suite passes for
  all touched and adjacent files; `tsc --noEmit` introduces zero new errors
  versus the prior branch state.

Fixes #711

## Takeover hardening update

During takeover review, an additional edge case was found and fixed in `eb53bcf`: if the sweep deadline expired while selecting a leaf chunk or condensation candidate, `compactFullSweep` could still start one more summarizer pass after the budget was gone.

The branch now rechecks the sweep budget after selection and before starting either the leaf or condensed summarizer pass. Two regression tests cover deadline expiry during selection for both paths.

Validation run from `/Volumes/LEXAR/repos/lossless-claw-fix-sweep`:

- `./node_modules/.bin/vitest run test/lcm-integration.test.ts test/config.test.ts test/circuit-breaker.test.ts --maxWorkers=1` -> 160/160 passing
- `npm run build` -> passing
- GitHub checks on the updated head -> passing


100yenadmin:
@jalehman

PR #719, opened 2026-05-19
feat: v4.1 foundation — schema + concurrency primitives

100yenadmin:
## TL;DR

**First PR of the LCM v4.1 stacked series** — the bones the next 12 PRs build on. Adds the v4.1 database schema (16 tables) + worker concurrency primitives. All-additive: new files plus append-only, idempotent migration steps. **No existing code path changes; no user-visible behavior change yet.** This PR ships the skeleton; the next 12 ship the muscles.

## The v4.1 story in one paragraph

v4.1 rebuilds LCM the way a person actually remembers: **keep the raw conversation forever, embed it for similarity search, synthesize fresh views on demand.** The v3 `lcm_recent` tool produced summaries-of-summaries-of-summaries that got worse the further back you looked; v4.1 stops building rollups and adds an agent surface mapped to the five question types a person actually asks:

> "What did we work on yesterday?" · "Have we ever discussed X?" · "What did X *exactly* say?" · "Tell me about person/project Y" · "Where did this claim come from?"

This first PR is the SQL schema + worker coordination primitives that make the rest possible.

## Before / after — what this enables

```mermaid
flowchart LR
  subgraph Before["Before — v3 rollup era"]
    M1[messages] --> L1[leaf summaries]
    L1 --> D[day rollup]
    D --> W[week rollup]
    W --> Mo[month rollup<br/>summary OF summary<br/>OF summary]
    style Mo fill:#f99
  end
  subgraph After["After — this stack lands"]
    M2[messages<br/>+ suppressed_at] --> L2[leaf summaries<br/>+ suppressed_at]
    L2 -.->|sidecar| V[vec0 embeddings<br/>this PR creates table]
    L2 -.->|sidecar| E[lcm_entities<br/>this PR creates table]
    L2 -->|on demand only| S[synthesis_cache<br/>this PR creates table]
    style V fill:#9f9
    style E fill:#9f9
    style S fill:#9f9
  end
```

The green tables are what this PR creates; the next 12 PRs light them up.

### Schema (`src/db/migration.ts`)
Append-only `runMigrationStep(...)` blocks creating the v4.1 surface:

| Cluster | Tables |
|---|---|
| Embeddings substrate | `lcm_embedding_profile`, `lcm_embedding_meta`, `lcm_embedding_meta_cleanup_summary` (suppression-cascade trigger) |
| Entity catalog | `lcm_entities`, `lcm_entity_mentions`, `lcm_entity_type_registry`, `lcm_extraction_queue` |
[...]

PR #724, opened 2026-05-19
feat: v4.1 — lcm_grep upgrade (hybrid / semantic / verbatim modes)

100yenadmin:
## TL;DR

The headline retrieval feature of v4.1. Upgrades `lcm_grep` from a 2-mode keyword/regex search into a **5-mode Swiss-army recall surface** the agent can pick the right blade from. Adds `hybrid` (FTS + semantic + rerank), `semantic` (pure vector), and `verbatim` (full untruncated messages) modes alongside the existing `regex` and `full_text`.

**Headline number: +52.5pp paraphrastic recall lift** over FTS-only (measured Phase A spike against the user's actual 4,187-leaf corpus, 31-query stratified eval).

## The pain this solves — five questions, one tool

Look at the five questions humans actually ask of their memory:

| # | Question shape | Today (v3) | After this PR |
|---|---|---|---|
| A | "What did we work on yesterday?" | `lcm_recent` rollup (lossy) | (lands in #731) |
| **B** | **"Have we ever discussed X?"** | **FTS5 only — fails on paraphrase** | **`lcm_grep --mode hybrid`** |
| **C** | **"What did Eva *exactly* say about Y?"** | **No way — summaries paraphrase** | **`lcm_grep --mode verbatim`** |
| D | "Tell me about person/project Z" | No way — no entity catalog | (lands in #728) |
| E | "Where did this synthesis claim come from?" | Manual drilldown | (lands in #725) |

This PR closes the headline gaps **B** and **C** — the two retrieval shapes a person uses most often when their memory is foggy on the words but clear on the topic.

## The five modes (decision tree)

```mermaid
flowchart TB
  Q[Agent's question]
  Q --> A{Shape?}
  A -->|"I know the exact keyword"<br/>e.g. error code, ID, file name| FT[full_text<br/>FTS5 BM25 ranked<br/>FREE]
  A -->|"I have a regex pattern"<br/>e.g. version-number regex| RE[regex<br/>JS post-filter on FTS-loaded set<br/>FREE]
  A -->|"I'm fuzzy on words but clear on topic"<br/>'did we discuss X?'| HY[hybrid<br/>FTS + Voyage semantic + rerank<br/>~$0.001/query]
  A -->|"Exploration, cost matters"| SE[semantic<br/>pure Voyage embed + KNN<br/>~$0.0001/query]
  A -->|"I need EXACT WORDS<br/>not paraphrased summary"<br/>quote/citation| VE[verbatim<br/>full untruncated message rows<br/>FREE — direct DB read]
```

**The decision tree IS the docstring.** When an agent reaches `lcm_grep`, the description guides it to the right mode based on the question shape. No "try one then another" loop.

### Scenario B: "have we ever debugged that thing where workers were colliding?"

| Step | Today (FTS5 only) | After this PR (`mode=hybrid`) |
|---|---|---|
| 1 | `lcm_grep "workers colliding"` | `lcm_grep --mode hybrid "workers colliding"` |
| 2 | FTS5 matches 0 leaves (exact keyword absence) | FTS5 still 0 — but semantic catches the leaf about "two backfill workers stomping on the same row" |
| 3 | Agent reports "no results" | Voyage reranker scores the semantic hit highest |
| 4 | User tries 3 more variants | Agent quotes the leaf with full lineage |

### Scenario C: "what did Eva exactly say when rejecting `lcm_recent`?"

| Step | Today | After this PR (`mode=verbatim`) |
|---|---|---|
| 1 | `lcm_grep "lcm_recent"` | `lcm_grep --mode verbatim --role user "lcm_recent"` |
| 2 | Returns SUMMARY paraphrases that mention `lcm_recent` | Returns the actual `messages.content` rows where Eva typed `lcm_recent` |
| 3 | Agent paraphrases the paraphrase ("Eva expressed concerns about…") | Agent quotes Eva verbatim |
| 4 | Quote is wrong-but-plausible | Quote is exact |

Verbatim mode is the difference between "agent reconstruction" and "citation." For research, legal review, or any audit-bearing context, only verbatim quotes hold up.

### `src/tools/lcm-grep-tool.ts` — the 5-mode tool
- **`mode: 'regex'`** — JS regex post-filter on FTS-loaded set. Unchanged from v3.
- **`mode: 'full_text'`** — FTS5 BM25-ranked. Sanitization improvements: `sanitizeFts5Pattern()` rejects malformed FTS expressions before they hit SQLite (returns a structured error pointing the agent at `mode=regex` for true regex syntax).
- **`mode: 'hybrid'`** (NEW) — calls `runHybridSearch` from #723. FTS top-K ∪ semantic top-K → Voyage rerank → top N. Returns `cosineSimilarity` + `confidenceBand` + `ftsRank` so the agent can introspect why a hit ranked.
- **`mode: 'semantic'`** (NEW) — calls `runSemanticSearch` from #723. Pure vector KNN, no rerank. ~10× cheaper than hybrid. Returns the same hit shape.
- **`mode: 'verbatim'`** (NEW) — bypasses summaries entirely. Reads `messages.content` directly. Optional `role` filter (`user|assistant|tool|system`) for citation queries. Caps at 20 messages by default (these are full rows; protects context budget).

Plus the `summaryKinds: ['leaf'|'condensed']` filter (semantic/hybrid only) for kind-scoped recall — useful when the agent wants source leaves but not higher rollups.

### `src/engine.ts` (`getDb()` accessor — 11 lines)
The grep modes need direct DB access for FTS + JOIN + LIMIT queries that don't fit cleanly through the store helpers. Adds a small read-mostly accessor (`getDb(): DatabaseSync`) on `LcmContextEngine`. Tools should treat the returned handle as read-only; long-running writes still go through the store helpers.

### Tests
- `test/lcm-grep-tool-hybrid.test.ts` — hybrid-mode pipeline against a deterministic Voyage mock.
- `test/lcm-grep-verbatim-mode.test.ts` — verbatim returns full message rows, respects role filter, caps at limit.

## The science — the +52.5pp number

A Phase A spike on the user's actual corpus (`docs/v4.1/PR_DESCRIPTION.md` §"Why Voyage embeddings"), 31 stratified queries:

| Stratum | n | FTS-only | Hybrid (Voyage rerank-2.5) | Lift |
|---|---|---|---|---|
| FTS-easy (keyword matches) | 14 | 40.5% | **69.0%** | +28.5pp |
| Paraphrastic | 8 | 5.0% | **57.5%** | **+52.5pp** |

The decision gate was ≥30pp paraphrastic lift. The actual result beat that by 22pp. **Spike cost: $0.58** one-time. The decision was empirical, not a guess.

### Why rerank, not just embed?

The +52.5pp lift came primarily from **the reranker**, not from the embedding model. Embedding-only (`mode=semantic`) gives part of the lift; cross-encoder rerank on top gives the rest. That's why the agent's "primary recall mode" pointer goes to `hybrid` even though `semantic` is cheaper — the rerank cost is the differentiator on hard queries.

### Why all 5 modes in one tool?

Considered: 5 separate tools (`lcm_grep`, `lcm_semantic_recall`, `lcm_hybrid_recall`, `lcm_verbatim_recall`, `lcm_regex_recall`). **Reach-for analysis** (cold-agent simulation against 25 canonical scenarios) showed agents reach for the broadest tool whose description is clearest. They picked `lcm_grep` for 18 of 25 scenarios regardless of which "specialized" tool would be better. Lesson: agents follow gravity to the recognizable tool name; if you scatter modes across tools, the niche tools never get reached for.

Solution: ONE tool, FIVE modes, ONE description that routes the agent to the right mode by question shape. Wave-12 retired the standalone `lcm_semantic_recall` tool and folded it into `mode=semantic`. Net surface: simpler for the agent, same capability.

## Cost reality

| Workload | Cost |
|---|---|
| `mode=regex` / `full_text` / `verbatim` (DB only) | $0 |
| `mode=semantic` (Voyage embed only) | ~$0.0001/query |
| `mode=hybrid` (embed + rerank) | ~$0.001/query |

Annual projection for a heavy user (~10 hybrid queries/day): **~$3/year**. Not a budget concern.

## What this PR explicitly does NOT add

- The **`lcm_describe` expand-flags** (drilldown without sub-agent) — that's its sibling Feature-C piece, lands in **#725**.
- The **suppression-filter on `summary-store.ts`** — that's the soft-purge pair, lands with `/lcm purge` in **#729**.
- The **agent compaction `runWithTokenGate` wrapper** that wraps every tool's `execute` for context-pressure refusal — that lands LAST in **#732**. This PR ships `lcm_grep` without the wrapper; #732 re-adds it across all 6 tools at once.

## Stack position

Stacked on **#723** (semantic + hybrid services). Part of the 13-PR v4.1 series — see #719 for the full stack.

## Test plan

- [x] `npm run build` — esbuild bundle clean
- [x] `npx tsc --noEmit` — exact `main` baseline (631), zero net-new errors
- [x] CI green — `npm test` (incl. 2 new grep-mode suites) + smoke-import

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.10.0",
+  "version": "0.11.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.10.0",
+      "version": "0.11.1",
       "license": "MIT",
       "dependencies": {
         "@earendil-works/pi-agent-core": ">=0.74 <1",
@@ -20,6 +20,9 @@
         "esbuild": "^0.28.0",
         "typescript": "^5.7.0",
         "vitest": "^3.0.0"
+      },
+      "optionalDependencies": {
+        "sqlite-vec": "^0.1.7-alpha.2"
       },
       "peerDependencies": {
         "openclaw": ">=2026.5.12"
@@ -3116,7 +3119,7 @@
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
       "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
-      "devOptional": true,
+      "dev": true,
       "license": "Python-2.0"
     },
     "node_modules/array-union": {
@@ -3687,7 +3690,7 @@
       "version": "4.1.0",
       "resolved": "https://registry.npmjs.org/find-up/-/find-up-4.1.0.tgz",
       "integrity": "sha512-PpOwAdQ/YlXQ2vj8a3h8IipDuYRi3wceVQQGYWxNINccq40Anw7BlsEXCMbt1Zt+OLA6Fq9suIpIWD0OsnISlw==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "locate-path": "^5.0.0",
@@ -3961,7 +3964,7 @@
       "version": "0.7.2",
       "resolved": "https://registry.npmjs.org/iconv-lite/-/iconv-lite-0.7.2.tgz",
       "integrity": "sha512-im9DjEDQ55s9fL4EYzOAv0yMqmMBSZp6G0VvFyTMPKWxiSBHUj9NW/qqLmXUwXrrM7AvqSlTCfvqRb0cM8yYqw==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "safer-buffer": ">= 2.1.2 < 3.0.0"
@@ -4153,7 +4156,7 @@
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/locate-path/-/locate-path-5.0.0.tgz",
       "integrity": "sha512-t7hw9pI+WvuwNJXwk5zVHpyhIqzg2qTlklJOf0mVxGSbe3Fp2VieZcduNYjaLDoy6p9uGpQEGWG87WpMKlNq8g==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "p-locate": "^4.1.0"
@@ -4373,7 +4376,7 @@
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.3.0.tgz",
       "integrity": "sha512-//88mFWSJx8lxCzwdAABTJL2MyWB12+eIY7MDL2SqLmAkeKU9qxRvWuSyTjm3FUmpBEMuFfckAIqEaVGUDxb6w==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "p-try": "^2.0.0"
@@ -4389,7 +4392,7 @@
       "version": "4.1.0",
       "resolved": "https://registry.npmjs.org/p-locate/-/p-locate-4.1.0.tgz",
       "integrity": "sha512-R79ZZ/0wAxKGu3oYMlz8jy/kbhsNrS7SKZ7PxEHBgJ5+F2mtFW2fK2cOtBh1cHYkQsbzFV7I+EoRKe6Yt0oK7A==",
-      "devOptional": true,
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "p-limit": "^2.2.0"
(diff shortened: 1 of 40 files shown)


coolmanns:
Downstream operator signal: this is exactly the LCM retrieval upgrade I want to see land.

The big win for us is not “make LCM smarter” in the abstract. It is the recall ladder:

`lcm_grep` broad candidate generation -> reranked anchors -> `lcm_describe` for cheap exact inspection or `lcm_expand_query` for expensive DAG expansion.

Today, the expensive step can start from whatever FTS happened to surface. That works for exact names, IDs, file paths, and dates, but it is weak for the way humans actually ask memory questions: fuzzy topic, wrong wording, partial recollection. Hybrid search plus rerank before expansion is the clean place to fix that without changing LCM’s role as the source-of-truth/timeline substrate.

`verbatim` mode is also important. Summaries are fine for orientation, but when the user asks “what did X exactly say?” the answer needs message rows, not a paraphrase of a paraphrase.

One implementation nudge: if this starts Voyage-first, please keep the embedding/rerank provider boundary easy to swap. We already run local OpenAI-compatible embedding/reranker services successfully in adjacent memory search, and LCM would benefit from the same shape eventually: candidate retrieval stays cheap/local where possible, rerank improves anchor choice, and truth still comes from expanding/citing the underlying LCM nodes.

So: strong +1 from the OpenClaw operator side. This should reduce failed recalls, reduce unnecessary expansion cost, and make `lcm_expand_query` sharper without putting a model in charge of deciding what is true.

PR #737, opened 2026-05-21
fix(engine): demote ENOENT during afterTurn checkpoint refresh to debug

### LINKED ISSUE #609: afterTurn: bootstrap checkpoint refresh fails with ENOENT for sessions/<id>.jsonl on openclaw 2026.5.3
## Summary

After every agent turn, lossless-claw's `afterTurn` hook tries to stat session jsonl files at a path that openclaw doesn't currently write to:

```
[lcm] bootstrap checkpoint refresh failed for conversation=<n>: ENOENT: no such file or directory, stat '/data/.openclaw/agents/<agent>/sessions/<session-id>.jsonl'
```

(Path sanitized.)

Cosmetic — does not block compaction or summary, just emits one WARN per turn.

## Reproduction

OpenClaw `2026.5.3` with `lossless-claw@0.9.2`. Any agent conversation. The WARN appears in logs after every turn.

## Suggested fix

Either guard the stat with a fast-path existence check and silently skip if the file isn't written, or update the expected path to match openclaw 2026.5.x's actual session jsonl layout.

mvanhorn:
## Summary

Demote the `[lcm] afterTurn: bootstrap checkpoint refresh failed` warn to debug when the underlying failure is ENOENT/ENOTDIR (the session jsonl file is not where lossless-claw expects it). Other failure modes (DB writes, unexpected stat errors) continue to surface as warn.

## Why this matters

#609 reports the warn firing once per turn under OpenClaw `2026.5.3+`, `lossless-claw@0.9.2` through `0.9.4`. @holgergruenhagen's follow-ups confirm reproductions on `2026.5.7` and `2026.5.12`, including a six-warning sequence triggered by `/new` and an Active Memory variant. The issue's own description calls it cosmetic: compaction and summary handle the missing transcript on their own, so the only effect is log noise.

`isMissingFileError` already exists at `src/engine.ts:232` and is used by the bootstrap path one screen up to short-circuit identical "session file not where we expect" cases without warning. Reusing it inside `refreshAfterTurnBootstrapState`'s catch keeps the demotion logic consistent with the surrounding code: ENOENT/ENOTDIR falls to a debug line that names the missing path so operators can still trace it if they crank log verbosity up, and any other thrown error continues through the existing warn path.

No behaviour change beyond log level: `refreshBootstrapState` was already a fail-open call from the afterTurn perspective (the catch swallows the error either way), so demoting the missing-file branch does not change which followups run.

## Testing

- `npm test` reports 923/923 passing.
- `npm run build` succeeds (the change touches only the afterTurn catch block).

Closes #609

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -7095,6 +7095,12 @@ export class LcmContextEngine implements ContextEngine {
           sessionFile: params.sessionFile,
         });
       } catch (err) {
+        if (isMissingFileError(err)) {
+          this.deps.log.debug(
+            `[lcm] afterTurn: bootstrap checkpoint refresh skipped (session file missing) for ${sessionLabel} sessionFile=${params.sessionFile}`,
+          );
+          return;
+        }
         this.deps.log.warn(
           `[lcm] afterTurn: bootstrap checkpoint refresh failed for ${sessionLabel}: ${describeLogError(err)}`,
         );


100yenadmin:
@mvanhorn I linked this PR to issue #609 during cleanup. It looks like the direct ENOENT/log-noise fix candidate, so I labeled both sides linked-pr and am keeping #609 open until this merges and current CI verifies it.

mvanhorn:
Thanks for linking it to #609. Makes sense to keep that open until this merges and CI verifies. Let me know if you'd like any changes in the meantime.

mvanhorn:
@100yenadmin Following up on the link to #609: this PR has gone stale in a way that a rebase will not fix, so I am closing it.

The patch demoted one `log.warn` to `log.debug` inside the `refreshAfterTurnBootstrapState` closure in `src/engine.ts`. That closure no longer exists on current `main`. Searching the tree, `refreshBootstrapState` now appears only in `CHANGELOG.md` and `specs/transcript-reconciliation-by-entry-id.md`, with no hits under `src/`. That is why the rebase produced a ~110 line conflict against an empty region rather than an ordinary textual clash: the whole block was removed, not moved. `isMissingFileError` is still there in `src/value-utils.ts`, but the call site this was meant to guard is gone, and none of the three remaining `afterTurn` warns is the ENOENT one.

So there is nothing left for this diff to apply to, and keeping #609 open pending this merge will just hold it open forever.

On #609 itself: it is still labeled needs-repro, and I cannot tell from the outside whether the restructure removed the noisy warn as a side effect or whether the same ENOENT surfaces somewhere else now. Worth someone confirming against current main before deciding whether it is fixed or still live. Happy to open a fresh PR if it reproduces and you would like one.

Sorry for the stale link, and thanks for doing the cleanup pass that caught it.

PR #742, opened 2026-05-21
fix: populate systemPromptAddition from host memory supplements in context engine assemble()

### LINKED ISSUE #741: Context engine assemble() does not call buildMemorySystemPromptAddition — wiki digest missing from prompt
## Bug Description

When lossless-claw registers as OpenClaw's context engine via `api.registerContextEngine("lossless-claw", ...)`, the engine's `assemble()` method does not call `buildMemorySystemPromptAddition()`. This causes the wiki digest (e.g. `## Compiled Wiki Snapshot`) and any other registered memory supplements to be absent from the assembled prompt.

## Root Cause

OpenClaw's gate behavior delegates prompt assembly entirely to the registered context engine. The host's own `buildMemorySystemPromptAddition` is only called when no context engine is registered. Since lossless-claw's engine does not invoke this function, memory supplements (wiki digest, etc.) are silently dropped.

## Reproduction Steps

1. Enable lossless-claw as context engine (default behavior when plugin is installed on a compatible OpenClaw version)
2. Enable a wiki/memory supplement that registers via `api.registerMemorySupplement()`
3. Restart gateway
4. Inspect the assembled prompt — no `## Compiled Wiki Snapshot` section appears
5. Disable lossless-claw → wiki digest reappears in prompt

## Expected Behavior

The engine's `assemble()` should populate `result.systemPromptAddition` with the output of `buildMemorySystemPromptAddition()` (or equivalent) so that registered memory supplements are included in the assembled context.

## Current Workaround

Monkey-patching the engine object returned by `registerContextEngine` to wrap `assemble()`:

```javascript
const originalAssemble = engine.assemble.bind(engine);
engine.assemble = async function(params) {
  const result = await originalAssemble(params);
  if (!result.systemPromptAddition) {
    const addition = buildMemorySystemPromptAddition({
      availableTools: params.availableTools ?? new Set(),
      citationsMode: params.citationsMode
    });
    if (addition) result.systemPromptAddition = addition;
  }
  return result;
};
```

## Proposed Fix

In the `wirePluginHandlers` function where `api.registerContextEngine("lossless-claw", ...)` is called, the engine factory should ensure `assemble()` populates `systemPromptAddition` from the host's memory supplement registry. This could be done by:

1. Importing `buildMemorySystemPromptAddition` from the plugin SDK core
2. Calling it after the engine's own assembly and setting `result.systemPromptAddition` if not already set

## Environment

- lossless-claw: v0.11.1
- OpenClaw: 2026.5.x
- Wiki plugin enabled with compiled wiki supplements registered

## Impact

Any memory supplement registered via `api.registerMemorySupplement()` is silently dropped when lossless-claw is the active context engine. This affects wiki digests, compiled knowledge snapshots, and any other plugin that relies on the memory supplement injection path.

Yiaos:
## Summary

Fixes #741

When lossless-claw registers as OpenClaw's context engine, the host delegates prompt assembly entirely to the engine. The engine's `assemble()` method did not call `buildMemorySystemPromptAddition()`, causing all registered memory supplements (wiki digest, compiled knowledge snapshots, etc.) to be silently dropped from the assembled prompt.

## Changes

Modified `src/plugin/index.ts` — the `registerContextEngine` factory now wraps the engine to call `buildMemorySystemPromptAddition()` after the original `assemble()` returns, populating `result.systemPromptAddition` when it is not already set by the engine itself.

### Behavior:
- If `assemble()` already sets `systemPromptAddition` (future-proof), the wrapper is a no-op
- `buildMemorySystemPromptAddition` is lazily imported from the `openclaw` peer dependency (`openclaw/plugin-sdk/core` with fallback to `openclaw/plugin-sdk`)
- If the import fails (older OpenClaw versions without this export), the wrapper gracefully degrades — no error, no supplement injection
- Engine wrapping is idempotent (guarded by `_lcmMemorySupplementPatched` flag)

## Reproduction

1. Install lossless-claw on OpenClaw >= 2026.5.12
2. Enable a wiki/memory supplement via `api.registerMemorySupplement()`
3. Observe that `## Compiled Wiki Snapshot` (or equivalent) is missing from the assembled prompt
4. With this fix, the wiki digest appears correctly in the prompt

## Side Effects

- **Token cost:** ~400 tokens/turn for the wiki digest system prompt addition (< 0.3% of 200k context window)
- **No extra API calls:** `buildMemorySystemPromptAddition` only reads already-registered in-memory supplements
- **No cache invalidation:** The supplement content is deterministic per-turn based on registered supplements

## Notes for Maintainers

The dynamic import path (`openclaw/plugin-sdk/core`) may need adjustment depending on OpenClaw's actual package exports configuration. The fallback to `openclaw/plugin-sdk` is included for compatibility. If there's a preferred way to access `buildMemorySystemPromptAddition` from within a plugin (e.g., via the `api` object), I'm happy to refactor accordingly.

### DIFF
--- a/.changeset/context-engine-memory-supplements.md
+++ b/.changeset/context-engine-memory-supplements.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Require OpenClaw 2026.5.28 so context-engine assembly can include host memory supplements.
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ The bundled skill focuses on configuration, diagnostics, architecture, and recal
 - Node.js 22+
 - An LLM provider configured in OpenClaw (used for summarization)
 
-> **Compatibility:** `lossless-claw@0.10.0` and newer require OpenClaw `2026.5.12` or newer. These releases call OpenClaw's `api.runtime.llm.complete` capability for summarization, which is unavailable in older OpenClaw builds. If you cannot upgrade OpenClaw yet, stay on `lossless-claw@0.9.4` and remove `0.10.x`-only config such as `sweepMaxDepth`.
+> **Compatibility:** `lossless-claw@0.10.0` and newer require OpenClaw `2026.5.12` or newer for OpenClaw's `api.runtime.llm.complete` summarization capability. Releases that include context-engine memory supplement support require OpenClaw `2026.5.28` or newer for `buildMemorySystemPromptAddition`. If you cannot upgrade OpenClaw yet, stay on `lossless-claw@0.9.4` and remove `0.10.x`-only config such as `sweepMaxDepth`.
 
 ### Install the plugin
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@martian-engineering/lossless-claw",
-  "version": "0.11.3",
+  "version": "0.12.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@martian-engineering/lossless-claw",
-      "version": "0.11.3",
+      "version": "0.12.0",
       "license": "MIT",
       "dependencies": {
         "@earendil-works/pi-agent-core": ">=0.74 <1",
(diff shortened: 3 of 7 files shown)


100yenadmin:
@Yiaos I linked this PR to issue #741 during cleanup. It appears to be the direct systemPromptAddition/memory-supplement fix candidate, so I labeled both sides linked-pr and am keeping #741 open until merge and verification.

jalehman:
Thanks for the contribution, @Yiaos. This is landed now.

PR #756, opened 2026-05-27
fix(compaction): consult leaf trigger in evaluate() so under-threshold conversations still compact

baghvn:
## Problem

`compaction.evaluate()` (in `src/compaction.ts`) only returns `shouldCompact:true` when the live context tokens cross the budget threshold (`contextThreshold * tokenBudget`). Two related leaf-level mechanisms exist but are not consulted from `evaluate()`:

- `evaluateLeafTrigger()` measures raw message tokens outside the protected fresh tail against `leafChunkTokens`.
- The comment on `evaluateLeafTrigger` explicitly says: *"Automatic compaction no longer uses this as a trigger, but it remains useful for diagnostics and explicit maintenance commands."*

The result is that a conversation can stay forever under `contextThreshold * tokenBudget` and accumulate unbounded raw messages outside the fresh tail without ever being compacted. `afterTurn`'s threshold gate never fires, so neither `inline` nor `deferred` mode ever schedules anything for these conversations.

## Real-world evidence (one heavy install)

* 310 conversations, only 27 with any summaries (≈9%).
* 14 conversations with `pending=1` in `conversation_compaction_maintenance`, with reasons including `leaf-trigger` and `cold-cache-catchup` (written by older code paths). 9 of those had `last_started_at` NULL — never picked up.
* Several conversations sitting at 200K–269K raw tokens and `tokenBudget=500000` had been under the 75% threshold for the whole period, so `evaluate()` returned `shouldCompact:false` on every turn.
* `consumeDeferredCompactionDebt` would also clear those legacy-reason debts on the next maintain pass, because it re-checks `evaluate()` and clears when `shouldCompact === false`.

## Fix

Single, narrow behavioural change in `evaluate()`:

1. Keep the existing token-budget threshold gate behaviour as the primary trigger (still returns `reason: "threshold"`).
2. When the threshold gate does not fire, also call `evaluateLeafTrigger()`. If the leaf trigger is hot, return `shouldCompact: true` with `reason: "leaf-trigger"`.
3. `CompactionDecision` gains optional `rawTokensOutsideTail` and `leafTriggerThreshold` telemetry fields (always populated when leaf trigger was consulted) so callers can see what each gate observed.
4. `afterTurn` forwards the decision's actual reason to `recordDeferredCompactionDebt` and `scheduleDeferredCompactionDebtDrain`, so the maintenance row reflects which gate fired.
5. A `debug` log fires inside `evaluate()` when the leaf-trigger gate fires, so this is observable in the field without changing the existing log shape.

The existing `consumeDeferredCompactionDebt` legacy-clear path (which re-evaluates non-`"threshold"` debts and clears them when `shouldCompact === false`) now correctly *retains* `leaf-trigger` debt as long as the leaf trigger remains hot.

This also handles the staleness escape hatch implicitly: a conversation that has grown to many multiples of `leafChunkTokens` outside the fresh tail is by definition over the leaf trigger and is now picked up automatically. No new config knob is added.

## Test coverage

Added in `test/engine.test.ts` under `LcmContextEngine.compact evaluate leaf-trigger gate`:

1. `evaluate()` returns `shouldCompact:true` with `reason: "leaf-trigger"` when raw tokens exceed `leafChunkTokens` but stay under the context threshold.
2. `evaluate()` returns `shouldCompact:false` with `reason: "none"` when neither gate fires.
3. `afterTurn` enqueues deferred compaction debt with `reason: "leaf-trigger"` (in `conversation_compaction_maintenance`) and forwards the same reason to `scheduleDeferredCompactionDebtDrain`, when only the leaf gate fires.

`npm test` results on this branch: **48 files passed, 927 tests passed, 0 failed** (was 924 before this change).

## Risk / scope

* `CompactionDecision.reason` adds `"leaf-trigger"` to its union — additive type change. Consumers that switch on `reason` keep working; the only place that pattern-matches on it is `consumeDeferredCompactionDebt` (`reason === "threshold"`), and the new value flows through the non-threshold branch correctly.
* Existing `afterTurn` test "afterTurn ignores raw leaf pressure below the context threshold" still passes — it mocks `evaluate()` directly, and our behavioural change is inside `evaluate()`.
* No new config flags. No new dependencies. No DB migration.

## Related

This is a real fix for the long-tail "permanently-hot, sub-threshold conversation never compacts" failure mode that lossless-claw users hit on any conversation small enough to live under `contextThreshold * tokenBudget` but long-running enough to accumulate raw messages.

### DIFF
--- a/.changeset/leaf-trigger-evaluate-gate.md
+++ b/.changeset/leaf-trigger-evaluate-gate.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+`compaction.evaluate()` now consults the leaf trigger when the token-budget threshold has not been crossed. Conversations that stay under the context threshold but accumulate raw messages outside the protected fresh tail past `leafChunkTokens` will now correctly schedule deferred compaction with `reason: "leaf-trigger"` instead of being silently skipped on every turn. Fixes the case where conversations under the context threshold accumulated tens of thousands of raw tokens for weeks without ever being compacted.
--- a/src/compaction.ts
+++ b/src/compaction.ts
@@ -17,13 +17,22 @@ import { LcmProviderAuthError } from "./summarize.js";
 
 export interface CompactionDecision {
   shouldCompact: boolean;
-  reason: "threshold" | "manual" | "none";
+  reason: "threshold" | "leaf-trigger" | "manual" | "none";
   /** Persisted Lossless context tokens before runtime prompt overhead. */
   storedTokens: number;
   /** Runtime-observed prompt tokens, when supplied by the host. */
   observedTokens?: number;
   currentTokens: number;
   threshold: number;
+  /**
+   * Raw message tokens accumulated outside the protected fresh tail, when
+   * the leaf trigger was consulted. Populated when `reason === "leaf-trigger"`
+   * or when the threshold path also crossed the leaf trigger threshold; left
+   * unset otherwise so existing consumers see no behavioural change.
+   */
+  rawTokensOutsideTail?: number;
+  /** Leaf-trigger threshold (`leafChunkTokens`) used during evaluation. */
+  leafTriggerThreshold?: number;
 }
 
 export interface CompactionResult {
@@ -577,6 +586,10 @@ export class CompactionEngine {
     const threshold = Math.floor(this.config.contextThreshold * tokenBudget);
 
     if (currentTokens > threshold) {
+      // Token-budget threshold path: also surface leaf-trigger telemetry so
+      // callers can see when both gates would fire on the same turn. Reason
+      // stays `"threshold"` to preserve the existing debt-routing contract.
+      const leafTrigger = await this.evaluateLeafTrigger(conversationId);
       return {
         shouldCompact: true,
         reason: "threshold",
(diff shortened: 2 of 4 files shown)


jalehman:
Thanks for digging into this and for the detailed field data. I’m going to close this PR because it reintroduces leaf-triggered automatic compaction, which is behavior we intentionally removed in the compaction redesign.

The current policy is:

- automatic scheduling is driven by `contextThreshold` only
- once that threshold is crossed, Lossless runs a full sweep
- the sweep creates leaf summaries in `leafChunkTokens` chunks, then condenses summaries according to config
- `evaluateLeafTrigger()` is only diagnostic/manual support, not an automatic scheduling trigger

That distinction matters for prompt-cache behavior. The old incremental model compacted whenever raw history outside the fresh tail crossed `leafChunkTokens`, which could repeatedly rewrite the prompt prefix during an otherwise healthy conversation. The redesign intentionally lets the context grow until the threshold is crossed, then pays the prefix-break cost once with a full sweep.

This PR changes `evaluate()` so raw leaf pressure can make `afterTurn` enqueue automatic compaction debt below the context threshold. Even if the downstream full-sweep guard currently makes some of those cases no-op, the policy change is in the wrong direction: it makes leaf pressure an automatic trigger again.

The “forever under threshold” case described here is therefore expected under the new policy, not a bug by itself. A conversation can remain below the context threshold while accumulating more than one `leafChunkTokens` worth of raw history; that is the tradeoff we chose to preserve prompt-cache stability.

I do think the confusion is partly on us: `evaluateLeafTrigger()` still has legacy naming and returns `shouldCompact`, which makes it look like an active automatic policy. We should probably follow up by renaming/removing that API or changing it into a diagnostic raw-pressure inspection helper so it cannot be mistaken for an automatic trigger again.

PR #758, opened 2026-05-27
refactor: support sqlite

jalehman:


jalehman:
Closing this old SQLite-support draft as superseded by the current migration work.

The direction is still right, but this branch is based on much older main, now conflicts with current Lossless heavily, and predates the newer OpenClaw SQLite projection/session-target design. In particular, the current migration should be built around typed visible transcript entries, verified transcript epochs/frontiers, and the current compatibility metadata/docs/test surface rather than porting this runtime-message bootstrap approach.

Thanks for preserving the early exploration here.

PR #765, opened 2026-05-29
fix: surface prompt-matched recall after rotate

vincentkoc:
## Summary

- add a prompt-recall cue during context assembly for exact uppercase memory keys such as `CRABPOT_LCM_FACT`
- search the preserved raw message store when active summaries/fresh tail do not already contain the exact stored snippet
- add a regression that mirrors Crabpot's post-rotate memory recall failure shape
- add a patch changeset

## Why

Crabpot's forward LCM release gate currently classifies `lcm-basic-memory-turn` as the known `memory-recall-mismatch` failure: after `/lossless rotate`, a follow-up turn on the same stable session key cannot reliably observe the seeded memory fact through summarized LCM context.

Crabpot evidence: https://github.com/openclaw/crabpot/blob/6cf82936ceb463c54569ccabb68e598469841a40/reports/crabpot-behavior-evals.md#L39-L48

## Verification

- `npm run build`
- `npm test -- engine.test.ts -t "adds raw prompt-recall"`
- `npm test -- engine.test.ts`
- `npm test -- --maxWorkers=1`

Note: default parallel `npm test` hit the existing timestamp-sensitive bootstrap replay test (`keeps legitimate repeated assistant text when the ordering does not match a prior replay`) on this machine. The same test passes isolated, `engine.test.ts` passes, and the full suite passes with `--maxWorkers=1`.

### DIFF
--- a/.changeset/lcm-prompt-recall-after-rotate.md
+++ b/.changeset/lcm-prompt-recall-after-rotate.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Improve post-rotate recall by surfacing prompt-matched raw memory snippets when active summaries omit exact recall keys.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1197,6 +1197,11 @@ const PROMPT_RECALL_IDENTIFIER_PATTERN = /\b[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+\b/g;
 const PROMPT_RECALL_MAX_IDENTIFIERS = 4;
 const PROMPT_RECALL_MAX_MESSAGES = 4;
 const PROMPT_RECALL_MAX_MESSAGE_CHARS = 1200;
+const PROMPT_RECALL_SEARCH_LIMIT = PROMPT_RECALL_MAX_MESSAGES * 2;
+const PROMPT_RECALL_SENSITIVE_IDENTIFIER_PATTERN =
+  /(?:^|[^A-Za-z0-9])(?:ACCESS_?KEY|API_?KEY|AUTH|CREDENTIALS?|DEPLOY_?KEY|KEY|PASS(?:WORD)?|PRIVATE_?KEY|SECRET|TOKEN)(?=$|[^A-Za-z0-9])/i;
+const PROMPT_RECALL_SENSITIVE_VALUE_PATTERN =
+  /(?:-----BEGIN [A-Z ]*PRIVATE KEY-----|\bAKIA[0-9A-Z]{16}\b|\b(?:ghp|gho|ghu|ghs|ghr)_[A-Za-z0-9_]{10,}\b|\bgithub_pat_[A-Za-z0-9_]{20,}\b|\bxox[baprs]-[A-Za-z0-9-]{10,}\b|\b(?:sk|rk|pk)-[A-Za-z0-9_-]{10,}\b)/i;
 
 /**
  * Normalize AgentMessage variants into the storage shape used by LCM.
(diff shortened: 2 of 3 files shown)


100yenadmin:
## Mergeability write-up for PR #765

I did a hardening pass because this PR touches `src/engine.ts`, specifically the canonical `assemble()` path. The final state is mergeable, but the important part is why the engine change is bounded and what bugs were removed before calling it clean.

### What this PR is fixing

After session rotation/compaction, the active assembled context can contain summaries plus a fresh tail while older raw messages remain durable in LCM storage. If a later prompt asks for an exact all-caps key such as `CRABPOT_LCM_FACT`, and the active summary/tail omitted that exact key, the model can fail even though the raw history still exists.

This PR adds an optional prompt-recall cue during `assemble()`:

- It extracts a small number of exact identifier-like keys from the current prompt.
- It only searches raw stored messages when the active assembled context does not already cover the key.
- It renders bounded, quoted historical snippets as inert context.
- It drops the cue whenever the cue would compete with current/live context.

```mermaid
flowchart TD
  A["assemble(params)"] --> B["Build normal DB context<br/>summaries + raw tail"]
  B --> C["Extract prompt identifiers<br/>max 4, non-secret only"]
  C --> D{"Identifier already covered by<br/>summary, tail, or volatile live input?"}
  D -- "yes" --> E["Skip recall for that identifier"]
  D -- "no" --> F["Search stored raw messages<br/>full_text, recency, bounded"]
  F --> G["Fetch candidate message"]
  G --> H{"Eligible role<br/>user/assistant?"}
  H -- "no" --> I["Skip candidate"]
  H -- "yes" --> J["Extract exact boundary snippet<br/>containing sentence/line"]
  J --> K{"Sensitive identifier/value<br/>inside snippet?"}
  K -- "yes" --> I
  K -- "no" --> L["Render quoted recall cue"]
  L --> M{"Cue fits token budget<br/>without displacing live context?"}
  M -- "no" --> N["Drop optional recall cue"]
  M -- "yes" --> O["Prepend cue and add projection fingerprint"]
```

### Bugs found and fixed during review

1. **Optional recall could compete with current context**

   The recall cue is helpful but must never have priority over live input or already assembled DB context. I hardened `assemble()` so the cue is only retained if it fits the token budget and, after volatile live-input preservation runs, did not cause an over-budget result or evict assembled context. If it would compete, the engine drops the cue and retries without it.

2. **Sensitive snippets could leak through a safe-looking requested key**

   A prompt might ask for `PROJECT_ID`, while the same stored sentence contains `api_key`. The first implementation could have surfaced that sentence. The fix filters both requested identifiers and extracted snippets with secret-shaped identifier/value patterns, including lowercase/punctuation forms like `api_key`. Snippets containing secret-like material are not rendered.

3. **Substring matches could recall the wrong fact**

   `CRABPOT_LCM_FACT` must not match `CRABPOT_LCM_FACT_BACKUP` or `CRABPOT_LCM_FACTv2`. Search uses full text for candidate lookup, then the engine validates an exact identifier boundary before rendering any snippet.

4. **Value-before-key memories could be clipped into useless snippets**

   A stored message like `Remember blue-lantern-42 as CRABPOT_LCM_FACT.` needs the value before the key. The snippet extractor now returns the containing sentence/line and clips around the key with preceding context preserved.

5. **Stale recall could override newer corrections**

   If an active summary, raw tail, or volatile live message already mentions the requested key, the engine does not search older raw history for that key. This prevents stale prompt-recall from reintroducing an older value over a newer correction.

6. **Same historical message could only contribute one identifier**

   The dedupe key now includes both `messageId` and `identifier`, so a single historical message can safely supply separate requested facts such as `ALPHA_FACT` and `BETA_FACT`.

7. **Recall lookup failures could turn an optional feature into an assemble failure**

   Prompt recall now runs inside a catch boundary. If lookup fails, `assemble()` logs a warning and returns the normal DB context instead of falling back or failing the turn.

8. **Projection semantics needed to distinguish cue content**

   The base context projection epoch can remain stable while the prompt-specific recall cue changes. The PR adds a prompt-recall projection fingerprint when a cue is retained, so downstream cache/reuse logic can distinguish different cue content without changing the ordinary summary epoch.

### Why the engine path is safe

```mermaid
flowchart LR
  A["Normal assembled context"] --> D["Final messages"]
  B["Volatile live input"] --> D
  C["Prompt recall cue<br/>optional"] --> D

  B -. "higher priority" .-> C
  A -. "higher priority" .-> C

  C --> E{"Any pressure on<br/>budget/live/context?"}
  E -- "yes" --> F["Drop cue"]
  E -- "no" --> G["Keep cue"]
```

The critical invariant is: **prompt recall is additive only when cheap and non-disruptive**. It cannot be the reason we lose volatile live messages, cannot evict protected assembled context, and cannot hard-fail assembly.

### Validation

Local validation ran from `/Volumes/LEXAR/repos/lossless-claw-pr765`:

- `npm run build` passed.
- `npm test -- engine.test.ts -t "recall"` passed.
- `npm test -- --run test/engine.test.ts` passed.
- `npm test -- --maxWorkers=1` passed: 48 files, 940 tests.
- `git diff --check` passed.
- `codex review --base 640a0a869c8b2b172257ed8edcf3b19ba325cc8b` reported no discrete regression.

GitHub validation:

- `test` passed.
- `smoke-latest-openclaw` passed.
- PR is non-draft, mergeable, `mergeStateStatus=CLEAN`.
- No unresolved review threads.
- Cross-repo maintainer edits are enabled.

### Mergeability assessment

This is now 10/10 safe mergeable because the feature is bounded, optional, covered by targeted regressions, validated by the full local suite and GitHub CI, and the risky engine behaviors have explicit tests:

- no budget overrun,
- no displacement of live/current context,
- no stale recall over active summary/tail/live corrections,
- no substring recall,
- no secret-shaped identifier/snippet recall,
- no assemble failure when recall lookup fails,
- projection fingerprint changes when retained recall content changes.

100yenadmin:
## Additional PR option: move prompt recall out of `assemble()`

This PR is now hardened enough to merge as a tactical patch, but there is a cleaner architecture if we want the "correct" long-term fix instead of adding prompt-specific recall into the canonical `assemble()` path.

### Why consider a different PR shape?

The current patch works by injecting an optional synthetic `<lossless_claw_prompt_recall>` user message during `assemble()` when the prompt asks for an exact key that active summaries/tail omitted.

That solves the immediate post-rotate recall failure, but it puts retrieval behavior in the hottest and most sensitive part of the engine:

- `assemble()` should ideally be deterministic canonical context assembly: active summaries + fresh tail + existing focus overlays.
- Prompt-specific retrieval makes the assembled prompt depend on the current prompt text, not only on stored context state.
- Because the cue is injected as a user message, it must defend against stale facts, prompt injection, secret leakage, token pressure, projection/cache semantics, and live-context displacement.
- The amount of hardening in this PR is a signal that the behavior is crossing too many ownership boundaries for a minimal assemble-path feature.

So I would treat the current PR as a safe tactical bridge, and propose the following as the architecture-correct follow-up or replacement PR.

```mermaid
flowchart TD
  A["Current PR path"] --> B["assemble()"]
  B --> C["Normal summaries + fresh tail"]
  B --> D["Prompt-specific raw recall search"]
  D --> E["Synthetic user cue"]
  E --> F["Final model prompt"]

  G["Cleaner PR path"] --> H["assemble()"]
  H --> I["Canonical summaries + fresh tail only"]
  G --> J["Explicit recall layer"]
  J --> K["Tool result or structured retrieval metadata"]
  I --> L["Final model prompt"]
  K --> L
```

### Option B: explicit recall layer instead of assemble-time prompt injection

Build a separate recall path for exact-key recovery:

1. Keep `assemble()` canonical.
   - It should not search raw history based on the current prompt.
   - It should not prepend synthetic user messages for recall.
   - It should keep returning the same summary/tail projection for the same LCM state.

2. Add an exact-key recall helper behind the existing tool/retrieval lane.
   - Reuse the same strict identifier extraction and secret filtering from this PR.
   - Search only within the current session family by default.
   - Return bounded evidence with source message IDs and ordinals.
   - Do not render as a user-authored message.

3. Expose the result through one of these surfaces:
   - Preferably as an `lcm_grep` / `lcm_expand_query` improvement, so the model explicitly asks for recall.
   - Or as a new lightweight `lcm_recall_keys` tool for exact all-caps identifiers.
   - Or, if the host supports it, as structured `contextEngine` retrieval metadata that the host can render as tool/evidence context rather than as user prompt text.

4. Update the agent/system guidance to use the explicit recall surface when a prompt asks for a named old fact and the active context lacks it.

```mermaid
sequenceDiagram
  participant User
  participant Host
  participant LCM as lossless-claw
  participant Model

  User->>Host: "What is CRABPOT_LCM_FACT?"
  Host->>LCM: assemble()
  LCM-->>Host: canonical summaries + fresh tail
  Host->>Model: normal prompt with LCM tools available
  Model->>LCM: lcm_recall_keys(["CRABPOT_LCM_FACT"])
  LCM-->>Model: bounded evidence with source ids
  Model-->>User: answer based on explicit evidence
```

### Even better root-cause fix: summary retention contract

The deeper failure is that the summary/tail projection can omit exact recall handles. Another complementary PR would harden summarization itself:

- Preserve identifier-like keys (`FOO_BAR`, `FOO_BAR_FACT`, etc.) in summaries when they appear near user-stated facts.
- Add a compact "exact keys retained" appendix to generated summaries.
- Include a regression where raw history contains `CRABPOT_LCM_FACT is blue-lantern-42`, compaction runs, and the active summary retains the exact key/value.

This reduces how often runtime recall is needed at all.

```mermaid
flowchart LR
  A["Raw message<br/>CRABPOT_LCM_FACT is blue-lantern-42"] --> B["Summarizer"]
  B --> C["Summary body"]
  B --> D["Exact-key appendix"]
  C --> E["assemble()"]
  D --> E
  E --> F["Model sees key in canonical context"]
```

### Proposed alternative PR scope

If we want to keep the current PR small and merge it, this can be a follow-up PR:

**Title:** Move exact-key recall out of assemble-time prompt injection

**Implementation:**

- Add a reusable exact-key recall service/helper:
  - identifier extraction,
  - sensitive identifier/value filtering,
  - exact-boundary snippet extraction,
  - bounded current-session-family search,
  - source IDs in results.
- Wire it into an explicit tool surface such as `lcm_recall_keys` or an `lcm_grep` exact-key mode.
- Remove or disable automatic `<lossless_claw_prompt_recall>` injection from `assemble()`.
- Add prompt/tool guidance so agents know to use explicit recall when a requested key is missing from active context.
- Add regression tests proving:
  - no synthetic recall user message is emitted by `assemble()`,
  - exact-key recall returns the missing raw fact,
  - secret-shaped identifiers/snippets are suppressed,
  - stale/current-context priority is handled by the caller/tool result semantics,
  - current session family scoping is preserved,
  - source message IDs are returned for auditability.

### Decision recommendation

I would view the two paths like this:

| Path | Pros | Cons |
| --- | --- | --- |
| Merge current PR | Fixes the immediate post-rotate exact-key recall failure now; fully hardened and CI-green | Keeps prompt-specific retrieval inside `assemble()` |
| Alternative PR | Cleaner ownership boundary; recall is explicit, auditable, and not silently mixed into canonical context | Requires tool/host/prompt behavior changes, so it is larger and slower |

My recommendation: merge the current PR only if we need the tactical fix now. Then open the alternative PR as the architecture cleanup, with the goal of retiring automatic assemble-time prompt recall once explicit exact-key recall is available.

100yenadmin:
@vincentkoc I did a focused P1 review. CI is green, but I found two issues worth fixing before merge.

[P1] The sensitive-value guard misses underscore-style provider secrets. `PROMPT_RECALL_SENSITIVE_VALUE_PATTERN` currently catches `sk-...` / `rk-...` / `pk-...` (`src/engine.ts:1201-1204`) but not common underscore forms such as `sk_live_...`. Because `extractPromptRecallSnippet()` only suppresses when this regex matches (`src/engine.ts:1327`) and `buildPromptRecallCue()` renders the snippet into prompt recall (`src/engine.ts:7427-7433`), a matched project identifier can pull a nearby secret-like value back into the prompt.

Repro sketch without real secrets: a stored user message like `PROJECT_ID is launch-alpha; sk_live_${"a".repeat(24)}` should be rejected by recall, but the current regex does not match it. Please broaden the test matrix and matcher for underscore provider-token forms without committing any real credentials.

[P2] Recall applies the search limit before eligibility filtering. `buildPromptRecallCue()` asks search for only `PROMPT_RECALL_SEARCH_LIMIT` rows (`src/engine.ts:7399-7405`), then filters by stored message, role, exact snippet, and sensitive content (`src/engine.ts:7411-7419`). Since `ConversationStore.searchMessages()` applies the limit inside the FTS/LIKE query first, eight newer ineligible hits can starve an older valid fact. A fix could overfetch/paginate until enough eligible recall candidates survive, or push the role/eligibility constraints into the search query.

Validation: static inspection plus a narrow Node regex probe from `/Volumes/LEXAR/Codex`; I did not run the local suite. GitHub `test` and `smoke-latest-openclaw` are green on the current head.

100yenadmin:
@vincentkoc I pushed the P1/P2 hardening fixes to this PR branch now that maintainer edits are enabled.

Resolved:
- P1: `PROMPT_RECALL_SENSITIVE_VALUE_PATTERN` now suppresses underscore-style provider tokens such as fake `sk_live_...` shapes before prompt-recall snippets can render.
- P2: prompt recall now overfetches a bounded candidate set before eligibility/sensitive filtering, so recent ineligible hits cannot starve an older valid recall fact.
- Follow-up subagent review found the starvation regression did not prove the bound strongly enough, so I tightened the test to assert `limit: 32`, `mode: full_text`, `query: STARVED_FACT`, and `sort: recency`.
- A changed-line typecheck complaint in the new fixture was cleaned up; full repo `tsc --noEmit` still has unrelated baseline failures, but the new prompt-recall lines no longer show the reported error.

Commits pushed:
- `a2a8c6b` `fix: harden prompt recall candidate filtering`
- `1b75668` `test: assert prompt recall overfetch bound`

Validation from `/Volumes/LEXAR/repos/lossless-claw-pr736-prompt-lock`:
- Red tests first confirmed the fake underscore-token leak and filtered-candidate starvation.
- `npm test -- test/engine.test.ts -t "underscore provider tokens|continues prompt-recall search"` passed.
- `npm test -- test/engine.test.ts -t "prompt-recall"` passed: 14 tests.
- `npm test -- test/engine.test.ts` passed: 251 tests.
- `npm run build` passed.
- `git diff --check` passed.
- Full local `npm test` passed: 48 files, 942 tests.
- `npm pack --dry-run` passed.
- GitHub CI is green on head `1b75668`: `test` and `smoke-latest-openclaw` both passed.

Current assessment: no remaining P0/P1 blockers found on #765 after the hardening pass. I did not find a same-repo issue to link this PR to; the PR body links the external Crabpot eval evidence.

100yenadmin:
@vincentkoc refreshed this on top of the current merged P0/P1 stack and revalidated the current head `1fc5916`.

Validation from `/Volumes/LEXAR/repos/lossless-claw-pr765`:
- `npm test -- test/engine.test.ts -t "prompt-recall|underscore provider tokens|continues prompt-recall search|STARVED_FACT|recall"` passed: 17 targeted tests.
- `npm test -- test/engine.test.ts` passed: 269 tests.
- Full `npm test` passed: 49 files, 999 tests.
- `npm run build` passed.
- `git diff --check` passed.
- `npm pack --dry-run` passed.
- GitHub CI is green on `1fc5916`: `test` and `smoke-latest-openclaw`.

Current assessment: no remaining P0/P1 blockers found after the main refresh. This PR still has no same-repo issue to link, but the external Crabpot eval evidence in the PR body matches the prompt-recall-after-rotate symptom.

PR #766, opened 2026-05-29
feat: add explicit exact-key recall tool

100yenadmin:
## Summary

Alternative implementation for the architecture option proposed in this comment:
https://github.com/Martian-Engineering/lossless-claw/pull/765#issuecomment-4577449530

This adds an explicit `lcm_recall_keys` tool for exact all-caps identifier recall from stored raw messages. It avoids changing `assemble()` or injecting prompt-matched recall into every runtime turn.

## Why this PR exists

PR #765 is a tactical fix for prompt-matched recall after rotation, but it touches the canonical context assembly path. That is high blast radius because `assemble()` runs for normal runtime context, not just a crabbox/test-only path.

This PR is the compatibility-first alternative:

- no `engine.ts` changes
- no `assembler.ts` changes
- no database migrations
- no automatic prompt scanning during context assembly
- no mutation of assembled context
- explicit agent action when a user asks for a durable exact key

## Architecture

```mermaid
flowchart LR
  A["User asks for exact key\nCRABPOT_LCM_FACT"] --> B["Agent sees active context\nvalue missing"]
  B --> C["Call lcm_recall_keys"]
  C --> D["Session-family scoped raw-message search"]
  D --> E["Exact boundary validation"]
  E --> F["Secret-shaped key/snippet filter"]
  F --> G["Bounded evidence snippets\nmessageId + conversationId"]

  H["Normal assemble()"] -. "unchanged" .- B
  H -. "no prompt recall injection" .- C
```

## Behavior

`lcm_recall_keys` accepts up to four exact all-caps identifiers such as `CRABPOT_LCM_FACT` and returns bounded evidence snippets with source IDs. By default it searches the current session family, while still allowing explicit `conversationId` or `allConversations`.

The tool refuses broad fuzzy searches and secret-shaped identifiers such as `API_KEY`, `TOKEN`, and `PRIVATE_KEY`. It also re-reads candidate messages by ID and validates exact key boundaries before returning evidence, so `CRABPOT_LCM_FACT_BACKUP` is not accepted as evidence for `CRABPOT_LCM_FACT`.

## Compatibility Notes

This should be easier to merge safely than an assemble-time fix because the new behavior is isolated behind one registered tool and one manifest contract entry. Existing context assembly, compaction, summaries, and engine rotation behavior remain canonical.

Prompt guidance is limited to instructing agents when to choose the explicit tool. The recall itself happens only when the agent calls `lcm_recall_keys`.

## Validation

Ran locally from `/Volumes/LEXAR/repos/lossless-claw-exact-key-recall-tool`:

- `npm test -- --run test/lcm-tools.test.ts test/plugin-prompt-hook.test.ts test/manifest.test.ts`
- `npm run build`
- `git diff --check`
- `npm test -- --maxWorkers=1`

Result: 48 test files passed, 928 tests passed.

I also tried `npx tsc --noEmit`; that is not a repo script/gate and current `main` reports many existing type-environment/strictness diagnostics. After typing the new tool's `execute` signature, no `src/tools/lcm-recall-keys-tool.ts` diagnostics remain in that output.

## Draft Status

Opening as a draft alternative to #765 for review. Do not merge until maintainers choose between the tactical #765 path and this explicit-tool architecture.

### DIFF
--- a/.changeset/lcm-recall-keys-tool.md
+++ b/.changeset/lcm-recall-keys-tool.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add `lcm_recall_keys` for explicit exact-key recall from stored raw messages without mutating assembled context.
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ When a conversation grows beyond the model's context window, OpenClaw (just like
 2. **Summarizes chunks** of older messages into summaries using your configured LLM
 3. **Condenses summaries** into higher-level nodes as they accumulate, forming a DAG (directed acyclic graph)
 4. **Assembles context** each turn by combining summaries + recent raw messages
-5. **Provides tools** (`lcm_grep`, `lcm_describe`, `lcm_expand`) so agents can search and recall details from compacted history
+5. **Provides tools** (`lcm_recall_keys`, `lcm_grep`, `lcm_describe`, `lcm_expand_query`, `lcm_expand`) so agents can search and recall details from compacted history
 
 Nothing is lost. Raw messages stay in the database. Summaries link back to their source messages. Agents can drill into any summary to recover the original detail.
 
@@ -478,6 +478,7 @@ src/
     fts5-sanitize.ts        # FTS5 query sanitization
   tools/
     lcm-grep-tool.ts        # lcm_grep tool implementation
+    lcm-recall-keys-tool.ts # lcm_recall_keys exact-identifier recall tool
     lcm-describe-tool.ts    # lcm_describe tool implementation
     lcm-expand-tool.ts      # lcm_expand tool (sub-agent only)
     lcm-expand-query-tool.ts # lcm_expand_query tool (main agent wrapper)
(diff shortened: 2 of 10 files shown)


100yenadmin:
@vincentkoc this PR accomplishes what you wanted in your PR however it keeps it out of the engine and assembler ie is safer regression wise.

jalehman:
Thanks for opening this alternative. I think the explicit-tool architecture is conceptually aligned with Lossless, but this draft is no longer a good active merge candidate.

Since this PR was opened, #765 was merged and current main now contains the post-rotate exact-key prompt-recall fix, including shared prompt-recall helper logic and regression coverage. This branch is also stale against current main and now conflicts in src/plugin/index.ts and test/lcm-tools.test.ts.

If we still want an explicit lcm_recall_keys tool, I’d prefer a fresh follow-up based on current main that reuses the existing prompt-recall helpers instead of duplicating identifier extraction, boundary matching, and sensitive-value filtering. That would keep the behavior aligned with the merged recall path and make the tool reviewable as an additive follow-up rather than an obsolete alternative to #765.

PR #773, opened 2026-05-31
Fix missing session file afterTurn reread

100yenadmin:
## Summary
- short-circuit afterTurn transcript reconciliation when the tracked session file is missing (`ENOENT`/`ENOTDIR`)
- preserve an existing bootstrap checkpoint, or seed the existing placeholder checkpoint path when no checkpoint exists
- allow live afterTurn persistence to continue while logging an explicit degraded missing-file diagnostic
- add a regression for a large persisted conversation whose tracked transcript has disappeared

Refs #699

## Validation
- `npm test -- test/engine.test.ts -t "missing tracked transcripts|stat-fail fallback"`
- `npm test -- test/engine.test.ts -t "transcript reconcile slow path|stat-fail fallback|missing tracked transcripts|path-mismatched|no-anchor|same-path shrink|capped reconcile|retries a capped reconcile|checkpoint refresh"`
- `npm test -- test/engine.test.ts`
- `npm run build`
- `git diff --check`

## Notes
This keeps #699 open until review/CI/merge. The fix targets the missing-file/orphan transcript path only; path-mismatch and same-path-shrink full reconciliation remain unchanged when the current transcript is readable.

PR #772, opened 2026-05-31
Fix projected raw backlog threshold compaction

100yenadmin:
## Summary
- include raw backlog outside the fresh tail in threshold compaction decisions when runtime observed tokens are available
- persist projected/raw-backlog diagnostics for deferred threshold debt and emergency assemble decisions
- force threshold sweeps for projected raw backlog without treating compactable raw tokens as permanent runtime overhead
- surface projected/raw backlog maintenance diagnostics in `lcm status`

Refs #603

## Validation
- `npm test -- test/compaction-maintenance-store.test.ts test/lcm-integration.test.ts -t "projected|evaluate"`
- `npm test -- test/engine.test.ts -t "projected raw backlog|projected raw backlog pressure|projected deferred pressure|recorded runtime tokens are over budget"`
- `npm test -- test/engine.test.ts test/lcm-integration.test.ts test/compaction-maintenance-store.test.ts test/lcm-command.test.ts`
- `npm run build`
- `git diff --check`

## Notes
Adversarial review initially found two semantic hazards: deferred debt lost projected pressure, and compactable raw backlog was being treated like permanent runtime overhead. This version persists projected pressure separately and uses it only to decide when debt is urgent, while compaction still receives the non-projected observed prompt count.

PR #770, opened 2026-05-30
fix: scrub reasoning fields from compaction summaries

100yenadmin:
## Summary
- Fixes #471 by keeping provider reasoning/thinking fields out of compaction summaries and diagnostics.
- Preserves visible `message.content`, `summary`, and streaming `delta.content` text for vLLM/Qwen-style responses while ignoring sibling `reasoning`, `reasoning_content`, `thinking`, and `redacted_thinking` payloads.
- Prevents fallback warning/detail paths and engine-stored message content from serializing private reasoning text back into summarizer input.

## Validation
- `pwd` confirmed `/Volumes/LEXAR/repos/lossless-claw-p0-reasoning-fields` before local validation.
- `npm exec vitest run test/summarize.test.ts test/engine.test.ts test/lcm-integration.test.ts -- --pool=threads --maxWorkers=1 --minWorkers=1` → 391 tests passed.
- `git diff --check` → no whitespace errors.
- `npm run build` → succeeded.

## Triage note
This is the P0 privacy/data-retention fix for #471. I am opening the PR for review and CI, not merging it from the agent side.

PR #757, opened 2026-05-27
Fix stale session-key rollover replay

dr00-eth:
## Summary

Fixes a stale replay path for stable session keys when the host prunes a tracked transcript file before Lossless observes a lifecycle split.

When `agent:main:main` resumed on a new runtime session id after OpenClaw session cleanup, Lossless could still resolve the active conversation by `sessionKey` and assemble the prior DB tail. In production this repeatedly reintroduced an old assistant message whose content was exactly `openai-codex/gpt-5.5`.

This PR extracts the existing bootstrap-only missing-transcript guard and applies the same lifecycle repair before:

- `assemble()`, so stale context is not returned to a new runtime UUID.
- `afterTurn` transcript reconciliation, so a new transcript is not imported into the archived conversation.

## Evidence

Observed OpenClaw/Lossless logs:

- `assemble: done conversation=838 session=8dea0efd... sessionKey=agent:main:main ... inputMessages=0 outputMessages=40`
- immediately followed by `afterTurn: transcript reconcile slow path could not stat/read transcript; preserving existing checkpoint ... sessionFile=.../8dea0efd....jsonl`
- then `afterTurn: transcript reconcile slow path (full re-read) conversation=838 reason=path-mismatch ... historicalMessages=2 importedMessages=0`

The LCM DB had one historical assistant row with the repeated visible content:

```text
conversation=838 sessionKey=agent:main:main seq=46 role=assistant content=openai-codex/gpt-5.5
```

So the repeated Discord output was stale context replay, not a new model fallback.

## Validation

- `npm test -- --run test/engine.test.ts -t 'rotates before assemble|prior keyed transcript was pruned|stable sessionKey resumes on a new transcript after the old file disappears'`
- `npm test -- --run test/engine.test.ts`
- `npm run build`

PR #774, opened 2026-05-31
fix: defer runtime session file rewrites

### LINKED ISSUE #736: afterTurn hook writes session file during embedded prompt lock release, causing silent reply failure
## Summary

The agent frequently completes tool calls and then silently fails to send any reply — the conversation just stops mid-task with no error shown to the user. Investigation of openclaw logs reveals the root cause: a race condition between the `lossless-claw` plugin's `afterTurn` hook and openclaw's embedded agent prompt lock fence mechanism.

## Environment

- `@martian-engineering/lossless-claw` version: `0.11.2`
- openclaw version: `2026.5.19`
- Platform: macOS (Apple Silicon)

## Observed behavior

The user sees the agent performing tool calls, then receives no reply. No error is surfaced to the user. From openclaw logs:

```
[WARN]  [lcm] dedup: oversized, storedCount=37 batchLen=5, no overlap found — fail-closed skipping full batch
[INFO]  [lcm] assemble: done conversation=406 session=<id> sessionKey=agent:main:...
[ERROR] lane task error: lane=main durationMs=56775
        error="EmbeddedAttemptSessionTakeoverError: session file changed while
        embedded prompt lock was released: .../sessions/<id>.jsonl"
[ERROR] Embedded agent failed before reply: session file changed while
        embedded prompt lock was released: .../sessions/<id>.jsonl
```

## Root cause

When the embedded agent executes tool calls, it releases the prompt lock and installs a file fence (fingerprint: size + mtimeNs + ctimeNs) to detect unauthorized writes. During this window, lossless-claw's `afterTurn` hook — specifically the `assemble` / `dedup` / `auto-rotate` operations — writes to the same session `.jsonl` file without checking for an active prompt lock.

When the embedded agent resumes after the tool call completes, `assertSessionFileFence()` detects the file has been modified and throws `EmbeddedAttemptSessionTakeoverError`, aborting the agent before it can send a reply.

The timing is consistent: `[lcm] assemble: done` appears in the logs immediately before the error, confirming lossless-claw's write is the triggering event.

## Frequency

This occurred at least twice within a single day's log (two separate confirmed incidents), suggesting it is reliably reproducible whenever a long-running tool call overlaps with a lossless-claw `afterTurn` operation on the same session.

## Suggested fix

One of:
1. lossless-claw's `afterTurn` / `assemble` should check for an active embedded prompt lock before writing to the session file and defer the write until the lock is released, or
2. openclaw's embedded agent should use a write lock (not just a fence-based check) to actively prevent concurrent writes from plugins during tool call execution.

100yenadmin:
## Summary
- Defer runtime auto-rotate session JSONL rewrites from `afterTurn()` and `maintain()` so embedded prompt-lock fences are not tripped by active transcript rewrites.
- Keep runtime `rotate` observable by logging deferral to startup/manual rotation, while leaving startup and `/lcm rotate` as direct rewrite paths.
- Gate transcript GC behind host-approved background maintenance and preserve replay-flood protection for user-leading exact replay batches.

Fixes #736

## Validation
- `npm test -- test/engine.test.ts -t "legitimate repeated assistant text|user-leading append-only replay|does not rewrite prior assistant transcript rows|fresh tool-only rows|transcript GC|session-file rewrites|background maintenance|maintain runtime checks|below-threshold|runtime deferral"`
- `npm test -- test/engine.test.ts`
- `npm run build`
- `git diff --check`
- `npm test`
- `npm pack --dry-run`


100yenadmin:
Refreshed after #770, #757, #773, and #772 merged.

I merged current `main` into this branch at head `8ebd2d4` and revalidated runtime session-file rewrite deferral:
- Targeted rewrite-deferral/auto-rotate slice passed: 5 tests.
- `npm test -- test/engine.test.ts`: 251 passed.
- Full local `npm test`: 48 files, 961 tests passed.
- `npm run build`: passed.
- `git diff --check`: passed.
- `npm pack --dry-run`: passed.
- GitHub CI on `8ebd2d4`: both `test` jobs passed and both `smoke-latest-openclaw` jobs passed.

Current assessment: no P0/P1 blocker visible after the refresh. I would keep linked issue #736 open after merge until a current build or reporter confirmation verifies the prompt-lock/session-file rewrite race is gone.

PR #783, opened 2026-05-31
fix(db): allow SQLite extension-capable LCM connections

### LINKED ISSUE #402: sqlite-vec fails to load on OpenClaw 2026.4.10+ (node:sqlite requires allowExtension at construction)
## Summary

LCM's `DatabaseSync` constructor in `src/db/connection.ts:84` does not pass `{ allowExtension: true }`, which prevents `sqlite-vec` from loading on OpenClaw 2026.4.10+.

## Context

OpenClaw 2026.4.10 switched its SQLite backend from `better-sqlite3` to Node's built-in `node:sqlite` (`DatabaseSync`). The two have different extension-loading semantics:

- **`better-sqlite3`**: `db.loadExtension()` works if you call `db.enableLoadExtension(true)` at any time after construction.
- **`node:sqlite`**: `enableLoadExtension(true)` throws `ERR_INVALID_STATE` ("Cannot enable extension loading because it was disabled at database creation") unless `{ allowExtension: true }` was passed to the `DatabaseSync` constructor.

OpenClaw's memory-core host SDK calls `enableLoadExtension(true)` before loading `sqlite-vec`. This succeeds when the database was created with `{ allowExtension: true }`, but LCM's `createLcmDatabaseConnection()` omits this flag.

## Symptoms

On every indexed file (sessions + memory), the gateway logs:

```
chunks written for <path> without vector embeddings — chunks_vec not updated (sqlite-vec unavailable). Vector recall degraded for this file.
```

This causes:
1. **Complete loss of vector recall** — hybrid search falls back to FTS-only
2. **Gateway I/O saturation on startup** — the reindex churns through every session/memory file doing disk I/O for writes that produce no usable vectors, starving WebSocket handshakes and causing TUI connection timeouts

## Affected versions

- **LCM**: v0.5.1 (and likely all versions using `node:sqlite`)
- **OpenClaw**: 2026.4.10+ (first version using `node:sqlite` instead of `better-sqlite3`)
- **Node**: v24.14.0 (but applies to any Node version with `node:sqlite`)

## Fix

One-line change in `src/db/connection.ts`, line 84:

```diff
- const db = configureConnection(new DatabaseSync(dbPath));
+ const db = configureConnection(new DatabaseSync(dbPath, { allowExtension: true }));
```

Verified: after this patch, `sqlite-vec` v0.1.9 loads successfully and the `chunks_vec` warnings disappear entirely.

## Reproduction

```typescript
import { DatabaseSync } from "node:sqlite";

// This fails:
const db1 = new DatabaseSync(":memory:");
db1.enableLoadExtension(true); // throws ERR_INVALID_STATE

// This works:
const db2 = new DatabaseSync(":memory:", { allowExtension: true });
db2.enableLoadExtension(true); // OK
db2.loadExtension("vec0");     // OK
```

100yenadmin:
## Summary

Fixes #402.

This makes LCM-owned `DatabaseSync` handles extension-capable at construction time by passing `{ allowExtension: true }`, which is required by node:sqlite before callers can explicitly enable/load SQLite extensions such as sqlite-vec.

The connection configuration immediately disables extension loading at rest with `enableLoadExtension(false)`, so the handle has the capability but does not leave extension loading open by default.

## Validation

- `npm test -- test/db-connection.test.ts`
- `npm run build`
- `npm test` (49 files / 1020 tests)
- `git diff --check`
- `npm pack --dry-run`

## Notes

This PR does not add sqlite-vec loading itself; it removes the node:sqlite construction-time blocker so explicit extension loading can work safely where the host/runtime needs it.

### DIFF
--- a/.changeset/sqlite-extension-loading.md
+++ b/.changeset/sqlite-extension-loading.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Allow LCM SQLite handles to explicitly enable extension loading for sqlite-vec on node:sqlite hosts.
--- a/src/db/connection.ts
+++ b/src/db/connection.ts
@@ -60,7 +60,20 @@ function configureConnection(db: DatabaseSync): DatabaseSync {
   db.exec("PRAGMA synchronous = NORMAL");
   // Keep temp tables/indexes in RAM (helps ordinal resequencing).
   db.exec("PRAGMA temp_store = MEMORY");
+  // Keep extension loading disabled by default, while still allowing explicit
+  // enablement later on handles created with allowExtension: true.
+  if (typeof db.enableLoadExtension === "function") {
+    db.enableLoadExtension(false);
+  }
   return db;
+}
+
+function createDatabaseSync(dbPath: string): DatabaseSync {
+  const supportsExtensionLoading =
+    typeof DatabaseSync.prototype.enableLoadExtension === "function";
+  return supportsExtensionLoading
+    ? new DatabaseSync(dbPath, { allowExtension: true })
+    : new DatabaseSync(dbPath);
 }
 
 function trackConnection(dbPath: string, db: DatabaseSync): void {
@@ -112,7 +125,7 @@ function closeDatabase(db: DatabaseSync | undefined): void {
  */
 export function createLcmDatabaseConnection(dbPath: string): DatabaseSync {
   ensureDbDirectory(dbPath);
-  const db = new DatabaseSync(dbPath);
+  const db = createDatabaseSync(dbPath);
   try {
     configureConnection(db);
   } catch (err) {
--- a/test/db-connection.test.ts
+++ b/test/db-connection.test.ts
@@ -1,11 +1,20 @@
-import { describe, expect, it } from "vitest";
+import { mkdtempSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
 import {
+  closeLcmConnection,
+  createLcmDatabaseConnection,
   getFileBackedDatabasePath,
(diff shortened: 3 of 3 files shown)


100yenadmin:
Adversarial review found and fixed a Node 22 compatibility blocker on the first head.

Issue found: unconditionally calling `db.enableLoadExtension(false)` would crash on Node 22 builds that expose `node:sqlite` before the extension-loading API.

Fix pushed in `cd6822f`:
- feature-detect `DatabaseSync.prototype.enableLoadExtension` before passing `{ allowExtension: true }`
- guard `db.enableLoadExtension(false)` before calling it
- keep current modern-Node behavior for sqlite-vec capable runtimes without raising the documented Node 22 floor

Re-run local validation after the fix: focused DB connection test, `npm run build`, full `npm test` (49 files / 1020 tests), `git diff --check`, and `npm pack --dry-run`.

100yenadmin:
Maintainer review: passed, 10/10 mergeable.

**Summary**
This is ready to merge on current head `cd6822f49076af31b6148ee14565a74fc9e355cb`. The adversarial P1 concern from the first pass was fixed by feature-detecting `DatabaseSync.prototype.enableLoadExtension`, only constructing with `{ allowExtension: true }` when the API exists, and guarding the disable call. That keeps OpenClaw/node:sqlite extension loading enabled where supported without crashing older Node builds.

**Evidence checked**
- Exact-head GitHub CI is green: both `test` runs and both `smoke-latest-openclaw` runs succeeded.
- GitHub reports `MERGEABLE` / `CLEAN`; PR is not draft.
- Review-thread query returned no unresolved threads.
- Local validation on the hardened head passed: `npm test -- test/db-connection.test.ts`, `npm run build`, full `npm test`, `git diff --check`, and `npm pack --dry-run`.

**Next step**
Merging this as the concrete fix for #402.

PR #784, opened 2026-05-31
fix(assembler): strip foreign thinking signatures

### LINKED ISSUE #365: Assembler preserves foreign thinkingSignature, causing 400 errors on cross-provider model switch
## Summary

When a session switches between different model providers (e.g., Qwen/DashScope fallback → Anthropic Claude primary), lossless-claw's assembler preserves the original provider's `thinkingSignature` in thinking blocks read from the DB. Anthropic's API rejects these foreign signatures with a 400 error, which OpenClaw then interprets as a primary model failure — triggering another fallback and creating a loop.

## Root Cause

`blockFromPart()` in `assembler.ts` (around line 386) returns `metadata.raw` directly for non-tool blocks. This raw object includes the `thinkingSignature` field from whichever provider originally produced the thinking block.

The pipeline gap:
1. OpenClaw's `sanitizeSessionHistory()` runs **before** lossless-claw's `assemble()`
2. `assemble()` reads from its own SQLite DB (not the sanitized messages)
3. Assembled messages **replace** the sanitized ones
4. No provider-specific sanitization runs on the assembled output

`tryRestoreOpenAIReasoning()` handles OpenAI reasoning blocks specifically, but Qwen and other provider signatures pass through unmodified.

## Steps to Reproduce

1. Configure OpenClaw with an Anthropic primary model and a non-Anthropic fallback (e.g., `dashscope/qwen3.6-plus`)
2. Trigger fallback (e.g., VPN disconnect) — Qwen produces thinking blocks with its own `thinkingSignature`
3. Restore primary connectivity and switch back to Anthropic model
4. Continue chatting until LCM compaction triggers
5. Assembled context includes Qwen's `thinkingSignature` → Anthropic API returns 400 → fallback loop

## Suggested Fix

In `blockFromPart()`, after `tryRestoreOpenAIReasoning()` returns null and before `return metadata.raw`, strip `thinkingSignature` from `type: "thinking"` blocks:

```typescript
if (
  rawType === "thinking" &&
  typeof (metadata.raw as Record<string, unknown>).thinkingSignature === "string"
) {
  const { thinkingSignature: _stripped, ...cleaned } = metadata.raw as Record<string, unknown>;
  return cleaned;
}
return metadata.raw;
```

This is safe because:
- OpenAI reasoning blocks are already handled by `tryRestoreOpenAIReasoning()` before this point
- For DB-sourced (assembled) blocks, signatures are stale anyway — they were valid only for the original streaming session
- The `thinking` content text is preserved; only the provider-specific signature metadata is removed

## Environment

- lossless-claw 0.6.3 (also checked 0.7.0 and 0.8.0 changelogs — not fixed)
- OpenClaw 2026.4.9
- Primary: Anthropic Claude Opus 4.6 (via proxy)
- Fallback: DashScope Qwen 3.6 Plus (direct)

100yenadmin:
## Summary

Fixes #365.

This keeps OpenAI reasoning restoration intact, but strips stale foreign `thinkingSignature` metadata from DB-sourced `type: "thinking"` blocks before assembly returns the raw block. That prevents provider-specific signatures from being replayed after a provider/model switch.

## Validation

- `npm test -- test/assembler-blocks.test.ts`
- `npm run build`
- `npm test` (49 files / 1020 tests)
- `git diff --check`
- `npm pack --dry-run`

## Notes

This is intentionally narrow: it only removes `thinkingSignature` after the existing OpenAI JSON reasoning-signature restoration path has had a chance to convert valid OpenAI reasoning metadata back to `{ type: "reasoning", id: ... }`.

### DIFF
--- a/.changeset/strip-foreign-thinking-signature.md
+++ b/.changeset/strip-foreign-thinking-signature.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Strip stale foreign thinking signatures from DB-sourced thinking blocks during assembly.
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -500,7 +500,17 @@ export function blockFromPart(part: MessagePartRecord): unknown {
     // arguments (stringify if object) and format for the target provider.
     // Returning raw here causes arguments to be passed as a JS object instead
     // of a JSON string, which breaks xAI/OpenAI Chat Completions API (422).
-    const rawType = (metadata.raw as Record<string, unknown>).type as string | undefined;
+    const rawRecord = metadata.raw as Record<string, unknown>;
+    const rawType =
+      typeof rawRecord.type === "string" ? rawRecord.type : metadata.rawType;
+    if (
+      rawType === "thinking" &&
+      typeof rawRecord.thinkingSignature === "string"
+    ) {
+      const { thinkingSignature: _thinkingSignature, ...cleaned } = rawRecord;
+      return cleaned;
+    }
+
     const isToolBlock =
       rawType === "toolCall" ||
       rawType === "tool_use" ||
@@ -520,7 +530,6 @@ export function blockFromPart(part: MessagePartRecord): unknown {
     // from the DB columns.  For rows stored as part_type='text' those columns are
     // often NULL — the values only live inside metadata.raw.  Backfill them here
     // so the reconstructed block keeps the original id/name.
-    const rawRecord = metadata.raw as Record<string, unknown>;
     const rawToolCallId =
       typeof rawRecord.id === "string" && rawRecord.id.length > 0
         ? rawRecord.id
--- a/test/assembler-blocks.test.ts
+++ b/test/assembler-blocks.test.ts
@@ -392,6 +392,47 @@ describe("blockFromPart", () => {
     expect(block.id).toBe("rs_abc123");
   });
 
+  it("strips foreign thinkingSignature from DB-sourced thinking blocks", () => {
+    const part = makePart({
+      partType: "reasoning",
+      textContent: "provider-specific thinking text",
+      metadata: JSON.stringify({
+        raw: {
+          type: "thinking",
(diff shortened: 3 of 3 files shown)


100yenadmin:
Adversarial follow-up applied on current head `d3179b2`.

Turing’s partial review did not find a P0/P1 blocker, but it did call out one worthwhile hardening caveat: older rows could still carry a foreign `thinkingSignature` when `metadata.rawType` identifies the block as `thinking` but `metadata.raw.type` is absent or malformed.

I patched that by deriving `rawType` from `raw.type` when present, otherwise falling back to `metadata.rawType`, and added a regression test for that legacy-row shape.

Validation rerun locally from `/Volumes/LEXAR/repos/lossless-claw-issue-365-thinking-signature`:
- `npm test -- test/assembler-blocks.test.ts`
- `npm run build`
- `git diff --check`
- `npm pack --dry-run`

Waiting on exact-head GitHub CI before treating this as 10/10 mergeable.

100yenadmin:
Maintainer review: passed, 10/10 mergeable.

**Summary**
This is ready to merge on current head `d3179b22f8ca2cb934f7866013bcecb03668e20f`. It strips foreign provider thinking signatures during assembly while preserving OpenAI reasoning encrypted content through the existing restoration path, and the adversarial legacy-row caveat has been covered.

**Evidence checked**
- Exact-head GitHub CI is green: both `test` runs and both `smoke-latest-openclaw` runs succeeded.
- GitHub reports `MERGEABLE` / `CLEAN`; PR is not draft.
- Review-thread query returned no unresolved threads.
- Local validation from `/Volumes/LEXAR/repos/lossless-claw-issue-365-thinking-signature` passed: `npm test -- test/assembler-blocks.test.ts`, `npm run build`, `git diff --check`, and `npm pack --dry-run`.
- Adversarial follow-up covered the legacy `metadata.rawType` case where `raw.type` is missing or malformed.

**Next step**
Merging this as the concrete fix for #365.

PR #777, opened 2026-05-31
fix: gate doctor apply on hot conversations

100yenadmin:
## Summary
- Add a safe-by-default doctor apply preflight for large/hot current conversations.
- Block inline repair before summarizer calls when target count, message count, observed token pressure, or pending/running maintenance indicates active-channel risk.
- Add explicit `doctor apply confirm-offline` override for isolated/offline maintenance lanes.
- Add regressions proving large repairs and pending maintenance do not call the summarizer or mutate summaries, while explicit confirmation still repairs.

Refs #491.

## Validation
- `npm test -- test/lcm-command.test.ts -t "doctor apply|parses focus command forms"`
- `npm test -- test/lcm-command.test.ts`
- `npm test` (49 files, 1012 tests)
- `npm run build`
- `git diff --check`
- `npm pack --dry-run`

## Notes
- I am not closing #491 from this PR alone until CI and adversarial rereview complete and reporter/maintainer confirmation decides whether additional degraded assembly/status work is still required.

Issue #491, opened 2026-04-24
doctor apply can leave active channel sessions unresponsive via LCM assembly/context overflow


Adam-Researchh:
## Summary

Running `/lossless doctor apply` on a large long-lived conversation can leave the system in a state where channel sessions become temporarily unresponsive or delayed, even when the channel transport itself is healthy. In the observed case, Telegram was configured and able to send messages, but the agent turn behind the channel hit LCM assembly/context overflow after doctor/apply and related summary repair operations.

This issue is about operational isolation and recovery: doctor repair should not be able to monopolize or poison an active channel session, and it should provide a safe dry-run/apply path for large conversations.

## Environment

- OpenClaw: 2026.4.22
- lossless-claw: 0.9.2
- Channel: Telegram (transport healthy)
- Conversation involved: long-lived main session with millions of stored tokens due to duplicate import
- Doctor target count: hundreds of fallback/truncated summaries observed in the DB

## Observed sequence

1. Operator ran lossless doctor/apply to repair fallback summaries.
2. Gateway/channel transport was healthy after restart:

```text
Channels: Telegram ON OK
sendMessage ok ...
```

3. However, channel requests became delayed/unresponsive because the agent turn for the session repeatedly failed during LCM assembly/recovery:

```text
[lcm] assemble: done conversation=146 ... tokenBudget=500000 estimatedTokens=499239
embedded_run_agent_end ... context_length_exceeded
context overflow detected (attempt 1/3); attempting auto-compaction
context overflow detected (attempt 2/3); attempting auto-compaction
context overflow detected (attempt 3/3); attempting auto-compaction
```

4. The conversation also had pending compaction debt that maintenance could not process in that runtime context:

```text
[lcm] maintain: deferred compaction debt pending ... but host runtimeContext.allowDeferredCompactionExecution is disabled
```

5. The practical result was that the channel appeared “bricked” even though the bot token and network were OK.

## Expected behavior

- Doctor/apply should be operationally isolated from active delivery sessions.
- If a repair target set is large or a conversation is already over a safe token threshold, doctor/apply should warn and require staged repair or offline mode.
- After doctor/apply, active session assembly should not repeatedly fail with context overflow.
- Deferred compaction debt needed for safety should either run in a safe background lane or clearly report that the active session is unsafe until maintenance completes.

## Actual behavior

[...]

PR #786, opened 2026-05-31
fix: degrade LCM assembly when maintenance is hot

100yenadmin:
## Summary
- Add a high-pressure maintenance guard in assemble() so active delivery falls back to bounded live context instead of replaying stored LCM history when deferred compaction debt is pending/running near the assembly budget.
- Keep the emergency over-budget drain path first, but degrade if emergency maintenance remains pending or fails.
- Extend /lossless status with LCM health vs transport-health diagnostics, including the same 75% repair-source pressure warning used by doctor apply.

Refs #491.

## Validation
- `npx vitest run test/lcm-command.test.ts -t "repair-source pressure|token pressure|deferred compaction maintenance state"`
- `npx vitest run test/engine.test.ts -t "pending maintenance is near budget|emergency compaction leaves debt pending"`
- `npx vitest run test/lcm-command.test.ts`
- `npx vitest run test/engine.test.ts -t "deferred threshold debt|pending maintenance is near budget|emergency compaction leaves debt pending|assemble\(\) drains pending threshold debt|assemble\(\) does not wait|assemble\(\) waits"`
- `npm test` (49 files, 1029 tests)
- `npm run build`
- `npm pack --dry-run`
- `git diff --check`

## Notes
- This covers the remaining degraded delivery/status lane after #777. Issue #491 should remain open until reporter/maintainer confirmation on the original channel-unresponsive symptom.
- Full required checks are expected to run in GitHub Actions on this PR.

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -58,6 +58,7 @@ import {
 } from "./store/compaction-telemetry-store.js";
 import {
   CompactionMaintenanceStore,
+  type ConversationCompactionMaintenanceRecord,
 } from "./store/compaction-maintenance-store.js";
 import {
   ConversationStore,
@@ -108,6 +109,7 @@ type BootstrapImportObservation = {
 const MAX_PREVIOUS_ASSEMBLED_SNAPSHOTS = 100;
 const FORK_BOUNDED_BOOTSTRAP_REASON = "fork-bounded bootstrap import";
 const CONTEXT_ENGINE_PROJECTION_EPOCH_VERSION = "summary-prefix-v1";
+const DEFERRED_ASSEMBLY_DEGRADED_PRESSURE_RATIO = 0.75;
 type CircuitBreakerState = {
   failures: number;
   openSince: number | null;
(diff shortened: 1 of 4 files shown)


100yenadmin:
Adversarial follow-up addressed on the latest head `a00e7ac`.

Changes after the first review pass:
- Degraded live fallback now preserves protected leading `system` / `developer` live context before bounding the active live tail, so it does not drop runtime policy while avoiding stored LCM history under hot maintenance.
- `/lossless status` now only uses maintenance `currentTokenCount` / `projectedTokenCount` as assembly pressure while maintenance is pending or running, so stale idle counts do not produce false degraded status.
- Added regressions for both blockers plus the previous repair-pressure warning and hot-maintenance degraded assembly cases.

Latest local validation from `/Volumes/LEXAR/repos/lossless-claw-issue-491-degraded-status`:
- `npm test` (49 files, 1031 tests)
- `npm run build`
- `npm pack --dry-run`
- `git diff --check`

PR #803, opened 2026-05-31
fix: fail closed on compaction fallback summaries

### LINKED ISSUE #605: Runaway recursive compaction causes permanent context loss when summarize provider fails
## Summary

In a stressed window (summarize provider rate-limited + multiple gateway restarts) with `incrementalMaxDepth: -1` and `dynamicLeafChunkTokens.enabled: true`, LCM produced a burst of small leaves followed by a depth-1 condensed summary that prematurely rolled up still-immature leaves. The committed condensed summary is then loaded by every subsequent gateway bootstrap, **permanently collapsing usable context**. The user-visible effect: the conversation's usable context dropped from a healthy ~50K assembled tokens to ~13% (~33K of 258K budget) and **did not recover after restart**. Recovery required manually deleting the offending condensed row from `lcm.db` (raw messages were preserved, so data is recoverable — but only by hand).

## Environment

- LCM version: `@martian-engineering/lossless-claw@0.9.4`
- DB: `~/.openclaw/lcm.db` (SQLite)
- Conversation: telegram conv 1876 (Eva, active)
- Provider during incident: `openai-codex` summarizer rate-limited (`FailoverError: ChatGPT prolite plan, try again in ~61 min`)
- Multiple gateway restarts during the rate-limited window (3-5 in ~1 hour)

Plugin config (relevant fields, redacted otherwise):
```jsonc
{
  "incrementalMaxDepth": -1,
  "dynamicLeafChunkTokens": { "enabled": true, "max": 40000 }
  // ... operator opted into unbounded recursive depth + variable chunk sizing
}
```

## What we observed

Earlier in the day: 31 leaves on conv 1876, healthy assembly.
After the incident: 55 leaves and the appearance of a depth-1 condensed summary that swallowed several still-immature leaves.

Smoking-gun chronology from `summaries` table (~2-minute window):

```
22:30:13  leaf      depth=0  2415 tokens
21:40:08  condensed depth=1  6000 tokens   <-- the smoking gun
21:39:33  leaf      depth=0   890
21:39:24  leaf      depth=0   802
21:39:18  leaf      depth=0   685
21:39:09  leaf      depth=0   488
21:39:03  leaf      depth=0   629
21:38:57  leaf      depth=0   544
21:38:53  leaf      depth=0   966
21:38:45  leaf      depth=0   736
```

Thirteen tiny leaves (~500-1000 tokens each) created in 2 minutes, immediately followed by a depth-1 condensed summary that compressed several of them. Around the same window:

- `compactLeafAsync` was firing with `maxPasses=2 activityBand=medium`
- openai-codex auth was rate-limited at 05:30 with `FailoverError: ChatGPT prolite plan, try again in ~61 min`
- 3-5 gateway restarts happened during the same hour

## Why this is bad (sticky context loss)

LCM bootstrap re-derives assembled context from the DB. Once the depth-1 condensed summary was committed, **every subsequent restart loads the condensed view in place of the original leaves**. The compaction is sticky: restart did not recover the lost context. Manual SQL surgery (deleting the offending condensed row; raw messages were intact) was required to restore usable history.

## Hypothesised root cause

This looks like the intersection of three independently-reasonable behaviors that compound under provider stress:

1. **`incrementalMaxDepth: -1`** allows recursion to escalate without an upper bound, so partial/broken passes can still produce a depth-1 commit even when the underlying leaves are nowhere near the size they should be.
2. **`dynamicLeafChunkTokens.enabled`** adapts leaf chunk size under pressure, so when the summarizer is failing/timing out, retried leaf passes shrink to tiny chunks (~500-1000 tokens). This produces the leaf-burst pattern above.
3. **Compaction does not back off when the summarize provider is failing.** Rate-limit errors (`FailoverError`) are treated like any other transient failure; retries keep firing, and depth escalation continues to roll up the tiny / partial / failed leaves.

Combined, a stressed window can promote a half-formed rollup to depth-1 (or higher), commit it, and quietly amputate context permanently.

## Proposed mitigations

Roughly in order of "smallest sufficient fix" → "more invasive":

1. **Don't commit a condensed summary if the underlying summarize call failed or timed out.** Today, partial/broken summarize results can still hit `summaries`. A failed summarize should be a no-op (defer/retry later), not a commit. (Possibly related: #494 fallback-summary copy-through, #584 withTimeout doesn't abort — but the bug here is *committing the failed result*, not the timeout/abort plumbing itself.)
2. **Cap depth recursion when error rate is elevated.** A provider-error-rate gauge over the recent N minutes should hard-clip `incrementalMaxDepth` to 0 (or 1) regardless of operator config. `-1` means "trust me, run as deep as you need" — but only when the summarizer is healthy.
3. **Exponential backoff on summarize failure.** Rate-limit / `FailoverError` should pause `compactLeafAsync` for at least the indicated retry window, not just bounce on the next turn. Current behavior appears to be "keep firing."
4. **Don't roll up leaves that are below a minimum maturity size.** A depth-1 condensed pass over 6 leaves averaging ~700 tokens each is suspicious on its face; require a minimum leaf size (or count × size) before promotion. The dynamic-leaf path makes this concretely worse — its smaller chunks shouldn't get rolled up the same way normal-sized chunks do.
5. **Operator-visible warning when consuming raw leaves under provider-error conditions.** If the summarizer has been failing for >N minutes, surface a banner / log indicating that compaction may produce damaged commits and recommend deferring.
6. **At minimum: when a depth-N condensed commit lands, log enough provenance (which leaves it consumed, what the summarize call's status was, whether it was a retry-after-failure) that operators can trace which condensed row to delete on recovery.** Today the recovery path is "open SQLite, eyeball timestamps, guess."

## Recoverability

Important: the user's data **is recoverable** — raw messages survived, only the `summaries` row needed to be deleted. So this isn't data loss in the strict sense. It's *assembled context* loss that is sticky across restarts and only the operator (with SQL access) can repair. That's still a significant degradation: the conversation is effectively half-amnesiac until manual surgery.

## Related (NOT duplicates)

- #74 — no cost circuit breaker for summarization. This issue covers cost; the present bug is about *correctness under provider failure* (a damaging commit happens even when no extra spend is involved).
- #494 — fallback summaries can copy raw source text. Different failure mode (copy-through, not premature rollup), but adjacent: both result in damaging committed summaries.
- #553 — `cacheAwareCompaction` never drains debt for active Codex sessions. Effectively the *opposite* problem (compaction never fires); the present bug is compaction firing too aggressively and committing damage.
- #584 — `withTimeout` doesn't abort underlying summarize call. Adjacent infra; the bug here would still occur even if abort worked, because the issue is *committing a partial / failed summary* rather than the cancellation plumbing.
- #436 — Ollama cloud summaries return empty normalized output and degrade to truncation fallback. Related family ("compaction commits a broken result"), but provider-specific.

## Reproduction sketch

I don't have a clean repro harness, but the ingredients are clear:

1. Long-running conversation with active summarization.
2. Set `incrementalMaxDepth: -1` and `dynamicLeafChunkTokens.enabled: true, max: 40000`.
3. Force the summarize provider into a sustained failing state (e.g. mock `deps.complete` to throw `FailoverError` for a window).
4. Trigger gateway restarts during the failing window.
5. Observe the `summaries` table after a few restarts: expect a leaf burst followed by a small-but-prematurely-promoted depth-1 condensed summary.

100yenadmin:
## Summary
- Fixes #605
- Treat model-backed `[LCM fallback summary...]` output as a non-auth provider failure inside compaction instead of persisting it into the summary DAG.
- Propagate `providerFailure` through threshold sweeps and budget recovery so deferred debt remains pending/backed off without recording an auth failure.
- Cover both the leaf write path and the depth-1 condensed write path that caused sticky context collapse.

## Validation
- `npx vitest run test/lcm-integration.test.ts -t "model-backed summarizer|condensed summary persistence"`
- `npx vitest run test/lcm-integration.test.ts`
- `npx vitest run test/summarize.test.ts`
- `npx vitest run test/engine.test.ts -t "provider fallback failure|uses threshold target for proactive threshold compaction mode"`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/compaction.ts
+++ b/src/compaction.ts
@@ -11,7 +11,7 @@ import type { SummaryStore, SummaryRecord, ContextItemRecord } from "./store/sum
 import { estimateTokens, truncateTextToEstimatedTokens } from "./estimate-tokens.js";
 import { extractFileIdsFromContent } from "./large-files.js";
 import { NOOP_LCM_LOGGER, type LcmLogger } from "./lcm-log.js";
-import { LcmProviderAuthError } from "./summarize.js";
+import { isLcmDeterministicFallbackSummary, LcmProviderAuthError } from "./summarize.js";
 
 // ── Public types ─────────────────────────────────────────────────────────────
 
@@ -44,6 +44,8 @@ export interface CompactionResult {
   level?: CompactionLevel;
   /** Whether compaction was blocked by a provider auth failure */
   authFailure?: boolean;
+  /** Whether compaction was blocked by non-auth provider failure/degraded model fallback */
+  providerFailure?: boolean;
 }
 
 export interface CompactionConfig {
@@ -103,6 +105,7 @@ export interface CompactionConfig {
 }
 
 type CompactionLevel = "normal" | "aggressive" | "fallback" | "capped";
+type CompactionFailure = "auth" | "provider";
 type CompactionPass = "leaf" | "condensed";
 type CompactionSummarizeOptions = {
   previousSummary?: string;
@@ -122,6 +125,10 @@ type PassResult = {
   /** Token count of the newly created summary. */
   addedTokens: number;
 };
+type PassFailure = { failure: CompactionFailure };
+type SummaryEscalationResult =
+  | { content: string; level: CompactionLevel }
+  | PassFailure;
 type LeafChunkSelection = {
   items: ContextItemRecord[];
   rawTokensOutsideTail: number;
@@ -753,13 +760,13 @@ export class CompactionEngine {
       previousSummaryContent,
       input.summaryModel,
     );
-    if (!leafResult) {
+    if (!leafResult || "failure" in leafResult) {
       return {
         actionTaken: false,
         tokensBefore,
         tokensAfter: tokensBefore,
         condensed: false,
-        authFailure: true,
+        ...(leafResult?.failure === "provider" ? { providerFailure: true } : { authFailure: true }),
       };
     }
     // Delta tracking: compute token change from pass results instead of re-querying DB
@@ -798,7 +805,29 @@ export class CompactionEngine {
           summarize,
           input.summaryModel,
         );
-        if (!condenseResult) {
+        if (!condenseResult || "failure" in condenseResult) {
+          if (condenseResult?.failure === "provider") {
+            return {
+              actionTaken: true,
+              tokensBefore,
+              tokensAfter,
+              createdSummaryId,
+              condensed,
+              level,
+              providerFailure: true,
+            };
+          }
+          if (condenseResult?.failure === "auth") {
+            return {
+              actionTaken: true,
+              tokensBefore,
+              tokensAfter,
+              createdSummaryId,
+              condensed,
+              level,
+              authFailure: true,
+            };
+          }
           break;
         }
         const passTokensAfter = passTokensBefore - condenseResult.removedTokens + condenseResult.addedTokens;
@@ -899,6 +928,7 @@ export class CompactionEngine {
     let previousSummaryContent: string | undefined;
     let previousTokens = tokensBefore;
     let hadAuthFailure = false;
+    let hadProviderFailure = false;
     let stoppedForNoProgress = false;
 
     // Sweep bounds: a single full sweep must not run an unbounded number of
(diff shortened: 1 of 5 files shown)


jalehman:
Deterministic fallback summaries are an intentional part of the design and shouldn't be treated as a failure. /lossless doctor apply will repair them if the LLM call failures were transient.

100yenadmin:
Closing pull request at lead maintainer request for architecture decisional.

PR #755, opened 2026-05-27
fix(anti-replay): role-aware threshold to absorb legitimate sub-agent bursts

rafaelreis-r:
## Summary

Fixes a false-positive in `assertNoReplayTimestampFlood` that wedges legitimate cron/sub-agent workloads. Related to (and likely a sibling-mechanism of) #639 — surfaces with the same downstream symptoms (compaction skipped, reconcile failed, sessions stuck), distinct root cause inside the same guard.

**Status:** validated in production — patched build deployed on a Mac mini gateway running `openclaw 2026.5.22` + a forked `lossless-claw@0.11.2`. Smoke test below.

## Problem

`assertNoReplayTimestampFlood` (`src/store/conversation-store.ts:992` on `main`) throws:

```
[lcm] refused replay-like message batch: conversation=X createdAt=YYYY-MM-DD HH:MM:SS replicatedRows>=3
```

whenever ≥3 messages with identical `(conversation_id, role, content, created_at-in-seconds)` are seen. SQLite's `datetime('now')` only has second-level granularity, so any **legitimate fast burst** of identical tool/assistant outputs from a sub-agent trips the guard. Ingest aborts → compaction is skipped (`afterTurn: ingest failed, skipping compaction`) → reconcile fails → third-party monitors flag the session as unhealthy.

### Production repro (conversation_id=3790)

Cron-driven sub-agent (`gbrain-maintain-orphans-daily`):

- ~50 turns over ~5 min
- Each turn emits: 1 `assistant` message + 2 identical `{"status":"ok"}` `tool` messages
- One run produced **17 `skipping compaction` + 20 reconcile failures + 74 `auto-rotate ... ingest-failed` log lines** on the same `sessionKey`
- SQLite confirms 50+ identical tool messages clustered in second-aligned buckets

Direct DB grouping:

```sql
SELECT role, substr(content,1,80) AS preview, created_at, COUNT(*) AS dup
FROM messages WHERE conversation_id=3790
GROUP BY role, content, created_at HAVING dup>=2 ORDER BY dup DESC LIMIT 5;
-- tool|{"status":"ok"}|2026-05-27 07:01:20|4
-- tool|{"status":"ok"}|2026-05-27 07:01:30|2
-- tool|{"status":"ok"}|2026-05-27 07:01:37|2
-- tool|{"status":"ok"}|2026-05-27 07:01:44|2
-- tool|{"status":"ok"}|2026-05-27 07:02:44|2
```

### Why a UNIQUE constraint is not the right fix

[...]

Issue #639, opened 2026-05-08
BUG: Lossless-Claw: Dedup Fail-Closed and Deferred Compaction Loop


najef1979-code:
**Lossless‑Claw: Dedup Fail‑Closed, Deferred Compaction & Hot‑Cache Heuristic Issues**
I’ve opened several issues recently because I’m running a multi‑agent, multi‑channel OpenClaw setup that stresses the runtime heavily. After reviewing logs and source code, I believe several symptoms originate from one underlying subsystem: the Lossless‑Claw compaction + deduplication engine.

This issue consolidates the related failure modes so maintainers can evaluate them together.

Lossless-Claw's context management exhibits two related failure modes that compound each other, causing silent message loss and context growth without proper compaction:

1. **Failure Mode 1 - Dedup Fail-Closed**: When no overlap is found between batch and stored messages, the entire batch is skipped (returning empty array)
2. **Failure Mode 2 - Deferred Compaction Loop**: Incremental compaction is indefinitely deferred under hot-cache conditions

These failure modes create a feedback loop: deferred compaction → context growth → oversized batches → dedup fail-closed → message loss.

# Diagnostic Evidence

**OpenClaw Version:** 2026.5.7 (eeef486)
**Lossless-Claw Version:** 0.9.4

### Gateway Stability (CRITICAL)

| Metric | Value | Severity |
|--------|-------|----------|
| Dropped events | 2,324 | CRITICAL |
| Event loop delay P99 | 2447ms | HIGH |
| Event loop utilization | 92.8% | HIGH |
| Memory pressure count | 2 | WARNING |

### Log Evidence - Dedup Fail-Closed

```
2026-05-08T18:25:25.264Z warn [lcm] dedup: oversized, storedCount=79 batchLen=2, no overlap found — ingesting full batch
2026-05-08T18:25:25.265Z info [lcm] ingestBatch: session queue acquired queueKey=agent:marcus:main messages=2
```

Note: The log says "ingesting full batch" but the code returns `[]` (empty array), dropping messages.

### Log Evidence - Deferred Compaction

```
2026-05-08T18:25:25.282Z info [lcm] deferred compaction debt recorded: conversation=32 reason=leaf-trigger tokenBudget=200000 currentTokenCount=74679
2026-05-08T18:25:25.294Z info [lcm] maintain: deferred compaction skipped session=... changed=false ok=true reason=deferred compaction still needed
[...]

PR #806, opened 2026-06-02
fix(bootstrap): fail open on replay-flood guard instead of quarantining the whole context engine (#639)

rafaelreis-r:
## Problem (the dedup-fail-closed half of #639)

When the context engine bootstraps a conversation whose **persisted history contains a same-second burst of identical `user` messages**, `ConversationStore.assertNoReplayTimestampFlood` throws `refused replay-like message batch` during the bootstrap re-import.

`user`-role messages are **never** exempted by `filterBootstrapReplayMessages` (only `assistant`/`tool` count as bootstrap replay candidates), so the throw escapes straight out of `bootstrap()`. The host treats a bootstrap throw as fatal:

```
[context-engine] Context engine "lossless-claw" failed during bootstrap:
  [lcm] refused replay-like message batch: conversation=3113 role=user
  createdAt=... replicatedRows=3 threshold=3;
  quarantining it for this process and falling back to default engine "legacy".
```

→ **One bad conversation silently disables lossless context management for the entire process** — every session falls back to the legacy engine.

### Observed in production

A single durable `agent:*:main` conversation had hundreds of identical `user` rows at one `created_at` second (a historical burst folded into the main session; second-granularity `datetime('now')` collisions also arise from bulk transcript imports). On **every** gateway restart it re-tripped the guard at bootstrap and quarantined the engine, so unrelated sessions silently ran with **no** context management at all. After this patch, the same conversation logs:

```
[lcm] bootstrap: replay-flood guard tripped for sessionKey=agent:main:main;
  skipping this conversation's bootstrap to keep the context engine available
  for all other sessions (#639 fail-open). [lcm] refused replay-like message
  batch: conversation=3113 role=user replicatedRows=3 threshold=3
```

— engine stays up; only the offending conversation is skipped.

## Fix

- `assertNoReplayTimestampFlood` now throws a typed, exported **`ReplayTimestampFloodError`** (with a bundler-safe marker + message-prefix detection via `isReplayTimestampFloodError`).
- `bootstrap()` **catches it and fails OPEN**: it skips just that conversation and returns a degraded result (`reason: "replay-flood guard tripped at bootstrap; conversation skipped"`) instead of letting the throw escalate to an engine-wide quarantine.
- The guard **still blocks the replay** — the bootstrap transaction rolls back, nothing new is persisted. Only the *failure mode* changes: fail-open (skip one conversation) instead of fail-closed (kill the engine).

## Tests

- `test/bootstrap-flood-regression.test.ts`: a deterministic, offline repro (seed an identical user burst → advance one SQLite second → append-only re-import) reproduces the exact production error (`role=user replicatedRows=3 threshold=3`) on `main`, and asserts the patched behavior degrades instead of throwing. A second test proves the core #639 contract: **a healthy session still bootstraps on the same engine instance** after another conversation trips the guard.
- `test/engine.test.ts`: the existing `keeps the replay flood guard active for user-leading append-only replay batches` test is updated to assert fail-open (resolves, degraded) while keeping its "nothing new persisted" assertion — the guard's *detection* is unchanged.
- Full suite: **1068 passing** (50 files). `pnpm build` ✓.

## Scope

Fixes the **dedup-fail-closed** half of #639 (bootstrap throw → whole-engine quarantine). It does **not** touch the deferred-compaction loop, and is independent of #755 (role-aware thresholds) — they compose: #755 tunes *when* the guard fires, this changes *what happens* when it fires during bootstrap.


rafaelreis-r:
Closing: this fail-open catch is a band-aid, not the root cause. The guard should not run on trusted on-disk bootstrap re-imports at all. Reworking before re-proposing.

PR #809, opened 2026-06-03
fix(transcript-repair): dedupe duplicate assistant tool_use blocks at assemble time

jwavro:
# PR: dedupe duplicate assistant `tool_use` blocks at assemble time (+ same-span `tool_result` fix)

**Repo:** `Martian-Engineering/lossless-claw`
**Base:** built/verified against `4fc60c9` (npm `gitHead` for `@martian-engineering/lossless-claw@0.9.4`)
**Files:** `src/transcript-repair.ts`, `test/transcript-repair.test.ts`
**Companion patch:** `lcm-toolus-dedup.patch` (184 insertions / 11 deletions)

---

## Summary

`sanitizeToolUseResultPairing()` already de-duplicates tool **results** (`seenToolResultIds`) but has no symmetric guard for duplicate assistant **`tool_use`** blocks. When the same assistant `tool_use` id appears on more than one assembled message, the Anthropic Messages API rejects the turn (two `tool_use` blocks sharing an id can't be matched to results). This PR adds a block-level, keep-first dedup of assistant `tool_use` blocks during assembly, so a transcript that already contains duplicate `tool_use` ids still produces a valid outbound payload.

It also fixes a pre-existing correctness bug in the same function: a duplicate tool **result** appearing twice *within a single assistant span* was dropped without setting `changed`, so the function could return the original (unsanitized) array.

Read-path only — no schema, storage, or ingest changes.

## Motivation

Duplicate assistant `tool_use` ids enter the store via duplicate ingestion of the same logical message. `afterTurn()` (engine.ts) runs `reconcileTranscriptTailForAfterTurn()` (which ingests from the session file) and then `ingestBatch()`, with `deduplicateAfterTurnBatch()` in between guarding against replay. That guard matches on `(role, content)` identity, which **collapses for empty-content assistant tool-use turns** (a tool-use-only assistant message stores empty content, so every such turn is identity-equal). The result is two persisted assistant rows carrying the same `tool_use` id.

Fixing ingest idempotency is the real root cause and is proposed as a separate change below. This PR is the assemble-time safety net: it keeps the outbound payload valid regardless of how duplicates were stored, and it protects already-duplicated history.

Observed in production: duplicate `tool_use` ids formed in a live conversation across active turns; with this filter deployed, those turns assembled and sent with zero `invalid_request` rejections while the duplicate rows were still present in the DB.

## Changes

Suggested as two commits:

**1. Assistant `tool_use` block dedup (keep-first, block-level).**
New helper `filterAssistantToolUseBlocks(msg, seenToolUseIds, { dropAll, record })`. During assembly, for each assistant message:
- Drop `tool_use` blocks whose id was already emitted earlier in the assembled transcript (keep-first); preserve all other blocks (text, distinct tool calls).
- If the message has no blocks left after filtering, drop the message entirely.
- Track surviving ids in a transcript-scoped `seenToolUseIds` set.

Error/aborted turns are treated as non-pairable (consistent with the existing early-return): they use `{ dropAll: true, record: false }` — their `tool_use` blocks are stripped (they may be incomplete) and their ids are **not** recorded, so an aborted turn cannot "claim" an id that a later valid turn legitimately reuses. *This is a deliberate behavior change for error/aborted assistant turns that carry non-empty `tool_use` content* (empty ones are already skipped at ingest); previously such blocks passed through. Rationale: an incomplete/aborted `tool_use` with no matched result is itself an invalid outbound block.

**2. Same-span duplicate `tool_result` fix.**
In the span scan, a result whose id is already in `spanResultsById` (seen earlier in the same span, before being added to `seenToolResultIds`) now increments `droppedDuplicateCount` and sets `changed = true`, instead of silently falling through. Without this, a within-span duplicate could be the only change, leaving `changed === false` and causing the function to return the unsanitized input.

Optional `log?: { warn(msg): void }` second parameter emits a single warning when duplicate assistant `tool_use` blocks are dropped. Defaults to no-op; the assembler call site does not pass a logger yet (`ContextAssembler` has no logger in scope), so wiring it is a small optional follow-up.

## Edge cases covered

- Exact duplicate tool-use-only assistant message across two messages → one survives.
- Later message mixes a duplicate id with a distinct tool call / text → only the duplicate block is dropped, the rest preserved.
- Aborted turn carrying id `X` before a later valid turn with id `X` → the valid turn survives with its result; the aborted block does not poison it.
- Duplicate tool results within a single assistant span → reduced to one.

## Tests

`test/transcript-repair.test.ts` extended with the four cases above plus a logger-invocation test. Full suite passes (`vitest run --dir test`): 46 files, 859 tests. Build reproduces deterministically (`npm ci && npm run build`); pristine `4fc60c9` rebuild is byte-identical in size to the published 0.9.4 bundle.

## Note for maintainers

`transcript-repair.ts` documents itself as a temporary copy "to avoid depending on unexported internals … when the plugin SDK exports `sanitizeToolUseResultPairing`, this file can be removed." If the canonical implementation lives in openclaw core, this dedup likely belongs there too; happy to mirror it.

---

## Follow-up (separate issue/PR): ingest idempotency

The durable fix is to stop persisting the same logical message twice. Diagnosis, grounded in `4fc60c9`:

- `afterTurn()` ingests via both `reconcileTranscriptTailForAfterTurn()` and `ingestBatch()` in one pass.
- `deduplicateAfterTurnBatch()` is the intended guard but keys on `(role, content)` identity (`messageIdentity`).
- Tool-use-only assistant turns persist empty content, so `(role, content)` is identical across distinct turns — the guard can't distinguish a replay from a genuine new turn, and duplicate rows are written.

Proposed direction (for discussion, not implemented here): give ingest a stable per-message idempotency key rather than content identity — e.g. a stable source/transcript message id if one is available from the session file, or a composite of `(role, tool_use ids, ordinal/seq)` for tool-use turns — and short-circuit `ingestSingle()` when that key already exists for the conversation. `ingestSingle()` already skips empty error/aborted assistant messages; this would extend that guarding to the replay case. This needs design + tests and should not block the assemble-time fix above.

### DIFF
--- a/.changeset/dedupe-assistant-tool-use.md
+++ b/.changeset/dedupe-assistant-tool-use.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Sanitize duplicate assistant tool-use blocks during context assembly so replayed history cannot produce provider-invalid tool call payloads.
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -14,6 +14,7 @@ import { formatToolOutputReference } from "./large-files.js";
 type AgentMessage = Parameters<ContextEngine["ingest"]>[0]["message"];
 type AssemblySegment = "evictable" | "freshTail";
 type FocusBriefLookup = Pick<FocusBriefStore, "getActiveFocusBrief">;
+type RepairLogger = { warn: (message: string) => void };
 
 export interface AssemblyOverflowContributor {
   /** Context item ordinal in the persisted conversation window. */
@@ -1336,6 +1337,7 @@ export class ContextAssembler {
     private summaryStore: SummaryStore,
     private timezone?: string,
     private focusBriefStore?: FocusBriefLookup,
+    private log?: RepairLogger,
   ) {}
 
   /**
@@ -1561,7 +1563,7 @@ export class ContextAssembler {
     const preSanitizeFreshTailMessages = cleanedEntries
       .filter((entry) => entry.segment === "freshTail")
       .map((entry) => entry.message);
-    const repaired = sanitizeToolUseResultPairing(cleaned) as AgentMessage[];
+    const repaired = sanitizeToolUseResultPairing(cleaned, this.log) as AgentMessage[];
     return {
       messages: repaired,
       estimatedTokens,
--- a/src/assembler.ts
+++ b/src/assembler.ts
@@ -184,6 +184,32 @@ function freshTailProtectionMessageHashes(messages: AgentMessage[]): string[] {
   return hashes;
 }
 
+const FRESH_TAIL_PROTECTION_MARKER = Symbol("freshTailProtection");
+
+type FreshTailProtectionMarkedMessage = AgentMessage & {
+  [FRESH_TAIL_PROTECTION_MARKER]?: true;
+};
+
+function repairedFreshTailProtectionMessageHashes(
+  entries: Array<{ message: AgentMessage; segment: AssemblySegment }>
+): string[] {
+  const markedMessages = entries.map((entry): FreshTailProtectionMarkedMessage => {
+    if (entry.segment !== "freshTail") {
+      return entry.message as FreshTailProtectionMarkedMessage;
+    }
+    return {
+      ...entry.message,
+      [FRESH_TAIL_PROTECTION_MARKER]: true,
+    } as FreshTailProtectionMarkedMessage;
+  });
+  const repaired = sanitizeToolUseResultPairing(
+    markedMessages
+  ) as FreshTailProtectionMarkedMessage[];
+  return repaired
+    .filter((message) => message[FRESH_TAIL_PROTECTION_MARKER])
+    .map((message) => hashMessages([message]));
+}
+
 // ── Public types ─────────────────────────────────────────────────────────────
 
 export interface AssembleContextInput {
@@ -1564,6 +1590,10 @@ export class ContextAssembler {
       .filter((entry) => entry.segment === "freshTail")
       .map((entry) => entry.message);
     const repaired = sanitizeToolUseResultPairing(cleaned, this.log) as AgentMessage[];
+    const protectionHashes = new Set([
+      ...freshTailProtectionMessageHashes(preSanitizeFreshTailMessages),
+      ...repairedFreshTailProtectionMessageHashes(cleanedEntries),
+    ]);
     return {
       messages: repaired,
       estimatedTokens,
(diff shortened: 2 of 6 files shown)


100yenadmin:
Maintainer triage: marking this as a P1 bug-fix candidate. The PR is green/clean at `28da550985c672d0267a63d056002c4d07a40327` with no review threads, and the symptom is provider-request rejection from duplicate assistant `tool_use` ids during assembly. No matching open issue was linked in the PR metadata; the remaining ingest-idempotency root cause should get a separate follow-up ticket rather than being treated as closed by this assemble-time safety net.

100yenadmin:
Maintainer update: linked focused P1 tracking issue #828 for the exact provider-invalid duplicate assistant `tool_use` symptom.

I also pushed hardening in `8716ad6b607ce0d87479f36e3ef843b881c67e64` after adversarial review found merge-readiness blockers:
- snake_case terminal `stop_reason` was not treated as non-pairable;
- terminal or duplicate-only assistant tool-use turns could incorrectly block a delayed real result;
- duplicate filtering could leave reasoning-only assistant content that downstream provider layers strip to an invalid empty assistant;
- terminal removals were logged as duplicate removals, and production call sites did not pass the logger;
- the repo required a patch changeset for this user-facing fix.

Local validation after the hardening commit:
- `npm exec -- vitest run test/transcript-repair.test.ts`
- `npm exec -- vitest run test/engine.test.ts -t "drops orphan tool results|inserts synthetic tool results|preserves non-tool content and matched tool calls|evicts historical tool-call pairs"`
- `npm run build`
- `git diff --check`

Waiting for GitHub checks on `8716ad6` before any 10/10 assignment. I did not merge.

100yenadmin:
Maintainer review: passed / 10/10 mergeability candidate.

Current head: `8716ad6b607ce0d87479f36e3ef843b881c67e64`.
Linked issue: #828.

What was hardened after adversarial review:
- Added a patch changeset for this user-facing provider-validity fix.
- Treat both `stopReason` and `stop_reason` terminal assistant turns as non-pairable, so partial error/aborted tool-use blocks cannot claim IDs.
- Fixed lookahead span handling so terminal or duplicate-only assistant tool-use turns do not block a delayed real tool result.
- Dropped assistant turns that become provider-empty after filtering, including reasoning/blank-only remainder.
- Split duplicate vs terminal tool-use warning counts and wired the existing runtime logger into assembler and volatile live-input sanitizer paths.

Validation:
- `npm exec -- vitest run test/transcript-repair.test.ts`
- `npm exec -- vitest run test/engine.test.ts -t "drops orphan tool results|inserts synthetic tool results|preserves non-tool content and matched tool calls|evicts historical tool-call pairs"`
- `npm run build`
- `git diff --check`
- GitHub CI on `8716ad6`: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed.

Adversarial review: four read-only lanes found the blocker set above; all verified blockers are fixed and covered by regressions. Assigned to @jalehman for merge authority. I did not merge.

PR #812, opened 2026-06-03
fix: show conversation IDs in lcm_grep results

### LINKED ISSUE #811: lcm_grep all-conversation results omit per-hit conversation provenance
# Global `lcm_grep` results omit per-hit conversation provenance

## Summary

`lcm_grep(allConversations=true)` can return message and summary hits from multiple conversations, but the formatted result text does not include the `conversationId` for each hit. The backing retrieval rows already carry that metadata; it is dropped while rendering the tool response.

## Expected

When `lcm_grep` searches globally, each rendered hit should include enough provenance for the agent to distinguish conversations, for example:

```text
- [conv=42 msg#123] (user, 2026-01-01 00:00 UTC): ...
- [conv=99 sum_abc] (leaf, 2026-01-02 00:00 UTC): ...
```

Equivalent formatting is fine as long as each hit exposes its conversation/session scope.

## Actual

The tool header says the search covered all conversations, but individual hits are rendered only with message or summary IDs:

```text
## LCM Grep Results
**Pattern:** `SHARED_SCOPE_MARKER`
**Mode:** regex | **Scope:** both | **Sort:** recency
**Conversation scope:** all conversations
**Total matches:** 2

### Messages

- [msg#1] (user, 2026-01-01 00:00 UTC): SHARED_SCOPE_MARKER from conversation ten

### Summaries

- [sum_20] (leaf, 2026-01-02 00:00 UTC): SHARED_SCOPE_MARKER from conversation twenty
```

In the reproducer, the message row had `conversationId: 10` and the summary row had `conversationId: 20`.

## Impact

This is not default cross-session leakage: callers must explicitly opt into `allConversations=true`.

It is still a correctness/provenance bug. Global search is exactly when source labels matter most. Without per-hit conversation IDs, the agent can:

- answer with the right-looking fact from the wrong conversation;
- merge facts from unrelated conversations;
- fail to narrow follow-up calls with `conversationId`;
- cite or expand ambiguous `msg#...` / `sum_...` hits without knowing their source conversation.

## Reproduction

1. Stub `RetrievalEngine.grep()` to return:
   - a message hit with `conversationId: 10`;
   - a summary hit with `conversationId: 20`.
2. Execute:

```json
{
  "pattern": "SHARED_SCOPE_MARKER",
  "scope": "both",
  "allConversations": true
}
```

3. Inspect the tool text. It contains `[msg#1]` and `[sum_20]`, but not `conv=10` or `conv=20`.

## Suggested Fix

Include `conv=<conversationId>` in each rendered message and summary hit. The metadata is already available on every grep hit, and showing it consistently keeps scoped and global output formats aligned with `lcm_describe`'s `meta conv=...` output.

Verify the existing `lcm_grep` tool coverage still passes and that the focused output path renders `conv=<conversationId>` for both raw message hits and summary hits.

## Validation target

- `npx vitest run test/lcm-tools.test.ts`
- `npm run build`
- `npm test`

## Severity

P2: recall/search can return ambiguous or wrong-source data after an explicit global search. No persisted data loss was observed.

tadad:
## Summary

- Closes #811.
- Include `conv=<conversationId>` in each `lcm_grep` message result label.
- Include `conv=<conversationId>` in each `lcm_grep` summary result label.
- Add a patch changeset for the user-facing recall-tool output change.

## Validation

- `npx vitest run test/lcm-tools.test.ts`
- `npm run build`
- `npm test`

### DIFF
--- a/.changeset/lcm-grep-conversation-provenance.md
+++ b/.changeset/lcm-grep-conversation-provenance.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Include conversation IDs on each `lcm_grep` message and summary result so agents can disambiguate global recall hits.
--- a/src/tools/lcm-grep-tool.ts
+++ b/src/tools/lcm-grep-tool.ts
@@ -262,7 +262,7 @@ export function createLcmGrepTool(input: {
         lines.push("");
         for (const msg of result.messages) {
           const snippet = truncateSnippet(msg.snippet);
-          const line = `- [msg#${msg.messageId}] (${msg.role}, ${formatDisplayTime(msg.createdAt, timezone)}): ${snippet}`;
+          const line = `- [conv=${msg.conversationId} msg#${msg.messageId}] (${msg.role}, ${formatDisplayTime(msg.createdAt, timezone)}): ${snippet}`;
           if (currentChars + line.length > MAX_RESULT_CHARS) {
             lines.push("*(truncated — more results available)*");
             break;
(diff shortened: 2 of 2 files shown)


jalehman:
Thank you!

PR #814, opened 2026-06-03
fix: isolate stable cron session runs

### LINKED ISSUE #810: Isolated OpenClaw cron runs that reuse the same `sessionKey` leak prior runs' transcript into LCM context
### Problem

OpenClaw schedulers like cron often run as **isolated** sessions: each scheduled run is logically a brand-new conversation, but OpenClaw keeps the same `sessionKey` across runs.

These isolated runs still need LCM. A single run can produce a very long single-turn conversation (large tool outputs, deep agent loops, long retrieved documents), and when that turn approaches the context window, **LCM compaction inside the run is essential** — without it the run either truncates badly or fails outright.

But because OpenClaw keeps `sessionKey` stable across runs, LCM (which keys conversations by `sessionKey`) treats every new run as a continuation of the previous one:

- `getConversationBySessionKey(sessionKey)` returns the previous run's conversation row.
- The new run's assembled prefix contains messages and summaries from the previous run.
- Compaction summaries get polluted by unrelated prior runs.
- Output of run N may reference content the user only sent to run N-1.

### Why existing options don't help

| Option | Cross-run isolation | Compaction inside a run | Notes |
|---|---|---|---|
| `ignoreSessionPatterns` | ✅ | ❌ | LCM is fully bypassed — long isolated runs that need compaction can no longer use it. |
| `statelessSessionPatterns` + `skipStatelessSessions=true` | ❌ | ❌ | Reads from existing LCM rows by `sessionKey`, so still pulls prior runs' transcript into the prefix. Writes are skipped, but the *read* leak is the user-visible bug. Also no compaction. |

So today, isolated-cron-style users are forced to choose between **losing compaction** or **leaking memory across logically independent runs**. Neither is acceptable for long-running isolated jobs.

### Expected behaviour

Either lossless-claw correctly recognizes each isolated cron run as a separate session,
or — when `ignoreSessionPatterns` is configured — OpenClaw's built-in compaction is
correctly triggered as a fallback so long isolated runs can still compact their context.

### Environment

- lossless-claw version: `0.9.1`
- OpenClaw version: `2026.4.21`

jalehman:
## What
Isolates OpenClaw cron scheduler runs that reuse stable session keys so a new runtime session does not inherit prior-run LCM transcript context. Prior run conversations are preserved and archived rather than deleted.

## Why
Cron-style isolated runs can reuse the same sessionKey across logically independent executions. Without a separate runtime-run boundary, LCM can assemble stale prior-run context into a new run.

## Changes
- Detect stable cron key reuse
- Archive prior active run
- Create fresh run conversation
- Cover lifecycle paths
- Update docs and changeset

## Testing
- `npx vitest run test/engine.test.ts -t "isolates cron runs"`
- `npx vitest run test/engine.test.ts -t "missed reset|ambiguous runtime rollover|isolates cron runs"`
- `npm run build`
- `npm test`
- `git diff --check`

Closes #810

### DIFF
--- a/.changeset/isolate-stable-cron-runs.md
+++ b/.changeset/isolate-stable-cron-runs.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Isolate cron scheduler runs that reuse a stable session key so prior run transcripts do not enter the new run's LCM context.
--- a/README.md
+++ b/README.md
@@ -314,6 +314,8 @@ Lossless-claw applies `/new` pruning through `before_reset` and uses `session_en
 
 Use `ignoreSessionPatterns` or `LCM_IGNORE_SESSION_PATTERNS` to keep low-value sessions completely out of LCM. Matching sessions do not create conversations, do not store messages, and do not participate in compaction or delegated expansion grants.
 
+Cron scheduler keys (`agent:<agent>:cron:<job>...`) are isolated automatically when OpenClaw reuses the same `sessionKey` for a new runtime `sessionId`: lossless-claw archives the prior active run and creates a fresh LCM conversation for the new run. Leave cron sessions out of `ignoreSessionPatterns` when they need in-run LCM compaction.
+
 Pattern rules:
 
 - `*` matches any characters except `:`
@@ -322,7 +324,7 @@ Pattern rules:
 
 Examples:
 
-- `agent:*:cron:**` excludes cron sessions for any agent, including isolated run sessions like `agent:main:cron:daily-digest:run:run-123`
+- `agent:*:cron:**` excludes cron sessions for any agent when you want to bypass LCM entirely
 - `agent:main:subagent:**` excludes all main-agent subagent sessions
 - `agent:ops:**` excludes every session under the `ops` agent id
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -308,6 +308,8 @@ LCM_EXPANSION_MODEL=openai/gpt-5.4-mini
 - `*` matches any characters except `:`
 - `**` matches anything, including `:`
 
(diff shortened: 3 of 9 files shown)


100yenadmin:
Maintainer triage: promoted to the P1 bug lane for #810. Current live gate checked at `cf4859a292293163d644e53cd170558f56951509`: non-draft, `MERGEABLE` / `CLEAN`, CI green, and no review threads. Review note: the patch directly covers bootstrap/assemble isolation for cron `sessionKey` reuse; a direct afterTurn rollover regression would further harden the lifecycle surface, but I did not find a current merge blocker.

100yenadmin:
Maintainer review: passed / 10/10 mergeability candidate.

Hardening pushed in `2b00fb67b24218e2cf9c504e03ae3fe551d87f66`.

What changed:
- Fixed the direct `afterTurn` cron rollover path so it archives the prior run without pre-creating an empty replacement row; the new run transcript now seeds through the existing initial transcript import path.
- Scoped default LCM tool retrieval for isolated cron `sessionKey`s to the active conversation only, preventing `lcm_grep` / shared tool scope from reading archived prior runs in the same session-key family.
- Added regressions for long cron key suffixes, prior input/output non-leakage, direct `afterTurn` rollover, and active-only cron tool scope.

Validation:
- `npm exec -- vitest run test/engine.test.ts -t "isolates cron|ambiguous runtime rollover|afterTurn archives a stale active conversation|afterTurn skips assistant-only rollover"`
- `npm exec -- vitest run test/lcm-tools.test.ts`
- `npm run build`
- `git diff --check`
- GitHub CI on `2b00fb6`: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed.

Adversarial review:
- Four read-only lanes found two P1 blockers at the previous head: direct `afterTurn` could drop the new run transcript, and default LCM tool scope could still read archived prior cron runs.
- Both blockers were fixed and delta-rereviewed clean by the same lanes.

Assigned to @jalehman for merge authority. I did not merge.

Issue #807, opened 2026-06-02
Feature: prepare incremental summaries in the background and publish atomically


jalehman:
## Summary

Explore a hybrid compaction model that brings back incremental compaction as background preparation while preserving the cache-stable full-sweep publish behavior.

Instead of doing a serial full sweep on the turn-critical path, Lossless would incrementally summarize older raw-message chunks in the background. Those summaries would be stored as pending/non-canonical records and ignored by `assemble()` until the normal compaction threshold is reached. At threshold time, Lossless validates the prepared summaries and atomically promotes a complete batch into active context, causing the prompt prefix to change once.

This keeps the useful part of the old incremental design, cheap small summarizer calls with previous-summary continuity, without reintroducing cache-hostile incremental context mutation.

Related: #746 is about making deferred compaction drain background-only. This issue is different: it proposes precomputing the actual replacement summaries ahead of the threshold and publishing them atomically later.

## Desired shape

- Background worker prepares leaf summaries for compactable old raw-message chunks.
- Prepared summaries are persisted as pending/non-canonical and are invisible to normal assembly.
- Pending leaf generation can still use the existing continuity model: each chunk may receive the previous 1-2 pending summaries as context.
- When the full-sweep threshold is reached, foreground compaction should mostly be a validation + publish transaction rather than a long serial chain of LLM calls.
- Publish should change the active context epoch once, so prompt cache behavior remains close to the current full-sweep model.
- If pending work is stale or incomplete, Lossless should degrade safely to current full-sweep behavior or leave existing active context untouched.

## Why this matters

Current threshold full sweeps preserve prompt-cache stability better than the original incremental design, but they can take minutes because summary calls are serial: summary N receives prior summaries from N-1/N-2 for continuity.

The proposed model moves most of that serial chain off the active path:

```text
normal turns
  -> background prepares pending summaries over time
  -> assemble() still sees the old active context

threshold reached
  -> validate prepared batch
  -> atomically replace old raw-message range with prepared summaries
  -> next assemble sees the new summary prefix
```

The user-facing win is near-instant threshold compaction without sacrificing the one-prefix-change cache behavior.

## Safety requirements

This should not be implemented as just a `pending` boolean and a flip. The implementation needs explicit lifecycle and generation semantics so pending summaries are never confused with canonical memory.

Likely requirements:

- Add a batch/generation model, probably a `compaction_batches` table.
[...]

PR #821, opened 2026-06-03
Add background prepared compaction batches

tadad:
## Summary
- add durable prepared compaction batches and pending summary lifecycle for background leaf preparation
- keep pending summaries invisible until threshold compaction validates and atomically publishes a ready batch
- mark thrown background preparation failures as failed so durable batches do not stay pending forever
- add config/docs/schema updates and changesets for the user-facing background prep behavior

## Tests
- ./node_modules/.bin/vitest run test/prepared-compaction.test.ts
- npm test
- npm run build
- git diff --check

Refs #807


100yenadmin:
Maintainer triage: labeling this as `enhancement` / `priority:P3` / `linked-pr` for #807.

This is a draft, conflicting feature PR for background prepared compaction batches, so I am keeping it out of the current P0/P1 bug merge lane. Next useful maintainer action is author-side rebase and undraft when the feature stack is ready for review; no deep bug-lane review performed here.

jalehman:
Thanks for the thoughtful prototype for #807. I’m going to close this draft because the pending-summary design has moved to a different architecture: the current path depends on the host-owned SQLite transcript projection and a planned pending-summary DAG with separate planner/worker/publisher boundaries. This PR helped validate the hidden prepared-summary idea, but rebasing it would require replacing the core approach rather than making review-sized fixes. The best follow-up path is a narrow contribution against the current pending-summary branch or against `main` after that work lands.

PR #621, opened 2026-05-06
fix(afterTurn): evaluate compaction even when ingestBatch is empty

100yenadmin:
## TL;DR
**Bug:** When `afterTurnTranscriptReconcile` (or any host path that ingests messages before `afterTurn`'s dedup runs) imports the new turn's content, `deduplicateAfterTurnBatch` correctly identifies the batch as already-stored (via `tail-match`) and returns `[]`. The empty `ingestBatch` then triggers an early-return at `engine.ts:6294-6300` **before** the compaction-evaluation block at `engine.ts:6463` ever runs. Long-running conversations accumulate well past `contextThreshold` without any incremental leaf pass, deferred-debt record, or budget-trigger run — leaving only the host's emergency overflow truncation as a safety net.

**Fix:** Replace the early-return with a fall-through. Skip the `ingestBatch` call when nothing to ingest, but continue to compaction evaluation. ~30 lines of source change. Default behavior unchanged for the non-empty case.

## The bug, visualized

```
                     ┌────────────────────────────────────────┐
                     │  HOST CALLS afterTurn(messages)        │
                     └────────────────────────────────────────┘
                                     │
                                     ▼
                ┌────────────────────────────────────────┐
                │  afterTurnTranscriptReconcile          │
                │  (imports new turn from session jsonl) │  ← already stored ✓
                └────────────────────────────────────────┘
                                     │
                                     ▼
                ┌────────────────────────────────────────┐
                │  deduplicateAfterTurnBatch(newMessages)│
                │  → tail-match detected → returns []    │  ← correct dedup ✓
                └────────────────────────────────────────┘
                                     │
                                     ▼
                       ingestBatch.length === 0
                                     │
              ┌──────────────────────┴──────────────────────┐
              │                                              │
       BEFORE THIS PR                                  AFTER THIS PR
              │                                              │
              ▼                                              ▼
   ┌────────────────────┐                          ┌──────────────────────┐
   │ log + return       │                          │ log + skip ingest    │
   │ (compaction never  │  ✗ BUG                   │ + fall through       │
   │  evaluates)        │                          └──────────────────────┘
   └────────────────────┘                                    │
                                                             ▼
                                                  ┌──────────────────────┐
                                                  │ tokenBudget resolution│
                                                  │ conversation lookup   │
                                                  │ telemetry update      │
                                                  │ evaluateIncremental-  │
                                                  │   Compaction          │
                                                  │ → budget-trigger fires│
                                                  │ → deferred drain runs │
[...]

PR #823, opened 2026-06-04
fix(compaction): treat terminal compaction exhaustion as non-retryable (#639)

rafaelreis-r:
### What

When a threshold sweep takes no action and does not fail while still over target, that is **terminal exhaustion**, not a retryable failure: `thresholdPressureTokens = max(currentTokens, observedTokens, projectedTokens)` and `observedTokens` (the host's live count) is irreducible by compaction — compaction shrinks stored leaves, not the observed floor. When the floor alone exceeds target, every sweep ends still-over-target, the deferred-debt drain keeps the debt `pending`, and `retry_attempts` climbs forever, thrashing the assemble degraded-fallback.

This implements @Grynn's exhaustion-handling proposal from the #639 thread (credit to @Grynn).

### How (surgical)

- `executeCompactionCore` flags `exhausted` when `isThresholdSweep && !actionTaken && !authFailure && still-over-target`.
- **`ok` stays `false`** so overflow recovery / #15's honest "still over target" signal is unchanged.
- Only the deferred-debt drain (`consumeDeferredCompactionDebt`) honors `exhausted`: it clears the debt (`keepPending=false`, `failureSummary=null`) instead of pinning `pending` + climbing `retry_attempts`.

### Repro / validation

- Deterministic unit test in `test/engine.test.ts`: seed over-target threshold debt with no compactable candidates, drain. **Before:** `pending=true retry=1 "live context still exceeds target"`. **After:** `pending=false retry=0 null`.
- Full suite green.
- Matches production: a live session sat `pending=1 retry_attempts=7 "live context still exceeds target" current=247979 budget=150000`; after deploy the debt cleared (`pending=0`) on the next drain.

### Notes

- Answers @100yenadmin's needs-repro on #639 (reproduces on current `main`).
- Lands in the #639 "debt won't drain" family; relevant to #553 (P1). Complementary to the merged trigger-side work (#621/#15/#351/#681/#772) which makes compaction *fire*; this makes a *terminal over-target* state *stop retrying*.
- Distinct from the placeholder-recovery bug (separate issue/PR).

Closes part of #639 (the deferred-compaction loop). Refs #553.


100yenadmin:
Maintainer review: passed — 10/10 mergeability candidate for the #639 terminal-exhaustion slice.

**Summary**
Four-lane adversarial sweep completed on the original head. Three lanes were clean; the lifecycle lane found one P2 blocker where no-action sweep deadline/iteration-budget stops could be misclassified as terminal exhaustion. I patched that before assignment.

**Fix added**
- New head: `1b2c8c329ac3d4aa4acfa77906a4cc956b85f83e`
- Commit: `1b2c8c3 fix(compaction): keep budget-stopped debt retryable`
- `CompactionResult` now carries `stoppedAtBudget` out of `compactFullSweep`.
- The #639 exhaustion gate excludes budget/deadline-stopped no-action sweeps.
- Added regression coverage proving budget-stopped no-action threshold debt stays pending/retryable, while true no-candidate exhaustion still clears debt.
- Removed the debug console output from the new #639 regression and tightened the retry assertion to exactly 0.

**Validation**
- Local focused Vitest: `npm exec -- vitest run --dir test test/engine.test.ts -t 'no-action sweep stops at budget|over-target threshold debt with no compactable candidates is exhaustion'`
- Local typecheck: `npm run typecheck`
- Local whitespace: `git diff --check`
- GitHub exact-head CI: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed on `1b2c8c329ac3d4aa4acfa77906a4cc956b85f83e`.
- Review threads: none.
- Merge state: `MERGEABLE` / `CLEAN`.

Assigning @jalehman for maintainer merge consideration. This should close the terminal deferred-debt exhaustion slice after merge, but should not close the whole #639 umbrella by itself.

Issue #822, opened 2026-06-04
Placeholder-checkpoint recovery never re-imports a large pre-existing transcript when the DB frontier has no anchor (LCM stays a no-op for the session) — regression of #685


rafaelreis-r:
### Summary

A conversation whose `conversation_bootstrap_state` is a **placeholder** (all-zero: `last_seen_size=0, last_processed_offset=0, last_processed_entry_hash=NULL`) **and** whose DB frontier holds only **non-anchoring** rows (e.g. injected "Conversation info (untrusted metadata)" preambles, or any rows that don't appear in the transcript) can **never re-import its real transcript**. The placeholder-checkpoint-recovery branch added around #685 calls `reconcileSessionTail` **without** `allowNoAnchorImport`, so with no anchor it imports **0 messages — forever** — and even if that flag were passed, the no-anchor import cap (`max(floor(existingDbCount*0.2), 50)`, from #649) blocks any transcript larger than ~50 messages.

Net effect: `assemble()` falls back to the raw transcript (`safeFallback()`), compaction never runs, and **LCM is silently a no-op for that session** — the model receives the full uncompacted transcript until it overflows.

### Why this is a regression of #685

#685 ("seed placeholder bootstrap_state when afterTurn stat-fail fallback runs") created the all-zero placeholder and **claimed** recovery: *"on a successful stat the append-only fast path reads from offset 0 and re-walks the full transcript, with identity_hash dedup preventing re-ingestion."* That re-walk goes through the placeholder-checkpoint-recovery branch — but on `main` that branch passes only `{sessionId, sessionKey, conversationId, historicalMessages, noAnchorImportReason: "placeholder-checkpoint-recovery"}` to `reconcileSessionTail`, **omitting `allowNoAnchorImport`**. So for the no-anchor shape, the re-walk imports nothing. #685 created the placeholder but left its recovery a no-op for this case. (#714 later edited the same branch without addressing it.)

This is also the case that #686 was optimistically triaged as "possibly-resolved-by-#685" (2026-05-30) — it is **not** resolved for the no-anchor frontier. (@gorkem2020's 2026-05-23 report on #686 — `existingCount=7, historicalMessages=168, importedMessages=0, 161 messages never imported; archive+restart not a usable recovery` — is the same 0-import family, via the fast-path overlap detector rather than the placeholder branch.)

### Repro

[...]

PR #776, opened 2026-05-31
fix: guard transcript reconciliation against replay floods

100yenadmin:
## Summary
- Adds occurrence-aware transcript identity analysis for reconciliation imports.
- Makes append-only bootstrap/afterTurn fall back to full reconciliation when the suffix overlaps already-persisted transcript history.
- Blocks all-duplicate no-anchor transcript replays while preserving mixed same-path-shrink epochs by dropping persisted prefixes and importing only the new suffix.
- Adds regressions for stale append-only overlap, all-duplicate no-anchor replay blocking, and mixed same-path-shrink prefix+tail recovery.

Refs #489.

## Validation
- `npm test -- test/engine.test.ts -t "bounded no-anchor imports|same-path shrink keeps|same-path shrink|append-only suffix overlaps|duplicate frontier"`
- `npm test` (49 files, 1009 tests)
- `npm run build`
- `git diff --check`
- `npm pack --dry-run`

## Notes
- Not marking #489 closed yet; this is the prevention lane for duplicate transcript replay floods, and reporter/maintainer confirmation can decide whether the doctor diagnostic follow-up remains necessary.

Issue #686, opened 2026-05-15
Discussion: should reconcileSessionTail stat-fail branch return hasOverlap:false to match #649's no-proof-no-advance principle?


jetd1:
## Context

PR `#649` (transcript epoch rollover, merged 2026-05-11) added a permissive stat-fail fallback in `reconcileTranscriptTailForAfterTurn`'s slow path that returns `hasOverlap:true` when `stat(sessionFile)` fails, allowing the outer `afterTurn` to proceed with `ingestBatch` even though nothing was actually reconciled against the on-disk transcript:

```ts
if (!sessionFileState) {
  // ... warn ...
  return {
    importedMessages: 0,
    blockedByImportCap: false,
    hasOverlap: true,  // ← permissive: "I didn't read the transcript, but go ahead and ingest anyway"
  };
}
```

PR `#685` (this repo) just closed the most acute symptom of that decision: if the same `stat()` keeps failing, `refreshBootstrapState` also fails, the catch swallows the error, and the conversation gets stuck in a transparent-passthrough state with no compaction. `#685` seeds a placeholder `bootstrap_state` row in the stat-fail branch so the next turn can fast-path-recover. It deliberately does **not** revisit the original "permissive" policy.

This issue is to discuss whether that policy is the right one in the first place.

## The inconsistency

`#649` has a clear central principle stated in code comments:

> *"A no-overlap/no-import result leaves the checkpoint stale on purpose so future turns can retry."* (`engine.ts:5286`-ish)

> *"the transcriptReconcileUnsafeToAdvance guard skips afterTurn persistence to avoid creating a future anchor past unreconciled transcript history"* (`engine.ts:6510`-ish)

These translate to: **"if reconcile cannot prove it covered the on-disk transcript frontier, do not let `afterTurn` push more state forward."**

The stat-fail branch is the one place where that principle is silently violated. By returning `hasOverlap:true`, it tells the outer afterTurn "the frontier is covered, you may proceed" — but `historicalMessages.length === 0` and `sessionFileState === undefined`, so absolutely nothing was reconciled.

The justification is presumably "stat-fail is transient (e.g. a write-race on the host runtime's session JSONL), and skipping ingest would lose this turn." But:

1. If stat-fail is *truly* transient, the next turn's stat will succeed and the append-only fast path will read this turn's content from the transcript anyway — so skipping this turn loses nothing.
[...]

PR #824, opened 2026-06-04
fix(bootstrap): recover placeholder-checkpoint conversations with a non-anchoring frontier (#639 family)

rafaelreis-r:
### What

Fixes `#822`: a conversation with a placeholder `bootstrap_state` (all-zero, from #685) plus a DB frontier of only non-anchoring rows could never re-import its real transcript — the placeholder-checkpoint-recovery branch called `reconcileSessionTail` **without** `allowNoAnchorImport` (0 imports forever), and the no-anchor import cap (`max(floor(existingDbCount*0.2), 50)`, #649) blocked transcripts > ~50 messages; this PR now raises that recovery cap while keeping it finite. Result: `assemble()` falls back to the raw transcript, compaction never runs, LCM is silently a no-op for that session.

### How (2 hunks, minimal)

1. Pass `allowNoAnchorImport: true` in the placeholder-checkpoint-recovery reconcile — a placeholder checkpoint means never-ingested, so importing the transcript as a bounded new epoch is the correct recovery.
2. Use a larger finite import cap only for `noAnchorImportReason === "placeholder-checkpoint-recovery"` (initial-epoch catch-up) so large recoveries can proceed without disabling the replay-flood guard.

Safety:
- The replay-duplicate guard runs **before** the cap and is unchanged → #776's anti-flood protection is preserved; genuine duplicate floods still abort.
- Only conversations with this exact pathological shape (placeholder checkpoint + non-anchoring frontier) change behavior — and for those, prior behavior was a permanent no-op (0 imports forever), so the new behavior is strictly an improvement.
- Idempotent: after import, `refreshBootstrapState` advances the checkpoint to EOF (non-placeholder) → subsequent turns take the normal incremental append-only path. The recovery runs once.

### Repro / validation

- Deterministic unit test in `test/engine.test.ts`: seed a conversation with 2 non-anchoring metadata rows + a placeholder checkpoint + a 120-message transcript; call `afterTurn`. **Before:** only the 2 metadata rows exist (`expected [...(2)] to include 'real user turn 0'` fails). **After:** the full transcript imports once, checkpoint advances. Full suite green.
- Production: a real session stuck at 2 DB msgs / 0 summaries / placeholder checkpoint while its session file held ~764 messages. After the fix + one turn: 1999 messages, 19 summaries, checkpoint advanced, compaction healthy, no more overflow.

### Related

Regression of #685 (created the placeholder, left recovery a no-op for the no-anchor shape) · #714 edited the same branch without fixing it · refutes "possibly-resolved-by-#685" triage on #686 · cap origin #649, anti-flood #489/#776 (preserved) · part of the #639 "context lost" family but a distinct mechanism from the Mode-2 deferred-compaction exhaustion (separate PR).

Fixes `#822`. Refs #685, #686, #639.

### DIFF
--- a/.changeset/placeholder-checkpoint-recovery.md
+++ b/.changeset/placeholder-checkpoint-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Recover placeholder-checkpoint conversations whose persisted metadata frontier cannot anchor into the real transcript.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -6120,8 +6120,7 @@ export class LcmContextEngine implements ContextEngine {
           const replayThreshold = Math.max(3, Math.ceil(historicalMessages.length * 0.5));
           const shouldDropPersistedPrefix =
             replayAnalysis.firstNonOverlappingIndex > 0 &&
-            (persistedIdentityOverlaps >= replayThreshold ||
-              params.noAnchorImportReason === "placeholder-checkpoint-recovery");
+            persistedIdentityOverlaps >= replayThreshold;
           if (
             replayAnalysis.firstNonOverlappingIndex < 0 &&
             persistedIdentityOverlaps >= replayThreshold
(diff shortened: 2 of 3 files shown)


100yenadmin:
Maintainer review: passed — 10/10 mergeability candidate for the placeholder-checkpoint / non-anchoring frontier recovery slice.

**Summary**
Four-lane adversarial sweep completed. Correctness and public-surface lanes were clean. Test/lifecycle lanes found two merge-readiness gaps before assignment, and both were patched.

**Fixes added**
- New head: `c1b8228221a747312ed9fe7304ec042d8c3aa6a5`
- Commit: `c1b8228 test(bootstrap): harden placeholder recovery retry safety`
- Strengthened the large no-anchor placeholder recovery regression to assert exact total count, representative head/tail identity counts, exact checkpoint EOF offset, and second-pass idempotence.
- Added a mid-import failure regression proving placeholder recovery leaves the all-zero checkpoint retryable when transcript import throws partway through.
- Fixed `afterTurn` reconcile error handling so a thrown transcript reconcile is treated as unsafe/no-overlap instead of preserving the default safe-overlap result and refreshing the checkpoint past missing history.

**Validation**
- Local focused Vitest: `npm exec -- vitest run --dir test test/engine.test.ts -t 'placeholder-checkpoint conversation whose only DB rows are non-anchoring metadata|does not advance a placeholder checkpoint when no-anchor recovery throws mid-import'`
- Local typecheck: `npm run typecheck`
- Local whitespace: `git diff --check`
- GitHub exact-head CI: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed on `c1b8228221a747312ed9fe7304ec042d8c3aa6a5`.
- Review threads: none.
- Merge state: `MERGEABLE` / `CLEAN`.

Assigning @jalehman for maintainer merge consideration. This should close #822 after merge and is a distinct #639-family fix; it should not close the whole #639 umbrella by itself.

jalehman:
Closing because this implementation is unsafe on current main: it broadens placeholder-checkpoint recovery into no-anchor import for cases where the DB may already contain real rows, which can contaminate a conversation with an unrelated transcript. The underlying #822-class bug is still valid and needs a narrower follow-up fix.

PR #826, opened 2026-06-04
fix: deduplicate replayed ingest batches

### LINKED ISSUE #561: [hotfix][bootstrap] v0.9.3 leaves #489 partially open: raw-hash bypass + dedup gap
# Hotfix candidate — v0.9.3 leaves #489 partially open

**This is a hotfix candidate.** `#489` should remain open, blocked by this. Both sub-items are leaks in the v0.9.3 bootstrap fix (`#510`) that reintroduce the multi-million-token re-import / monotonic-conversation-growth path `#489` was meant to close.

## Why this is one issue, not two

Both sub-items concern the same bootstrap/reconcile import path that `#510` rewrote. Sub-item 1 leaks the *anchor* (raw-hash bypass on heartbeat pruning). Sub-item 2 leaks the *dedup* (no uniqueness on identity hash even with anchor present). The two leaks compose: if the anchor misses, dedup must catch; if dedup is absent, the anchor is the only line of defense. The fix shape — a single PR hardening `ingestSingle`/`refreshBootstrapState` on the bootstrap path — addresses both.

### Sub-item 1: PR #510 silently bypasses its own raw-hash anchor when heartbeat pruning fires on first bootstrap

**Repro hypothesis:** `lastImportedHash` is set only when `prunedMessages === 0`. When `pruneHeartbeatOk` strips any tail, the raw-hash anchor is `undefined` and `refreshBootstrapState` falls back to hashing externalized DB content, which won't match the JSONL on the next bootstrap. This re-introduces the `#489` multi-million-token re-import path that `#510` was meant to close.

**Suggested fix:** always pass the raw-transcript hash when ≥1 message imported, regardless of heartbeat pruning.

### Sub-item 2: bootstrap/reconcile still re-imports duplicates as new seqs even after #510

**Repro hypothesis:** `#510` routes bootstrap through `ingestSingle` so re-imported messages are externalized, but the underlying `#489` root cause (anchor drift causes already-stored messages to be re-inserted with fresh seqs) is unchanged. A 788-message re-import still inserts 788 new rows — they are now references rather than 60k-token blobs, but the conversation grows monotonically.

**Suggested fix:** add a `(conversation_id, identity_hash) WHERE identity_hash IS NOT NULL` partial unique index OR check-before-insert in `ingestSingle` when called from the bootstrap path; add a doctor diagnostic for repeated `(conversation_id, externalizedFileId)` references.

## Suggested PR scope

A single hotfix PR that (a) always sets `lastImportedHash` to the raw-transcript hash when ≥1 message was imported (drop the `prunedMessages === 0` gate), and (b) introduces a partial unique index on `(conversation_id, identity_hash)` plus a check-before-insert in the bootstrap-path `ingestSingle` call. The doctor diagnostic for repeated `(conversation_id, externalizedFileId)` references can land in the same PR or a follow-up. `#489` should remain open until both leaks are closed.

## Cross-links

- Related existing issues: #489 (should remain open, blocked by this), #492
- v0.9.3 PRs touching this area: #510

100yenadmin:
## Summary

- add a raw transcript-id replay filter at the `ingestBatch` boundary
- drop only messages whose raw event/tool id and stored identity already exist in the same conversation
- preserve content-only repeated rows, auto-compaction summaries, and mixed replay+new batches
- add regressions for all-duplicate replay, single-message raw-id replay, mixed replay prefix plus new tail, legitimate repeated text, and afterTurn summary persistence

## Issue linkage

Addresses #561. I am intentionally not marking this as `Closes #561` yet because the issue should stay open until this PR is merged and the reporter/maintainer has a chance to confirm the exact checkpoint-loss production trace no longer reproduces.

## Validation

From `/Volumes/LEXAR/repos/lossless-claw-issue-561-dedup`:

```bash
npx vitest run test/engine.test.ts -t "deduplicates persisted replay rows in ingestBatch|keeps content-only repeated rows in ingestBatch|deduplicates single raw-id replay rows in ingestBatch|deduplicates raw-id replay prefix while keeping new ingestBatch tail|afterTurn keeps auto-compaction summary that matches stored text"
npx vitest run test/engine.test.ts -t "ingestBatch|afterTurn dedup guard"
npm run typecheck
npm run build
git diff --check
```

All passed locally. Full-suite validation intentionally left to GitHub CI per local-resource policy.

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -6316,6 +6316,95 @@ export class LcmContextEngine implements ContextEngine {
     return { candidateRawIds: candidateRawIds.size, matchedRawIds };
   }
 
+  /** Drop exact raw-id transcript replays while preserving content-only repeated user turns. */
+  private async filterPersistedRawIdReplayBatch(params: {
+    conversationId: number;
+    sessionId: string;
+    sessionKey?: string;
+    messages: AgentMessage[];
+  }): Promise<AgentMessage[]> {
+    const matchStmt = this.db.prepare(
+      `SELECT 1 AS found
+       FROM message_parts mp
+       JOIN messages m ON m.message_id = mp.message_id
+       WHERE m.conversation_id = ?
+         AND m.role = ?
+         AND m.identity_hash = ?
+         AND mp.metadata IS NOT NULL
+         AND json_valid(mp.metadata)
+         AND (
+           json_extract(mp.metadata, '$.raw.id') = ?
+           OR json_extract(mp.metadata, '$.raw.call_id') = ?
+           OR json_extract(mp.metadata, '$.raw.toolCallId') = ?
+           OR json_extract(mp.metadata, '$.raw.tool_call_id') = ?
+           OR json_extract(mp.metadata, '$.raw.toolUseId') = ?
+           OR json_extract(mp.metadata, '$.raw.tool_use_id') = ?
+         )
+       LIMIT 1`,
+    );
+
+    const filtered: AgentMessage[] = [];
+    let replayedMessages = 0;
+
+    for (const message of params.messages) {
+      const stored = toStoredMessage(message);
+      const rawIds = new Set<string>();
+      for (const part of buildMessageParts({
+        sessionId: params.sessionId,
+        message,
+        fallbackContent: stored.content,
+      })) {
+        for (const rawId of extractRawIdsFromPartMetadata(part.metadata)) {
+          rawIds.add(rawId);
+        }
+      }
+
+      if (rawIds.size === 0) {
+        filtered.push(message);
+        continue;
+      }
+
+      const identityHash = buildMessageIdentityHash(stored.role, stored.content);
+      let alreadyPersisted = false;
+      for (const rawId of rawIds) {
+        const row = matchStmt.get(
+          params.conversationId,
+          stored.role,
+          identityHash,
+          rawId,
+          rawId,
+          rawId,
+          rawId,
+          rawId,
+          rawId,
+        ) as { found: number } | undefined;
+        if (row?.found === 1) {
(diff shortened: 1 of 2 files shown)


100yenadmin:
Maintainer review: 10/10 mergeability score for this focused P1 fix.

Current head: `f77fa4b53abe28ec20976633d431ea9bfb5669d8`

Why I am comfortable assigning this:
- PR is non-draft and `MERGEABLE` / `CLEAN`.
- GitHub CI is green on the current head: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed.
- Cross-repo maintainer edits are enabled (`maintainerCanModify: true`).
- Adversarial review found real blockers in the first implementation: content-only repeated rows could be dropped, direct mixed replay+new was under-tested, single-message replay was uncovered, and afterTurn summaries could be affected by the inner dedup. The final head addresses those by switching from content-based replay trimming to raw-id + identity replay proof, and adds regressions for each blocker class.

Local validation from `/Volumes/LEXAR/repos/lossless-claw-issue-561-dedup`:

```bash
npx vitest run test/engine.test.ts -t "deduplicates persisted replay rows in ingestBatch|keeps content-only repeated rows in ingestBatch|deduplicates single raw-id replay rows in ingestBatch|deduplicates raw-id replay prefix while keeping new ingestBatch tail|afterTurn keeps auto-compaction summary that matches stored text"
npx vitest run test/engine.test.ts -t "ingestBatch|afterTurn dedup guard"
npm run typecheck
npm run build
git diff --check
```

Assigned to @jalehman for merge consideration. This should not automatically close #561 until the PR is merged and the exact checkpoint-loss / duplicate `identity_hash` production symptom is confirmed fixed.

jalehman:

PR #830, opened 2026-06-04
fix(engine): normalize sensitive field names in dedup identity comparison

mpz4life:
## Summary

- Fix duplicate tool-call replay ingestion when OpenClaw stores a host-redacted form of sensitive message content.
- Port the fix to the current dedup owner in `src/batch-dedup.ts`.
- Keep OpenClaw optional by loading `openclaw/plugin-sdk/logging-core` dynamically when available.
- Require exact, non-empty tool-call ID agreement plus an adjacent exact/decorated anchor before accepting redaction equivalence.
- Add focused regression coverage and a patch changeset.

## Why

OpenClaw host hooks can redact sensitive field names before a message is stored. Lossless Claw previously compared the raw stored text during after-turn batch deduplication, so a redacted replay could appear new and be ingested again. The duplicate tool-call rows could later produce synthetic `[lossless-claw] missing tool result` repairs.

## Compatibility

No OpenClaw compatibility floor changes. If the optional SDK redactor cannot be resolved or throws, deduplication falls back to exact matching.

## Testing

- `npm test -- test/batch-dedup.test.ts test/f1-degraded-double-write.test.ts test/whitespace-divergence-double-write.test.ts`
- `npm run typecheck`
- Hosted CI: test, plugin-inspector, smoke-latest-openclaw

### DIFF
--- a/.changeset/stable-dedup-sensitive-content.md
+++ b/.changeset/stable-dedup-sensitive-content.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Prevent duplicate tool-call replay ingestion when OpenClaw stores a host-redacted form of sensitive message content.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -6,6 +6,7 @@
  *
  * Extracted from engine.ts (Phase 2 of the engine decomposition).
  */
+import { createRequire } from "node:module";
 import {
   formatFileReference,
   formatRawPayloadReference,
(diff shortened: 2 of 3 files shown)


mpz4life:
## Approach: Use SDK `redactSensitiveText` + bump openclaw version

The normalization relies on `import { redactSensitiveText } from "openclaw/plugin-sdk/logging-core"` because:

- Sensitive field-name patterns are **configurable** on the openclaw host side — the `beforeMessageWrite` hook can use custom patterns beyond the built-in set.
- Hardcoding a fixed regex in lossless-claw would drift from the host actual redaction behavior, producing false mismatches or false matches during dedup.
- By delegating to `redactSensitiveText`, the normalization always uses the same patterns the host configured, regardless of what those patterns are.

### Version bump rationale

`openclaw/plugin-sdk/logging-core` is only exported from `2026.6.0+` (verified locally on `2026.6.1`). CI with `2026.5.22` cannot resolve the import, causing:

```
Error: src/engine.ts(90,37): error TS2307: Cannot find module 'openclaw/plugin-sdk/logging-core'
```

The second commit bumps all compat entries from `2026.5.22` → `2026.6.0`:
- `peerDependencies.openclaw`
- `openclaw.compat.pluginApi`
- `openclaw.compat.minGatewayVersion`
- `openclaw.compat.tested`
- `openclaw.build.openclawVersion`

@jalehman please confirm this version bump is acceptable.

mpz4life:
@jalehman The CI typecheck failure is because `openclaw/plugin-sdk/logging-core` is not exported in `2026.5.22`. The PR currently bumps the minimum to `2026.6.0` (verified working on `2026.6.1`).

Two options:

- **Accept the bump** — merge as-is, lossless-claw requires openclaw >= 2026.6.0
- **Reject the bump** — we will implement `normalizeForDedup` locally using the patterns already in `engine.ts` (`PROMPT_RECALL_SENSITIVE_IDENTIFIER_PATTERN`), but this will not pick up custom host redaction patterns

Which direction should we take?

spector-in-london:
Heartbeat follow-up on the failing CI for this PR.

I reproduced the failure locally on the PR branch:

```
src/engine.ts(90,37): error TS2307: Cannot find module 'openclaw/plugin-sdk/logging-core' or its corresponding type declarations.
```

I tested a small fix locally:

- replace the optional host import with a local `src/redaction.ts` helper for dedup normalization
- add focused `test/redaction.test.ts` coverage
- keep the package OpenClaw compatibility floor aligned with the existing metadata test at `>=2026.5.22`

Verification on my local checkout of this PR:

```
npm run typecheck
npm test
```

Both pass; full suite result was 50 test files passed, 1,077 tests passed.

I attempted to push the fix to `mpz4life:fix/engine-dedup-sensitive-text-normalization`, but GitHub returned 403 for `spector-in-london`. Local commit is `dd9fc8d fix(engine): avoid host redaction import in dedup normalization` if you want me to provide the patch another way.

mpz4life:
Thanks for looking into this @spector-in-london. We have considered both approaches and prefer to keep the `redactSensitiveText` import from `openclaw/plugin-sdk/logging-core`.

The reason is that sensitive-text redaction patterns are **host-configurable** — the `beforeMessageWrite` hook can be customized with patterns beyond the SDK built-in set. If we copy a fixed regex into lossless-claw, the dedup normalization would drift from the host actual redaction behavior whenever the host config changes. By depending on the SDK, the normalization always uses the same patterns the host applies, regardless of configuration.

We also prefer to follow the platform evolution and version cadence rather than vendoring what should be a shared capability. The `>=2026.6.1` floor is a small bump from `>=2026.5.22` and the `plugin-sdk/logging-core` subpath is already stable in the SDK.

If this version bump is a concern, happy to discuss further — but we believe depending on the SDK surface is the right architectural call here.

mpz4life:
@100yenadmin

100yenadmin:
@mpz4life triage label sync: marking this `bug` / `priority:P2`.

The reported symptom is real for hosts that redact sensitive field names before write, but this is not ready for the P0/P1 merge lane yet because the current head has a failing `test` check and the OpenClaw compatibility-floor/import decision is still unresolved. Please link an existing issue or create a focused issue for the dedup/redaction symptom so we can track the fix without losing the compatibility decision.

jalehman:
Thanks for the contribution, @mpz4life. The deduplication fix has landed with the optional OpenClaw integration preserved and focused replay-safety guards in place.

PR #838, opened 2026-06-06
fix(engine): recover checkpoint-missing afterTurn frontier instead of looping forever (#837)

### LINKED ISSUE #837: afterTurn checkpoint-missing reconcile loops forever: allowNoAnchorImportOnCheckpointMissing only passed on rotate lane, not afterTurn
### Summary

When a conversation row has `bootstrapped_at` set but **no** `conversation_bootstrap_state` row, the `afterTurn` transcript-reconcile slow path classifies it as `reason="checkpoint-missing"`. In this path `reconcileSessionTail` is called **without** `allowNoAnchorImport`, so if the DB frontier holds only non-anchoring rows (e.g. an injected `Conversation info (untrusted metadata)` preamble) it imports **0 messages and never persists a checkpoint**.

Net effect: the session emits this pair **every single turn, forever**:

```
[lcm] afterTurn: transcript reconcile found no anchor and imported 0 messages; skipping checkpoint refresh conversation=<id> reason=checkpoint-missing sessionFile=.../<sid>.jsonl historicalMessages=<n>
[lcm] afterTurn: transcript reconcile did not cover the transcript frontier; skipping afterTurn persistence to avoid creating a future anchor past unreconciled transcript history session=<sid> sessionKey=<key>
```

The guard itself is correct (no-proof-no-advance, per #649). The bug is that the **only** branch able to break out of this state for a `checkpoint-missing` conversation never runs on the `afterTurn` lane, because the unlocking flag is not passed.

### Environment

- `@martian-engineering/lossless-claw` **0.12.0** (npm `latest`)
- OpenClaw `2026.6.1`
- Node `v22.22.2`
- `journal_mode=wal`, `auto_vacuum=0`, `foreign_keys` default off at connection

### Root cause

The afterTurn lane calls reconcile **without** `allowNoAnchorImportOnCheckpointMissing`:

```js
// afterTurn()
transcriptReconcileResult = await this.reconcileTranscriptTailForAfterTurn({
  sessionId: params.sessionId,
  sessionKey: params.sessionKey,
  sessionFile: params.sessionFile,
  isHeartbeat: params.isHeartbeat
}); // <-- no allowNoAnchorImportOnCheckpointMissing
```

Inside the slow path, `allowNoAnchorImport` for a `checkpoint-missing` conversation is gated on exactly that flag:

```js
const reason = !checkpoint ? "checkpoint-missing"
  : checkpoint.sessionFilePath !== params.sessionFile ? "path-mismatch"
  : transcriptEpochShrank ? "same-path-shrink"
  : "append-only-ineligible";

const reconcile = await this.reconcileSessionTail({
  /* ... */
  allowNoAnchorImport:
       reason === "path-mismatch"
    || reason === "same-path-shrink"
    || (reason === "checkpoint-missing" && params.allowNoAnchorImportOnCheckpointMissing === true),
  noAnchorImportReason:
    reason === "checkpoint-missing" && params.allowNoAnchorImportOnCheckpointMissing === true
      ? "rotate-checkpoint-missing"
      : reason
});
```

The flag is set to `true` in exactly **one** place — `reconcileRawTranscriptForRotate` (the rotate lane):

```js
const reconcileParams = {
  sessionId: params.sessionId,
  sessionKey: params.sessionKey,
  sessionFile: params.sessionFile,
  allowNoAnchorImportOnCheckpointMissing: true   // rotate lane only
};
```

So a `checkpoint-missing` conversation can only recover when a **rotate** happens to run. On the ordinary `afterTurn` lane it loops indefinitely. The placeholder-seed escape hatch does **not** apply here either — it only fires on `isMissingFileError` / empty-file branches, not when a real transcript exists with a non-anchoring frontier.

### How a conversation reaches `bootstrapped_at`-without-`bootstrap_state`

This is the state that triggers the loop. It is reachable through rapid, back-to-back session rollovers. On this host the trigger was 6 rollovers in ~90 minutes (`reason: auto-threshold`, `contextThreshold = 0.62`) during a burst of model switches + gateway restarts. After the dust settled, the live conversation had `bootstrapped_at` set, no `conversation_bootstrap_state` row, and a 1-row non-anchoring frontier (the injected `Conversation info (untrusted metadata)` preamble) — i.e. precisely the no-anchor `checkpoint-missing` shape.

### Reproduction

1. Have an active conversation whose DB frontier is only non-anchoring rows (e.g. a single injected `Conversation info (untrusted metadata)` preamble), with `bootstrapped_at` set and **no** `conversation_bootstrap_state` row.
2. Point it at a session file that holds a real, growing transcript.
3. Take any normal turn.

**Observed:** every turn logs `found no anchor and imported 0 messages; skipping checkpoint refresh ... reason=checkpoint-missing` followed by `did not cover the transcript frontier; skipping afterTurn persistence`. The checkpoint is never written, compaction never runs, and the loop is permanent until the conversation is manually archived.

**Expected:** the conversation either recovers (re-imports/seeds a checkpoint) on the `afterTurn` lane the same way the rotate lane does, or seeds a placeholder `bootstrap_state` so the next turn can advance — without an operator manually archiving the row.

### Production evidence

On a single live database this left **49** active conversations in the `bootstrapped_at`-set / `bootstrap_state`-missing state (channel sessions, sub-agent sessions, and NULL-key sessions). The affected channel session produced the warning pair on every turn until the row was archived by hand. Query used:

```sql
SELECT c.conversation_id, c.session_key, c.bootstrapped_at,
       (SELECT count(*) FROM messages m WHERE m.conversation_id = c.conversation_id) AS msgs
FROM conversations c
WHERE c.active = 1
  AND c.bootstrapped_at IS NOT NULL
  AND NOT EXISTS (
        SELECT 1 FROM conversation_bootstrap_state b
        WHERE b.conversation_id = c.conversation_id
      );
```

### Suggested fix

Any one of these closes the loop:

1. **Pass the flag on the afterTurn lane.** Allow `afterTurn`'s reconcile to set `allowNoAnchorImportOnCheckpointMissing: true` for `reason === "checkpoint-missing"`, mirroring the rotate lane — so a missing-checkpoint conversation with a non-anchoring frontier re-imports its transcript instead of looping.
2. **Seed a placeholder `bootstrap_state` on `checkpoint-missing`.** Extend the existing placeholder-seed (currently limited to `isMissingFileError` / empty-file) to also seed an all-zero checkpoint when `reason === "checkpoint-missing"` and a real transcript exists, so the next turn can advance via the append-only/recovery path.
3. **Never leave `bootstrapped_at` set without a `bootstrap_state` row.** Make `markConversationBootstrapped()` and `persistBootstrapState()` atomic across the rollover paths so the orphaned shape cannot be produced in the first place.

Option 1 is the smallest change and directly symmetric with the rotate lane.

### Relationship to existing issues

- **#822** (P1) — same `0-import / LCM-no-op` family, but for the **placeholder** (all-zero) `bootstrap_state` shape. This report is the sibling case where the `bootstrap_state` row is **absent entirely** (`reason=checkpoint-missing`), and highlights the rotate-vs-afterTurn flag asymmetry.
- **#390** — "checkpoint not persisted when reconcileSessionTail finds no overlap." Related root cause (conditional persist), different lane/branch.
- **#686 / #649** — the no-proof-no-advance guard this interacts with. The guard is working as intended; the gap is the missing recovery path on the afterTurn lane.

holgergruenhagen:
## Summary

Fixes #837 — an `afterTurn` deadlock that emits this warning pair **every single turn, forever** on affected sessions:

```
[lcm] afterTurn: transcript reconcile found no anchor and imported 0 messages; skipping checkpoint refresh ... reason=checkpoint-missing
[lcm] afterTurn: transcript reconcile did not cover the transcript frontier; skipping afterTurn persistence ...
```

For such a conversation, compaction never runs and LCM is a permanent no-op until the row is archived by hand.

## Root cause

A conversation with `bootstrapped_at` **set** but **no** `conversation_bootstrap_state` row classifies as `reason="checkpoint-missing"` on the `afterTurn` slow path. When the DB frontier holds only **non-anchoring** rows (e.g. a single injected `Conversation info (untrusted metadata)` preamble), `reconcileSessionTail` imports **0 messages and never persists a checkpoint**.

The recovery flag `allowNoAnchorImportOnCheckpointMissing` exists, but was only ever passed on the **rotate** lane (`reconcileRawTranscriptForRotate`), never on the `afterTurn` lane. So a `checkpoint-missing` conversation could only recover if a rotate happened to run; on the ordinary turn lane it looped indefinitely.

## Fix

`afterTurn` now also recovers a `checkpoint-missing` no-anchor frontier — but **gated on `conversation.bootstrappedAt`**:

- **bootstrapped_at set** (the #837 shape) → import the no-anchor epoch and persist a checkpoint, mirroring the rotate lane.
- **bootstrapped_at NULL** (a never-bootstrapped conversation whose transcript was rewritten to divergent content) → still freezes, preserving **#649**'s no-proof-no-advance guard.

The discriminator was verified empirically: the existing #649 divergent-transcript test runs with `bootstrappedAt = null`, while the #837 production shape has it set. The downstream no-anchor import stays guarded by replay-overlap detection, the import cap, and the delivery-only block — so this recovers instead of grafting a foreign transcript or flooding.

This is **Option 1** from the issue (smallest change, symmetric with the rotate lane), refined with the `bootstrappedAt` gate so it does not regress #649.

## Tests

- New regression test reproduces the exact production shape (bootstrapped, no state row, single injected metadata-preamble frontier) and asserts:
  - the transcript history is imported and a `bootstrap_state` checkpoint is persisted (`lastProcessedOffset > 0`),
  - the forever-loop warning pair does **not** fire,
  - a second ordinary turn keeps advancing on the fast path.
- Verified the test **fails without the fix** (checkpoint stays `null`) and **passes with it**.
- Full suite green: **1132/1132** tests pass. The previously-passing #649 test (`afterTurn skips persistence when full reread finds no anchor and imports nothing`) still passes unchanged.

## Production evidence

Reproduced on a live host: **49** active conversations in the `bootstrapped_at`-set / `bootstrap_state`-missing state on a single database — exactly the count reported in #837. The affected channel session produced the warning pair on every turn until the row was archived by hand.

## Notes

- `@martian-engineering/lossless-claw` **0.12.0**, branched off current `upstream/main`.
- Includes a `patch` changeset.
- `dist/` is not committed (built on release).

### DIFF
--- a/.changeset/afterturn-checkpoint-missing-no-anchor.md
+++ b/.changeset/afterturn-checkpoint-missing-no-anchor.md
@@ -1,0 +1,9 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix an `afterTurn` deadlock where a conversation with `bootstrapped_at` set but no `conversation_bootstrap_state` row (`reason="checkpoint-missing"`) and a non-anchoring DB frontier (e.g. a single injected `Conversation info (untrusted metadata)` preamble) imported 0 messages and never persisted a checkpoint. Such conversations emitted the `found no anchor and imported 0 messages` / `did not cover the transcript frontier` warning pair on every turn forever, with compaction permanently disabled until the row was archived by hand.
+
+The recovery path (`allowNoAnchorImportOnCheckpointMissing`) previously ran only on the rotate lane. The `afterTurn` lane now also recovers a `checkpoint-missing` no-anchor frontier, but only for already-bootstrapped conversations (`bootstrapped_at` set) — a never-bootstrapped conversation with a divergent rewritten transcript still freezes per #649's no-proof-no-advance guard. The downstream no-anchor import remains guarded by replay-overlap detection, the import cap, and the delivery-only block.
+
+Fixes #837.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -7379,6 +7379,24 @@ export class LcmContextEngine implements ContextEngine {
             hasOverlap: sessionFileState.size === 0,
           };
         }
+        // #837: a conversation with bootstrapped_at SET but no bootstrap_state
+        // row reaches reason="checkpoint-missing" with a non-anchoring frontier
+        // (e.g. a single injected metadata preamble). Without a no-anchor import
+        // it imports 0 messages and never persists a checkpoint, so afterTurn
+        // loops the "did not cover the transcript frontier" warning forever and
+        // compaction never runs. The rotate lane already recovers via
+        // allowNoAnchorImportOnCheckpointMissing; mirror that on the afterTurn
+        // lane, but ONLY for already-bootstrapped conversations. A never
+        // -bootstrapped conversation (bootstrapped_at NULL) with a divergent
+        // rewritten transcript must still freeze per #649's no-proof-no-advance
+        // guard, so we gate on conversation.bootstrappedAt to avoid grafting a
+        // foreign transcript onto genuine history. The downstream no-anchor
+        // import path is itself guarded (replay-overlap detection, import cap,
+        // delivery-only block).
+        const recoverCheckpointMissingNoAnchor =
+          reason === "checkpoint-missing" &&
+          (params.allowNoAnchorImportOnCheckpointMissing === true ||
+            conversation.bootstrappedAt !== null);
         const reconcile = await this.reconcileSessionTail({
           sessionId: params.sessionId,
           sessionKey: params.sessionKey,
@@ -7388,11 +7406,12 @@ export class LcmContextEngine implements ContextEngine {
           allowNoAnchorImport:
             reason === "path-mismatch" ||
             reason === "same-path-shrink" ||
-            (reason === "checkpoint-missing" && params.allowNoAnchorImportOnCheckpointMissing === true),
-          noAnchorImportReason:
-            reason === "checkpoint-missing" && params.allowNoAnchorImportOnCheckpointMissing === true
+            recoverCheckpointMissingNoAnchor,
+          noAnchorImportReason: recoverCheckpointMissingNoAnchor
+            ? params.allowNoAnchorImportOnCheckpointMissing === true
               ? "rotate-checkpoint-missing"
-              : reason,
+              : "checkpoint-missing-recovery"
+            : reason,
         });
         if (reconcile.blockedByImportCap) {
           return { importedMessages: 0, blockedByImportCap: true, hasOverlap: reconcile.hasOverlap };
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -6333,7 +6333,8 @@ export class LcmContextEngine implements ContextEngine {
       if (anchorIndex < 0) {
         if (params.allowNoAnchorImport) {
           if (
-            params.noAnchorImportReason === "path-mismatch" &&
+            (params.noAnchorImportReason === "path-mismatch" ||
+              params.noAnchorImportReason === "checkpoint-missing-recovery") &&
             isLikelyInjectedDeliveryOnlyTranscript(historicalMessages)
           ) {
             this.deps.log.warn(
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for the contribution, @holgergruenhagen. This is landed now.

holgergruenhagen:
Thanks Josh, appreciate the quick merge and the extra hardening commits.

PR #839, opened 2026-06-06
fix(lcm-tools): support session keys for resumed thread recall

KatTate:
## Summary
- Adds explicit `sessionKey` support to agent-facing LCM tools: `lcm_grep`, `lcm_describe`, `lcm_expand`, and `lcm_expand_query`.
- Keeps `conversationId` for real LCM database IDs, but validates finite positive safe integers and integer strings only.
- Rejects likely Discord snowflake IDs with an actionable error telling agents to use `sessionKey` for Discord/thread recall.
- Documents the tool contract and adds regression coverage.

## Why
Discord thread/session recall can resume under a stable session key while the tool contract currently exposes only numeric `conversationId`. Agents then have two bad failure modes: they miss resumed-thread context, or they pass Discord snowflakes as `conversationId`, which is not the LCM physical `conversation_id` and is unsafe to parse as a JS number.

This change makes resumed-thread recall explicit and fails closed on snowflake-like IDs.

## Verification
- `pnpm vitest run test/lcm-tools.test.ts --reporter=basic`
  - 19 tests passed
- `pnpm run typecheck`
  - passed
- `pnpm run build`
  - passed
- `pnpm test`
  - 50 test files passed, 1137 tests passed
- `pnpm run release:verify`
  - typecheck passed
  - build passed
  - test suite passed: 50 files, 1137 tests
  - `npm pack --dry-run` completed

## Live validation from downstream install
A local OpenClaw install using this patch was rebuilt and verified separately. The installed bundle contained the expected `sessionKey`, Discord-snowflake, and LCM database `conversation_id` diagnostics, and `lossless-claw` remained loaded/enabled/activated as the selected context engine.

### DIFF
--- a/docs/agent-tools.md
+++ b/docs/agent-tools.md
@@ -41,7 +41,8 @@ Use `mode: "full_text"` for keyword or topical recall. Full-text queries are not
 | `pattern` | string | ✅ | — | Search pattern |
 | `mode` | string | | `"regex"` | `"regex"` or `"full_text"` |
 | `scope` | string | | `"both"` | `"messages"`, `"summaries"`, or `"both"` |
-| `conversationId` | number | | current session family | Specific physical conversation to search |
+| `conversationId` | number/string | | current session family | LCM database conversation_id. Do not pass a Discord snowflake. |
+| `sessionKey` | string | | current session family | Stable LCM session key for resumed Discord/thread sessions |
 | `allConversations` | boolean | | `false` | Search all conversations |
 | `since` | string | | — | ISO timestamp lower bound |
 | `before` | string | | — | ISO timestamp upper bound |
@@ -81,7 +82,8 @@ Look up metadata and content for a specific summary or stored file.
 | Param | Type | Required | Default | Description |
 |-------|------|----------|---------|-------------|
 | `id` | string | ✅ | — | `sum_xxx` for summaries, `file_xxx` for files |
-| `conversationId` | number | | current session family | Scope to a specific physical conversation |
+| `conversationId` | number/string | | current session family | LCM database conversation_id. Do not pass a Discord snowflake. |
+| `sessionKey` | string | | current session family | Stable LCM session key for resumed Discord/thread sessions |
 | `allConversations` | boolean | | `false` | Allow cross-conversation lookups |
 
 **Returns for summaries:**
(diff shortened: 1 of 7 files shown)


KatTate:
Clarifying scope after tracing the current OpenClaw release and upstream issue state.

This PR is downstream hardening for LCM agent tools. It is still useful because it:

- adds explicit `sessionKey` support to `lcm_grep`, `lcm_describe`, `lcm_expand`, and `lcm_expand_query`
- keeps `conversationId` scoped to real LCM database IDs
- rejects likely Discord snowflake IDs instead of letting agents accidentally treat channel/thread IDs as LCM `conversation_id` values

But this is not the OpenClaw root-cause fix for sibling Discord thread identity collapse.

The upstream OpenClaw root-cause lane is:

- openclaw/openclaw#64199, currently closed and locked
- openclaw/openclaw#64322, currently open

Fresh check on latest published OpenClaw shows the parent-channel fallback still present in `2026.6.1 (2e08f0f)`:

```js
function matchDiscordAcpConversation(params) {
	if (params.bindingConversationId === params.conversationId) return {
		conversationId: params.conversationId,
		matchPriority: 2
	};
	if (params.parentConversationId && params.parentConversationId !== params.conversationId && params.bindingConversationId === params.parentConversationId) return {
		conversationId: params.parentConversationId,
		matchPriority: 1
	};
	return null;
}
```

That OpenClaw behavior can still materialize inherited Discord thread bindings using the parent channel ID instead of the live thread ID. LCM can harden its tool interface around this, but LCM cannot make OpenClaw assign distinct ACP/session identities to sibling Discord threads. The root fix needs #64322 or an equivalent OpenClaw patch.

Posted current-release proof on the OpenClaw PR here:

https://github.com/openclaw/openclaw/pull/64322#issuecomment-4640464805

jalehman:
Thanks for narrowing this to downstream LCM tool hardening. I agree the feature is a good fit: explicit `sessionKey` scoping and rejecting Discord snowflakes as `conversationId` are the right user-facing contract.

I think this needs one security/scope fix before full review. In `resolveLcmConversationScope`, the tool param `sessionKey` currently replaces the runtime `input.sessionKey` before delegated-session detection. For direct recall tools such as `lcm_grep` and `lcm_describe`, that appears to let a sub-agent pass a non-subagent `sessionKey` and have the resolver treat the call as non-delegated, bypassing the delegated grant boundary that currently protects sub-agent retrieval.

Could you update the resolver so delegated status is derived from the runtime caller identity, not from the requested lookup key, and add a regression where a sub-agent with no/limited grant cannot use `sessionKey` to search or describe another session family? It would also help to cover the advertised `sessionKey` schema/behavior beyond `lcm_grep`.

Since this changes user-facing tool behavior/docs, it will also need a patch changeset before merge.

jalehman:
Thanks for the careful scope clarification here. I agree this is the right area for Lossless: explicit `sessionKey` support for resumed thread recall and rejecting Discord snowflakes as `conversationId` are both useful tool-contract improvements.

I’m going to close this version for now because the current resolver still has a delegated-scope safety issue. `params.sessionKey` can replace the runtime `input.sessionKey` before delegated-session detection, so a sub-agent call can appear non-delegated when resolving scope. That risks bypassing the retrieval grant boundary protecting sub-agent access.

A revised PR would be welcome if it:

- derives delegated status from the runtime caller identity, not the requested lookup key
- adds regression coverage proving a sub-agent cannot use `sessionKey` to search/describe another session family without an appropriate grant
- covers the advertised `sessionKey` schema/behavior beyond `lcm_grep`
- adds a patch changeset for the user-facing tool/docs change

PR #844, opened 2026-06-08
fix: skip synthetic heartbeats during bootstrap reconcile

jalehman:
## Summary

- filter synthetic `[OpenClaw heartbeat poll]` transcript rows out of bootstrap/reconcile import candidates
- filter full HEARTBEAT_OK transcript turns during reconcile while preserving adjacent real tail messages
- keep the generic user anti-replay guard intact; this only removes known synthetic host traffic before import/replay evaluation

## Why

PR #840 fixed heartbeat-marked afterTurn append-only deltas, but an existing transcript can still contain synthetic heartbeat rows before bootstrap/full-reconcile runs. Those rows can collapse into one SQLite second during reconcile import and trip the role=user replay flood guard, causing host quarantine/fallback.

## Verification

- `env -u OPENCLAW_STATE_DIR npx vitest run test/engine.test.ts --testNamePattern "skips synthetic heartbeat polls|skips heartbeat ACK turns|heartbeat flag skips append-only transcript deltas"`
- `npm run build`
- `env -u OPENCLAW_STATE_DIR npm test`

### DIFF
--- a/.changeset/fuzzy-heartbeats-reconcile.md
+++ b/.changeset/fuzzy-heartbeats-reconcile.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Skip synthetic OpenClaw heartbeat transcript rows during bootstrap/reconcile imports so heartbeat-only tails cannot trip replay-flood quarantine.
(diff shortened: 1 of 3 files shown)


jalehman:
Closing as superseded by #846, which merged the same patch and is already on main.

PR #847, opened 2026-06-08
Add configurable context threshold overrides

hubofvalley:
## Summary
- add optional `contextThresholdOverrides` config rules for model, effective context window, and session pattern matching
- keep the existing global `contextThreshold` fallback unchanged
- thread the resolved threshold through compaction evaluation, manual compaction, sweeps, and derived summary-prefix target sizing
- document the new config and add validation/matching coverage

## Behaviour
- Overrides are optional and backward compatible.
- If no override matches, Lossless uses the existing global `contextThreshold`.
- If multiple rules match, the resolver picks the highest specificity, then the earliest rule in config order.
- Override values use the same numeric range as global `contextThreshold`.

## Motivation
A single global threshold behaves differently across models with very different context windows. Operators may want earlier compaction for smaller/effective budgets and different operational targets for larger-context models or specific session classes.

## Tests
- `npm run typecheck`
- `npm test` (53 files, 1209 tests)

### DIFF
--- a/.changeset/context-threshold-overrides.md
+++ b/.changeset/context-threshold-overrides.md
@@ -1,0 +1,6 @@
+---
+"@martian-engineering/lossless-claw": minor
+---
+
+Add scoped context threshold overrides for matching model ids, model context-window
+ranges, and session patterns.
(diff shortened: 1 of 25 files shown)


jalehman:
Thanks for the contribution, @hubofvalley. This is landed now.

PR #862, opened 2026-06-10
fix(assemble): intercept large tool results in live messages before degraded fallback

mpz4life:
## Summary

- **Problem**: When the live context exceeds the token budget, `assemble()` enters a degraded live fallback path that returns `params.messages` directly, bypassing the DB assembly and all stubbing. This means every tool result in the current turn is sent to the model as raw content even though the DB stores `[LCM Tool Output: …]` stubbed references from the `afterTurn` ingest path.
- **Why it matters**: Users see `max_completion_tokens=1` clamped requests because the pipeline sends raw full-content tool results to the model. Even with `stubLargeToolPayloads:true` and `largeFileTokenThreshold` configured, the degraded fallback circumvents all stubbing.
- **What changed**: Run `interceptLargeToolResults` on `params.messages` at `assemble()` entry point, before any fallback path. Gated behind `config.largeFileTokenThreshold > 0`.
- **What did NOT change**: `afterTurn` ingest still runs its own interceptor (unchanged). Normal assembly still processes DB content through assembler (unchanged).

## Evidence

- Regression test: `assemble() intercepts large tool results in live messages before degraded fallback` (passes)

## Compatibility

- Backward compatible: Yes
- Config changes: No
- Migration needed: No
- Gated behind `config.largeFileTokenThreshold > 0`, default no-op

### DIFF
--- a/.changeset/live-tool-result-stubs.md
+++ b/.changeset/live-tool-result-stubs.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Stub oversized live tool results before assemble fallback paths send context to the model, so degraded assembly keeps compact file references instead of raw payloads.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -10415,6 +10415,27 @@ export class LcmContextEngine implements ContextEngine {
         );
         return safeFallback();
       }
+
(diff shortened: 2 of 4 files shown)


mpz4life:
@jalehman  would you mind reviewing this PR? Thanks!

jalehman:

PR #873, opened 2026-06-11
fix: preserve assembly while transcript coverage recovers

sercada:
## Summary

- detect all-zero placeholder bootstrap checkpoints before deferred-debt assembly handling
- skip emergency deferred compaction drain / degraded live fallback while transcript coverage is still placeholder-only and summaryless
- add a regression test for the placeholder-checkpoint + pending-threshold-debt shape
- add a patch changeset

## Why

This is the user-visible fallout from the transcript-ingestion wedge family around #822/#639: if a conversation has only placeholder transcript coverage, a pending threshold-maintenance retry can run during `assemble()` and force the degraded live fallback before the transcript recovery path has advanced the checkpoint. In production that presents as a final round with a collapsed prompt even though the tool rounds were coherent.

Current `main` already has the newer entry-id reconciliation and bounded fallback work; this patch is intentionally narrower. It does not loosen the bounded fallback policy. It only prevents deferred debt from driving emergency compaction or degraded assembly while LCM has no summary coverage and the checkpoint is still the all-zero placeholder.

## Tests

- `npm run typecheck`
- `npm run build`
- `npx vitest run test/engine.test.ts -t "assemble\\(\\) (uses bounded live context when pending maintenance is near budget|ignores deferred debt while placeholder transcript coverage recovers|intercepts large tool results in live messages before degraded fallback|degrades to bounded live context if emergency compaction leaves debt pending)"`
- `npx vitest run test/ambiguous-rollover-rotation.test.ts`
- `npx vitest run --dir test --pool=forks --maxWorkers=1 --minWorkers=1` (60 files / 1281 tests passed)

### DIFF
--- a/.changeset/placeholder-debt-assembly.md
+++ b/.changeset/placeholder-debt-assembly.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Skip deferred assembly debt degradation while placeholder transcript coverage is still recovering.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3395,6 +3395,13 @@ export class LcmContextEngine implements ContextEngine {
         return degraded;
       }
 
+      if (incompleteTranscriptCoverage) {
+        this.deps.log.warn(
+          `[lcm] assemble: bounded live fallback while transcript coverage recovers conversation=${conversation.conversationId} ${sessionLabel} reason=incomplete-transcript-coverage contextItems=${contextItems.length} liveMessages=${liveMessages.length} placeholderBootstrapState=${placeholderBootstrapState} hasSummaryItems=${hasSummaryItems} tokenBudget=${tokenBudget}`,
+        );
+        return boundedLiveFallback("incomplete-transcript-coverage");
+      }
+
       if (contextItems.length === 0) {
         if (forkBoundedBootstrap) {
           const boundedFallback = buildForkBoundedLiveFallback({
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks, this is directionally useful as an assembly-side safety valve, and I agree with the data-preservation goal. I am holding it before autoreview/merge because the guard is currently broader than the post-#885 recovery proof on main.

Current main has the narrower shape-proven placeholder recovery path via `conversationFrontierIsEntirelyNonAnchoring`. This PR instead treats all-zero `bootstrap_state` plus no summary items as incomplete coverage and then skips deferred-debt degradation in favor of bounded live fallback. That can mitigate prompt collapse, but it also risks bypassing the emergency deferred-debt drain for placeholder/no-summary conversations that include real persisted rows where main intentionally freezes rather than risking contamination.

Could you tighten the guard against the post-#885 recovery model, or add focused coverage for real-frontier placeholder/frozen states, the #884 tail-only gap shape, and the #887 post-rollover freeze behavior? It would also help to clarify in the PR body that this is an assemble-time mitigation, not the root transcript-recovery fix.

jalehman:
Thanks for the careful work here. I’m going to close this for now because the PR is solving a real Lossless problem, but the current guard is still broader than the recovery proof now on `main`.

The specific concern is that `all-zero bootstrap_state + no summary items` can also describe placeholder conversations that already contain real persisted rows. Current `main` deliberately freezes those shapes rather than importing or assembling through them, because doing otherwise can contaminate unrelated/rotated transcript history.

A safer follow-up would either tighten this assembly-side mitigation to match the post-#885 `conversationFrontierIsEntirelyNonAnchoring` recovery model, or add focused regression coverage for:

- real-frontier placeholder/frozen states
- the #884 tail-only gap shape
- the #887 post-rollover freeze behavior

This remains a good fit as a narrow assemble-time mitigation, but it should not be framed as the root transcript-recovery fix.

PR #876, opened 2026-06-11
fix(engine): allow no-anchor import in placeholder-checkpoint and checkpoint-missing reconcile paths

### LINKED ISSUE #875: new sessions stuck with 1 DB message while JSONL grows — reconcile deadlocks on missing allowNoAnchorImport
## Problem

New sessions consistently get stuck with only 1 message in the LCM database while the JSONL transcript file accumulates dozens of messages. This affects **every new session**, preventing LCM from storing conversation context and blocking compaction entirely.

**Observed behavior:**
- JSONL: 70+ messages (normal user conversation)  
- LCM DB: 1 message (only the initial bootstrap entry)  
- Log loop:
```
[lcm] afterTurn: transcript reconcile did not cover the transcript frontier;
skipping afterTurn persistence to avoid creating a future anchor
past unreconciled transcript history
```

### 1. Two-phase afterTurn lifecycle + JSONL creation timing

A new agent session triggers LCM's `afterTurn` **twice** — once during the bootstrap phase and once after the LLM pipeline completes. The JSONL file is NOT created by the time the first `afterTurn` fires:

```
Step 1: bootstrap phase afterTurn
  stat(sessionFile) --> ENOENT (JSONL not created yet)
  seeds placeholder checkpoint (size=0, mtime=0, offset=0, hash=null)
  refreshBootstrapState --> stat() --> still ENOENT
  --> this is the "bootstrap checkpoint refresh failed" in user logs

Step 2: assemble() --> 1 message (user's first input)

Step 3: LLM streaming begins
  first assistant message --> writeJsonlEntriesSync() --> JSONL created
  (session-manager.ts:906)

Step 4: main afterTurn
  stat(sessionFile) --> SUCCESS (file now exists)
  reads 3+ historicalMessages
  reconcileSessionTail --> no anchor found
  --> "transcript reconcile did not cover the transcript frontier"
```

**JSONL creation proof**: `SessionManager.persist()` at `session-manager.ts:891-908` — first user message sets `flushed=false` (no file write), first assistant message triggers `writeJsonlEntriesSync()` (file created). Source: OpenClaw core `packages/agent-core`.

### 2. Session JSONL file exists at the expected path

Verified the session file exists at the path in the ENOENT logs:
```
ENOENT: stat '/home/xxxx/.openclaw/agents/main/sessions/576eae3c-...jsonl'
```
The file existed at the time the assertion was made (238KB, 70 lines), ruling out path mismatch or cleanup.

### 3. Actual log evidence

```
[lcm] afterTurn: bootstrap checkpoint refresh failed for session=576eae3c-...
  ENOENT: no such file or directory, stat '.../576eae3c-...jsonl'
[lcm] afterTurn: transcript reconcile found no anchor and imported 0 messages;
  skipping checkpoint refresh conversation=51 reason=checkpoint-missing
  historicalMessages=3
[lcm] afterTurn: transcript reconcile did not cover the transcript frontier;
  ...
```

`historicalMessages` grows from 3 to 7 to 9... but LCM DB stays at 1 message.

### 4. Code: reconcileSessionTail missing allowNoAnchorImport

Two paths call `reconcileSessionTail` without `allowNoAnchorImport: true`:

**A. append-only placeholder recovery** (`engine.ts:6558`):
```
const reconcile = await this.reconcileSessionTail({
  ...
  // MISSING: allowNoAnchorImport: true
  noAnchorImportReason: "placeholder-checkpoint-recovery",
});
```

**B. slow path checkpoint-missing** (`engine.ts:6823`):
```
allowNoAnchorImport:
  reason === "path-mismatch" ||
  reason === "same-path-shrink" ||
  // "checkpoint-missing" NOT enabled for normal afterTurn call
```

### 5. How the deadlock forms

1. First afterTurn seeds placeholder checkpoint (JSONL not yet created)
2. Main afterTurn: file exists, checkpoint exists, enters append-only path
3. Placeholder detected, calls `reconcileSessionTail`
4. No anchor found (DB content != JSONL content after normalize/storage conversion)
5. `allowNoAnchorImport` not set -> returns `hasOverlap=false, importedMessages=0`
6. `transcriptReconcileUnsafeToAdvance = true` -> skip all persistence
7. Checkpoint never refreshed -> next afterTurn repeats -> **infinite loop**

### 6. Guard mechanisms in reconcileSessionTail

The no-anchor import path already has flood protection:
- **importCap**: `max(existingDbCount * 0.2, 50)`
- **replay analysis**: identity overlap detection
- **delivery-only block**: filters injected traffic

These guards are sufficient — `allowNoAnchorImport` should be enabled for checkpoint-missing paths without additional restrictions.

## Fix

Two locations in `engine.ts` need `allowNoAnchorImport: true`:

1. **Append-only placeholder-checkpoint-recovery**: add `allowNoAnchorImport: true` to the `reconcileSessionTail` call
2. **Slow path checkpoint-missing**: when `existingMessageCount <= 3` and session matches, allow bounded no-anchor import

## Impact

- Every new session silently loses context history
- Compaction never triggers (no checkpoint -> no token count -> no threshold evaluation)
- LCM degrades to transparent pass-through

## Config

```json
"lossless-claw": {
  "enabled": true,
  "config": {
    "autoRotateSessionFiles": { "enabled": false },
    "largeFileThresholdTokens": 8000,
    "stubLargeToolPayloads": true,
    "contextThreshold": 0.7
  }
}
```

mpz4life:
## Summary

Fix JSONL/DB mismatch bug where new sessions get stuck with only 1 message in LCM DB while JSONL has many.

**Root cause:** Two places in `engine.ts` call `reconcileSessionTail` without `allowNoAnchorImport: true`:

1. **Append-only placeholder-checkpoint-recovery** — when a placeholder checkpoint is seeded (stat failure on first turn) and the next turn reads the transcript, the reconcile call lacked `allowNoAnchorImport`, so non-anchoring content was never imported.

2. **Slow path checkpoint-missing** — when `reason === "checkpoint-missing"` and the conversation has ≤ 3 messages (new session), the no-anchor import is now allowed. Sessions with substantial history (> 3 messages) remain protected from divergent transcript overwrites.

## Changes

- `src/engine.ts`: Add `allowNoAnchorImport: true` to placeholder-checkpoint reconcile call
- `src/engine.ts`: Add `checkpointMissingLowMessageCount` condition for sessions with ≤ 3 messages
- `test/engine.test.ts`: Two regression tests covering both paths

## Verification

- All checkpoint-related tests pass (8/8)
- Existing "does not recover checkpoint-missing divergent transcript over real history" test updated to 4+ messages to validate the protection boundary

Closes #875

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -7647,6 +7647,7 @@ export class LcmContextEngine implements ContextEngine {
                 sessionKey: params.sessionKey,
                 conversationId: conversation.conversationId,
                 historicalMessages: appended.messages,
+                allowNoAnchorImport: true,
                 noAnchorImportReason: "placeholder-checkpoint-recovery",
               });
               if (reconcile.importedMessages > 0) {
@@ -7953,6 +7954,7 @@ export class LcmContextEngine implements ContextEngine {
         // import path is itself guarded (replay-overlap detection, import cap,
         // delivery-only block).
         let checkpointMissingMetadataFrontier = false;
+        let checkpointMissingLowMessageCount = false;
         if (
           reason === "checkpoint-missing" &&
           conversation.sessionId === params.sessionId &&
@@ -7966,11 +7968,13 @@ export class LcmContextEngine implements ContextEngine {
             existingMessageCount === 1 &&
             latestPersistedMessage !== null &&
             isLikelyInjectedMetadataPreambleRecord(latestPersistedMessage);
+          checkpointMissingLowMessageCount = existingMessageCount <= 3;
         }
         const recoverCheckpointMissingNoAnchor =
           reason === "checkpoint-missing" &&
           (params.allowNoAnchorImportOnCheckpointMissing === true ||
-            checkpointMissingMetadataFrontier);
+            checkpointMissingMetadataFrontier ||
+            checkpointMissingLowMessageCount);
         // A transcript whose session header id differs from the checkpoint's
         // is a *declared* epoch change (rewrite/rotation) — no heuristics
         // needed, so a same-path full rewrite may import as a new epoch
--- a/test/engine.test.ts
+++ b/test/engine.test.ts
@@ -12561,6 +12561,8 @@ describe("LcmContextEngine fidelity and token budget", () => {
     writeLeafTranscript(oldSessionFile, [
       { role: "user", content: "old checkpoint-missing real-history question" },
       { role: "assistant", content: "old checkpoint-missing real-history answer" },
+      { role: "user", content: "old checkpoint-missing real-history followup" },
+      { role: "assistant", content: "old checkpoint-missing real-history followup answer" },
     ]);
     await engine.bootstrap({
       sessionId,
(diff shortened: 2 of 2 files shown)


mpz4life:
@jalehman would you mind reviewing this PR? Thanks!

jalehman:
Closing as superseded by the transcript reconciliation refactor and #885.

Three things converged on this PR:

1. **Conflicts with the refactor.** This PR edits the `reconcileSessionTail` call sites in `engine.ts`, but phase 4 (#872) moved that logic into `TranscriptReconciler`. It's now the only open reconcile PR GitHub reports as CONFLICTING.

2. **The checkpoint-missing half is already on main.** `transcript-reconciler.ts` (the #837 metadata-frontier gate, ~line 1922) already recovers checkpoint-missing conversations with a non-anchoring frontier — via an `existingMessageCount === 1` + injected-metadata-preamble proof rather than this PR's `≤ 3` message-count heuristic.

3. **#885 is the replacement.** #885 was written against the refactored reconciler and re-implements both halves of this same #822-class fix with a stronger gate: a `conversationFrontierIsEntirelyNonAnchoring` predicate that proves the persisted frontier is entirely injected metadata, rather than this PR's blanket `allowNoAnchorImport: true` + message-count heuristic. It also fixes the placeholder-checkpoint-recovery lane being dead code on main (the reason string is passed without ever setting `allowNoAnchorImport`).

Let's land the fix via #885. Reopen if I've missed a case #885 doesn't cover.

PR #881, opened 2026-06-11
fix: preserve transcript timestamps during recovery imports

christian-lallo:
## Summary

- Add an optional historical `createdAt` path for message inserts while preserving the existing current-timestamp default for runtime ingests.
- Resolve recovered transcript message timestamps from JSONL envelope metadata, with message-level timestamp fields as a fallback.
- Use those timestamps when bootstrap/reconciliation imports transcript history so recovered rows retain their original chronology.

## Why

The stable-entry-id reconciliation (#854, #866) makes transcript recovery idempotent and avoids content-identity ambiguity, but recovered rows still received the import-time SQLite timestamp. That flattens historical transcript ranges to the recovery moment, which can distort recency-ordered recall, telemetry, and timeline inspection.

Because entry-id-verified imports bypass the replay-timestamp flood guard, the flattening is silent — there is no longer any code path that surfaces it.

This keeps the stable-entry-id behavior while preserving the transcript's original chronology.

## Tests

- `npm run typecheck`
- `npm run build`
- `npx vitest run test/engine-bootstrap.test.ts --testNamePattern "does not advance the bootstrap checkpoint when reconcile aborts at the import cap"`
- `npx vitest run test/engine-bootstrap.test.ts test/transcript-entry-id.test.ts test/anti-replay-burst-regression.test.ts`
- `npm test`

### DIFF
--- a/.changeset/preserve-recovered-transcript-timestamps.md
+++ b/.changeset/preserve-recovered-transcript-timestamps.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve original transcript timestamps during recovery imports instead of stamping import time.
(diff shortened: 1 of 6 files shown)


jalehman:
Thanks for the contribution, @christian-lallo. This is landed now.

Issue #874, opened 2026-06-11
Message ingestion stops after first message per conversation (v0.11.3 / v0.12.0)


drmagnuson:
The key facts:

• Every conversation shows exactly 1 message in the DB across v0.11.3 and v0.12.0
• Active session JSONL is growing normally — OpenClaw writes fine, LCM doesn't ingest
• Session key agent:main:main doesn't match any ignore pattern
• No LCM env var overrides
• Started ~June 5-7, coinciding with automated openclaw configure/doctor --fix runs
• Reported healthy by /lossless doctor despite total ingestion failure

Not sure what else I can provide to be more helpful.  

Here is my config in case I am to blame. 🤦‍♂️
 

> `"lossless-claw": {
>         "enabled": true,
>         "config": {
>           "freshTailCount": 64,
>           "leafChunkTokens": 80000,
>           "newSessionRetainDepth": 2,
>           "contextThreshold": 0.75,
>           "sweepMaxDepth": 2,
>           "autoRotateSessionFiles": {
>             "enabled": false
>           },
>           "summaryModel": "anthropic/claude-haiku-4-5",
>           "expansionModel": "anthropic/claude-haiku-4-5",
>           "delegationTimeoutMs": 300000,
>           "summaryTimeoutMs": 60000,
>           "ignoreSessionPatterns": [
>             "agent:*:cron:**",
>             "agent:main:subagent:**"
>           ],
>           "statelessSessionPatterns": [],
>           "skipStatelessSessions": true
>         },
>         "subagent": {
>           "allowModelOverride": true,
>           "allowedModels": [
>             "anthropic/claude-haiku-4-5"
>           ]
>         },
>         "llm": {
>           "allowModelOverride": true,
>           "allowedModels": [
>             "anthropic/claude-haiku-4-5"
>           ]
>         }
>       },
> `

drmagnuson:
#876 seems to address this.

PR #885, opened 2026-06-12
fix(reconcile): recover non-anchoring-frontier conversations past the no-anchor import cap (#822)

### LINKED ISSUE #822: Placeholder-checkpoint recovery never re-imports a large pre-existing transcript when the DB frontier has no anchor (LCM stays a no-op for the session) — regression of #685

rafaelreis-r:
### What

Narrow follow-up to #824's closure ("the underlying #822-class bug is still valid and needs a narrower follow-up fix"). The unlocking condition is now proven from the DB shape itself, never from the recovery reason:

- New predicate `conversationFrontierIsEntirelyNonAnchoring` (in `TranscriptReconciler`): true only when every persisted row (bounded scan, ≤32 rows) is an injected `Conversation info (untrusted metadata)` preamble. This generalizes the `existingMessageCount === 1` gate from #837/#838 — hosts inject one metadata preamble per delivery, so a stuck conversation accumulates several before anything real persists (we hit a 2-row frontier in production; the count===1 gate rejects it).
- **Placeholder lane** (#822's literal shape): `placeholder-checkpoint-recovery` now actually opens `allowNoAnchorImport`, but only under that proof. On main the reason string is passed while `allowNoAnchorImport` stays unset, so the recovery is dead code (`handlePlaceholderRecovery` passes only `noAnchorImportReason`; `reconcileSessionTail` checks `params.allowNoAnchorImport`).
- **checkpoint-missing lane**: the same predicate replaces the count===1 gate; the runtime-session-match and `bootstrappedAt` requirements are unchanged.
- **Cap lift under proof only**: with a proven non-anchoring frontier there is no real content to flood or contaminate, so the no-anchor import cap no longer blocks large id-less transcripts — the second half of #822, where a 647-message transcript stays unimported forever at cap 50. (Entry-id-bearing epochs already chunk-drain since #860; id-less ones still froze.) The import stays bounded by the transcript length; a frontier holding any real row keeps the cap and the #649 freeze exactly as today.
- The delivery-only guard now also covers the placeholder reason.

### Safety vs the #824 failure

An unrelated/rotated transcript can only be imported when the conversation provably holds nothing but injected metadata — there is no real history to contaminate. The regression test "does NOT import an unrelated transcript onto a placeholder-checkpoint conversation that already holds real anchoring rows (#824 contamination guard)" pins this: with real rows present the conversation freezes and the placeholder checkpoint is not advanced.

### Related reports

- #875 / #876 propose opening the same two lanes with a `count <= 3` gate. A count-based gate still admits conversations whose few rows are real history — the same contamination class that closed #824 — while the shape-proven gate admits only frontiers that contain no real content at all, with no magic count, and under that proof can also lift the cap safely.
- #874 looks like the same family (every conversation stuck at 1 DB message while the JSONL grows).

### Tests

5 new regression tests (all fail on main, pass here): multi-row placeholder recovery past the cap; the #824 contamination guard; delivery-only freeze parity; checkpoint-missing recovery past the cap; checkpoint-missing multi-row frontier. Full suite 1308/1308, typecheck clean. Patch changeset included.

### Production validation

Validated against production data (copy of the live DB + the real ~1MB session transcript, current main + this patch): a stuck conversation holding 2 metadata rows against a 647-message transcript recovers fully in one pass through the checkpoint-missing lane (2→649 messages, `capped=false`, checkpoint advanced only after full coverage). The build has been running in production since 2026-06-12.

Closes #822

### DIFF
--- a/.changeset/822-placeholder-noanchor-recovery.md
+++ b/.changeset/822-placeholder-noanchor-recovery.md
@@ -1,0 +1,9 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix a `#822` `afterTurn` deadlock where a conversation with an all-zero placeholder `conversation_bootstrap_state` row and a persisted frontier of only non-anchoring rows (e.g. one or more injected `Conversation info (untrusted metadata)` preambles) never re-imported its on-disk transcript. The placeholder-checkpoint-recovery reconcile ran without `allowNoAnchorImport`, so it imported 0 messages and never advanced the checkpoint; every turn then emitted the `did not cover the transcript frontier` warning, `assemble()` fell back to the raw transcript, and compaction never ran for that conversation.
+
+The placeholder recovery now opens a no-anchor import, but only when the persisted frontier is proven to hold no real conversation content — generalizing #837's single-injected-metadata-preamble check to a frontier composed entirely of non-anchoring injected-metadata rows (any count, via a bounded scan). A frontier with real anchoring rows conservatively freezes per #649's no-proof-no-advance guard, so an unrelated/rotated transcript can never be stitched onto real history (the failure mode that closed #824). For the proven-safe case the no-anchor import cap is lifted so a transcript larger than the cap recovers fully, bounded by the transcript length rather than left unbounded. The downstream import remains guarded by replay-overlap detection, the delivery-only block (now also applied to this lane), and the cross-conversation raw-id guard.
+
+The same generalized non-anchoring-frontier gate now also drives the afterTurn checkpoint-missing lane (#837), which previously required exactly one persisted frontier row: a conversation that accumulated several injected-metadata preambles before losing its checkpoint recovers there too, and the cap lift applies under the same proof.
--- a/src/transcript-reconciler.ts
+++ b/src/transcript-reconciler.ts
@@ -62,6 +62,12 @@ import { asRecord, formatDurationMs, isMissingFileError, safeString } from "./va
  */
 const FLUSH_LAG_ADOPTION_TAIL_WINDOW = 16;
 
+// Upper bound on the persisted-frontier scan used to decide placeholder-checkpoint
+// recovery eligibility (#822). A genuinely stuck placeholder frontier is a handful
+// of injected-metadata rows; a frontier larger than this conservatively freezes
+// rather than scan an arbitrarily large real DB tail (#649 no-proof-no-advance).
+const NON_ANCHORING_FRONTIER_SCAN_LIMIT = 32;
+
 export type BootstrapCheckpointFileState = {
   lastProcessedOffset: number;
   lastSeenSize: number;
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for the contribution, @rafaelreis-r. This is landed now.

PR #886, opened 2026-06-12
fix(engine): fail closed when the afterTurn transcript reconcile throws

rafaelreis-r:
### What

`afterTurn` initializes its reconcile result to an in-sync default (`hasOverlap: true`) and the catch around `reconcileTranscriptTailForAfterTurn` only logs the failure. A thrown reconcile therefore leaves the in-sync default in place: the live batch is persisted AND the checkpoint refreshes to EOF — silently advancing past transcript history that was never reconciled into the DB. Subsequent reads are append-only, so the skipped span becomes permanently unreachable.

The catch now reports the turn as not covered (`hasOverlap: false, importedMessages: 0`): batch persistence and checkpoint refresh are skipped for that turn. Nothing is lost — the transcript retains the turn and the next successful reconcile imports it. This matches the no-proof-no-advance doctrine (#649) and the fail-closed direction of the entry-id rework.

### Tests

New regression test (fails on main): the reconcile mock-throws → the live batch is NOT persisted, no checkpoint row is created, and the failure warn is logged. Full suite 1304/1304, typecheck clean. Patch changeset included.

### DIFF
--- a/.changeset/afterturn-reconcile-failclosed.md
+++ b/.changeset/afterturn-reconcile-failclosed.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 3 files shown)


jalehman:
Thanks for the contribution, @rafaelreis-r. This is landed now.

PR #890, opened 2026-06-14
fix(store): normalize injected user-turn decoration before computing message identity

### LINKED ISSUE #889: OpenClaw v2026.5.26+ session format change breaks afterTurn reconcile — all new sessions stuck with 1 DB message
## Description

Starting from OpenClaw v2026.5.26 (released 2026-05-27), all new sessions get permanently stuck: LCM stores only 1 message (the bootstrap message) while the JSONL transcript grows normally. The `afterTurn` reconcile loop fails on every turn with:

[lcm] afterTurn: transcript reconcile did not cover the transcript frontier; skipping afterTurn persistence to avoid creating a future anchor past unreconciled transcript history

This is **different from #875** — our issue is caused by a session JSONL format change in OpenClaw, not a timing/race condition.

## Root Cause

OpenClaw v2026.5.26 introduced **"cleaned user-turn persistence"** which changed how user messages are written to the session JSONL file.

### Before (working, ≤ v2026.5.25)

Session JSONL user messages included the full `Conversation info` metadata block:

{"chat_id":"user:ou_xxx","message_id":"om_xxx","sender":"John","timestamp":"...","inbound_event_kind":"user_request"}
Sender (untrusted metadata):

{"label":"John (ou_xxx)","id":"ou_xxx","name":"John"}
[message_id: om_xxx]
John: Hello, how are you?

### After (broken, ≥ v2026.5.26)

Session JSONL user messages are "cleaned" — the metadata block is stripped:

[message_id: om_xxx]
John: Hello, how are you?

### Why This Breaks Reconcile

1. **Bootstrap** reads the session JSONL → gets the "cleaned" message (no prefix) → stores in DB with identity hash A
2. **afterTurn** receives messages from OpenClaw runtime → these still have the `Conversation info` prefix → identity hash B
3. `reconcileSessionTail` compares hash A (DB) with hash B (transcript from afterTurn params) → **no match**
4. `hasOverlap=false`, `importedMessages=0` → `transcriptReconcileUnsafeToAdvance=true`
5. Persistence skipped → checkpoint never advances → **infinite loop**

### Evidence

| Conversation ID | Date | Status | DB Messages | Bootstrap Offset |
|----------------|------|--------|-------------|-----------------|
| 52 | 2026-05-22 | ✅ Working | 35 | 47800 |
| 127 | 2026-05-27 | ✅ Working | 150 | 1105293 |
| 151 | 2026-06-01 | ❌ Stuck | 1 | 0 |
| 162–305 | 2026-06-01+ | ❌ Stuck | 1 | 0 |

- All working conversations (≤ 127) have user messages **with** the `Conversation info` prefix in both DB and transcript
- All broken conversations (≥ 151) have user messages **without** the prefix in the session JSONL, but the bootstrap message in DB still has it (because `afterTurn` passes the prefixed version)

### Hash Comparison (conversation 305)

DB message (with prefix):     SHA-256 = aedaab2b089ab111...
Transcript (without prefix):  SHA-256 = c3e5c313a36aea...
→ No match → reconcile fails

### Affected Versions

- lossless-claw: 0.11.3 and 0.12.0 (both affected — the reconcile logic is the same)
- OpenClaw: ≥ 2026.5.26 (the version that changed session format)

### Suggested Fix

The reconcile logic needs to handle the prefix mismatch. Possible approaches:

1. **Strip the `Conversation info` block before computing identity hash** in both `ingestSingle` and `reconcileSessionTail` — normalize all user message content to the "cleaned" form
2. **Or**, when comparing identities during reconcile, try matching with and without the prefix
3. **Or**, have the bootstrap read messages from the same source as `afterTurn` (OpenClaw runtime params) instead of from the session JSONL file

### Workaround

None available. The issue affects every new session on OpenClaw ≥ 2026.5.26.

### Additional Context

- We confirmed that `@larksuite/openclaw-lark` and `@openclaw/feishu` both produce the same behavior (the prefix is added by OpenClaw core, not the channel plugin)
- The `isOpenClawRuntimeContextLeak` filter in 0.12.0 only checks assistant messages, not user messages
- 10+ conversations across 2 different users are affected in our production environment

gorkem2020:
## Summary

Fixes #889.

OpenClaw v2026.5.26+ "cleaned user-turn persistence" writes the transcript copy of a user message without the host-injected decoration, while the copy handed to `afterTurn` still carries it. Because message identity is computed over verbatim content, the two copies of the same turn hash differently, so reconcile never matches them.

This PR normalizes message content before the identity hash is computed, so the decorated and cleaned copies of a turn collapse to a single identity. This is option 1 from the issue's own suggested fixes.

## Root cause (confirming the issue analysis)

Exactly as described in the issue:

1. Bootstrap reads the cleaned session JSONL and stores the user turn with identity hash A (no decoration).
2. `afterTurn` receives the decorated copy from the OpenClaw runtime, identity hash B.
3. `reconcileSessionTail` compares A and B, finds no overlap, and treats the transcript tail as unreconciled.

The decoration that differs between the two copies includes the `Conversation info (untrusted metadata)` block, the `Sender (untrusted metadata)` block, the `[message_id: ...]` line, and, on the dashboard/webchat surface, an inline `[Wkdy YYYY-MM-DD HH:MM GMT+N]` timestamp.

## Two symptoms, one cause

The same identity mismatch surfaces differently by environment:

- **Stuck at 1 DB message** (the Lark/Feishu reports in this issue): reconcile skips persistence to avoid anchoring past unreconciled history, so the checkpoint never advances.
- **Duplicate rows** (our OpenClaw 2026.6.1 dashboard + Slack environment): both the decorated and the cleaned copy persist, producing two DB rows per user turn.

Both resolve once the two copies normalize to the same identity.

## Change

- `src/store/message-identity.ts`: add `normalizeIdentityContent()`, which strips the host-injected decoration listed above, and route both `buildMessageIdentityKey` and `buildMessageIdentityHash` through it.
- `src/message-signatures.ts`: `messageIdentity` delegates to `buildMessageIdentityKey`, centralizing normalization in one place rather than duplicating it at each call site.

Centralizing at the identity functions means every caller (bootstrap/ingest and afterTurn/reconcile) compares on the same normalized form, so the fix covers both paths the issue calls out. Only the value fed into the identity hash is normalized; the stored and displayed content (including the user-visible timestamp) is untouched.

## Verification

Deployed and live-verified in our environment (OpenClaw 2026.6.1, dashboard + Slack): user turns now persist exactly once, where previously each turn produced two rows, and reconcile advances normally. Verified across both channels with no session reset required. Happy to add a regression test for `normalizeIdentityContent` matching the repo's test conventions if that would help review.

## Relationship to prior work

The entry-id reconciliation rework ([#854](https://github.com/Martian-Engineering/lossless-claw/pull/854), [#866](https://github.com/Martian-Engineering/lossless-claw/pull/866)) reduced reconcile churn but does not cover this case, because its convergence still compares by content identity, which the decoration defeats. Normalizing content before hashing is complementary to that work.

### DIFF
--- a/src/message-signatures.ts
+++ b/src/message-signatures.ts
@@ -9,6 +9,7 @@ import type { CreateMessagePartInput } from "./store/conversation-store.js";
 import { extractToolResultIdForPairing } from "./tool-pairing.js";
 import { extractBootstrapMessageCandidate } from "./transcript.js";
 import { createHash } from "node:crypto";
+import { buildMessageIdentityKey } from "./store/message-identity.js";
 
 export function createBootstrapEntryHash(message: StoredMessage | null): string | null {
   if (!message) {
@@ -31,7 +32,7 @@ export function readBootstrapMessageFromJsonLine(line: string | null): AgentMess
 }
 
 export function messageIdentity(role: string, content: string): string {
-  return `${role}\u0000${content}`;
+  return buildMessageIdentityKey(role, content);
 }
 
 export function isBootstrapReplayCandidateMessage(message: AgentMessage): boolean {
--- a/src/store/message-identity.ts
+++ b/src/store/message-identity.ts
@@ -16,8 +16,8 @@ export function normalizeIdentityContent(content: string): string {
   if (typeof content !== "string") return content;
   let s = content;
   s = s.replace(/^\s*\[[A-Za-z]{3}\s+\d{4}-\d{2}-\d{2}[^\]]*GMT[^\]]*\]\s*/, "");
-  s = s.replace(/Conversation info \(untrusted metadata\):\s*\{[\s\S]*?\}\s*/g, "");
-  s = s.replace(/Sender \(untrusted metadata\):\s*\{[\s\S]*?\}\s*/g, "");
+  s = s.replace(/Conversation info \(untrusted metadata\):\s*(?:```[a-zA-Z]*\s*[\s\S]*?```|\{[\s\S]*?\})\s*/g, "");
+  s = s.replace(/Sender \(untrusted metadata\):\s*(?:```[a-zA-Z]*\s*[\s\S]*?```|\{[\s\S]*?\})\s*/g, "");
   s = s.replace(/\[message_id:[^\]]*\]\s*/g, "");
   return s.trim();
 }
--- a/src/store/message-identity.ts
+++ b/src/store/message-identity.ts
@@ -1,13 +1,37 @@
 import { createHash } from "node:crypto";
 
+/**
+ * Strip host-injected decoration before computing message identity.
+ * OpenClaw >=2026.5.26 "cleaned user-turn persistence"
+ * stores the transcript copy of a user turn RAW while the afterTurn/model-facing
+ * copy keeps its decoration (inline "[Wkdy YYYY-MM-DD HH:MM GMT+N]" timestamp on
+ * dashboard, "Conversation info (untrusted metadata): {...}" wrapper on Slack).
+ * Verbatim hashing makes the two copies distinct identities, so every user turn
+ * persists twice (lossless-claw #889; the #854/#866 entry-id rework does not cover
+ * this because its convergence still matches by content). Normalizing both copies
+ * to the same core text lets dedup collapse them to one row, while the
+ * user-visible timestamp stays intact (injected live at prompt-build).
+ */
+export function normalizeIdentityContent(content: string): string {
+  if (typeof content !== "string") return content;
+  let s = content;
+  s = s.replace(/^\s*\[[A-Za-z]{3}\s+\d{4}-\d{2}-\d{2}[^\]]*GMT[^\]]*\]\s*/, "");
+  s = s.replace(/Conversation info \(untrusted metadata\):\s*\{[\s\S]*?\}\s*/g, "");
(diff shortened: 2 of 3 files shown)


gorkem2020:
Updated to also strip the fenced json form of the injected metadata, not just the bare-brace form.

On current OpenClaw builds the inbound channel envelope (Conversation info / Sender untrusted-metadata) is rendered as a fenced code block. The original regex only matched a bare-brace object immediately after the label, so on fenced input it stripped nothing: the decorated copy hashed as raw, never collapsed against the cleaned transcript copy, and both rows persisted, reproducing the same double-persist this PR targets (still seen on Slack and Telegram).

The normalize now accepts either a fenced code block or a bare-brace object after each metadata label. Verified on real stored rows that the decorated and cleaned copies of the same turn produce identical identity hashes and collapse to one. Tests updated.

gorkem2020:
Superseded by #915. That PR fixes the same channel-turn duplication at the correct layer: the in-memory `messageIdentity` alignment signature used by after-turn dedup, rather than the persisted normalizer this PR touched (which #901 has since taken over and which deliberately keeps stable metadata keys for cross-chat distinction). Closing in favor of #915.

PR #855, opened 2026-06-10
fix(engine): use identity_hash for afterTurn dedup to prevent duplicate ingestion

bowenluo718:
## Summary

Fixes #922.

Fixes afterTurn duplicate/loss behavior by moving the non-transcript-covered heuristic dedup path to a lossless default and tightening externalized replay matching.

## Changes

- Use identity-hash based batch alignment for persisted afterTurn history.
- Ingest oversized no-overlap runtime batches when transcript reconcile has not proven coverage, instead of silently skipping live messages.
- Skip timestamp-proven stale no-overlap replays when the runtime batch matches already-persisted historical rows, without dropping legitimate repeated live turns with newer timestamps.
- Deduplicate provenance-backed externalized tool-result replays by comparing the stored sidecar payload to the incoming tool output.
- Preserve native-image externalization metadata and deduplicate anchored native-image replays only when metadata plus binary sidecar equality proves the match.
- Add focused regressions for no-overlap live preservation, stale timestamp-proven replay skip, repeated-content live preservation, externalized tool-output replay dedup, and native-image replay dedup.

## Validation

Local, from `/Volumes/LEXAR/repos/lossless-claw-pr855-fix`:

- `npm test -- test/engine-compaction.test.ts -t "oversized no-overlap|tool output reference replays|native image reference replays"`
- `npm test -- test/engine-compaction.test.ts`
- `npm test -- test/summary-store.test.ts`
- `npm test -- test/summary-store.test.ts test/engine-ingest.test.ts -t "externalizes native user image blocks|externalizes native assistant image blocks|externalizes native tool-result image blocks|externalizes native system image blocks"`
- `npm run typecheck`
- `git diff --check`

Remote CI is running on head `e1f6975`.

PR #899, opened 2026-06-16
fix(engine): reorder afterTurn to ingest live messages before transcr…

bowenluo718:
…ipt reconcile, with content-independent dedup keys

Live messages carry raw (unredacted) content while the transcript file holds redacted content. Previously, transcript reconcile ran first and ingested redacted messages, then the live-path dedup failed because SHA256(role + "\0"
+ content) differed between redacted and raw versions — causing duplicate rows in the messages table.

Reorder afterTurn so the live path ingests first, then transcript reconcile skips messages already persisted by matching on content-independent keys (responseId, toolCallId, toolUseId, timestamp) that are immune to redaction.


jalehman:
Thanks for the contribution. The duplicate-ingestion problem is real and in scope, but I do not think this PR is safe to review or merge as-is. It changes afterTurn to persist live messages before transcript reconcile has proven coverage, which weakens the current no-proof/no-advance guard that protects against anchoring past unreconciled transcript history. The current CI failures line up with that concern across rollover, no-anchor, capped frontier, bootstrap, fidelity, and transcript-entry-id tests, and the PR does not include regression tests or a changeset for the persistence behavior change.

I am closing this version so we can focus review on the narrower clean candidates in this area (#915, #920, and #855). A revised PR that preserves the reconcile-first safety boundary and has targeted regression coverage would be welcome.

PR #900, opened 2026-06-16
fix: keep heartbeat transcript reconciliation non-durable

jalehman:
## What

This PR keeps proven heartbeat/control transcript entries out of durable Lossless memory while still advancing transcript reconciliation checkpoints over those entries.

## Why

Heartbeat/control turns can appear in the OpenClaw transcript and runtime afterTurn batch, but they should not enter memory, summaries, search, or replay when they are proven control traffic. The important invariant is that Lossless may skip those entries durably, but it must still remember that the transcript frontier was reconciled so later real turns import normally and do not emit stale covered-frontier overlap warnings.

## Changes

- Add non-durable reconcile disposition
- Filter proven control transcript tails before import
- Preserve real mixed transcript deltas
- Preserve assistant-only durable prefixes
- Preserve unrecognized durable heartbeat-marked content
- Advance anchored placeholder checkpoints over control suffixes
- Treat `NO_REPLY` as exact-poll control traffic
- Keep rollover `NO_REPLY` checks turn-aware
- Require user-role exact heartbeat polls
- Preserve durable runtime `NO_REPLY` boundaries
- Keep full-read heartbeat filtering proof-based
- Propagate control disposition after mixed real/control reconcile
- Advance placeholder checkpoints over covered all-control slices
- Preserve placeholder durable history before heartbeat suffix filtering
- Filter runtime control messages before alignment
- Add regression coverage and changeset

## Testing

- `npx vitest run test/engine-after-turn.test.ts -t "full-read heartbeat reconcile preserves|full-read reconcile propagates"` passes
- `npx vitest run test/engine-after-turn.test.ts test/transcript-entry-id.test.ts test/ambiguous-rollover-rotation.test.ts` passes: 129 tests
- `npm run typecheck` passes
- `npm run build` passes
- `git diff --check` passes
- `/Users/phaedrus/Projects/prompts/skills/autoreview/scripts/autoreview --mode local` passes with no accepted/actionable findings

### DIFF
--- a/.changeset/heartbeat-control-reconcile.md
+++ b/.changeset/heartbeat-control-reconcile.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Keep heartbeat/control transcript turns non-durable while still advancing transcript reconciliation checkpoints.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -69,6 +69,7 @@ import { resolveBootstrapMaxTokens, trimBootstrapMessagesToBudget } from "./boot
 import {
   batchLooksLikeHeartbeatAckTurn,
   filterSyntheticHeartbeatMessages,
+  isAssistantControlAckContent,
   pruneHeartbeatOkTurns,
 } from "./heartbeat-filter.js";
 import { appendUncoveredVolatileLiveInputsWithinBudget, isVolatileLiveInputMessage, messageContentCoveredBySummary, resolveProtectedFreshTailAssembledIndexes } from "./live-coverage.js";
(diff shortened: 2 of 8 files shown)


jalehman:
Thanks for the careful work here. This looks like the right problem and the solution shape is promising, especially the explicit non-durable-control disposition and the durable-content guard coverage. I am deferring maintainer review while this remains draft: the branch is behind current main, now conflicts in test/ambiguous-rollover-rotation.test.ts, and CI is red on test/engine-bootstrap.test.ts. Please update onto current main, resolve the conflict, get full CI green, then run autoreview before asking for maintainer review.

jalehman:
Thanks for pushing this through a lot of difficult edge cases. I agree with the target invariant: heartbeat/control traffic should stay out of durable Lossless memory while transcript reconciliation still advances its checkpoint.

I’m going to close this draft rather than move it to autoreview as-is. The branch is now stale against current `main`, GitHub reports conflicts, CI is red, and the patch reaches across several high-risk durability paths at once: afterTurn runtime alignment, append-only transcript reconcile, full-read recovery, placeholder recovery, and rollover detection. The earlier review signal was that full-read control filtering still lacks the append-only proof boundary needed to guarantee exact durable `HEARTBEAT_OK`/`NO_REPLY` messages are not lost.

The concept is still valuable, but the safer next version should be smaller: isolate the proven append-only/control-suffix case first, keep full-read and placeholder recovery out unless they have their own proof, and rerun CI + autoreview after rebasing onto current `main`.

Issue #556, opened 2026-05-03
Harden conversation identity integrity for session rotation


ivalsaraj:

100yenadmin:
@ivalsaraj triage pass update: I marked this as priority:P1 because conversation identity integrity across session rotation is core memory safety. I do not see one merged PR that closes the whole concern; can you confirm the current failure mode that remains?

100yenadmin:
@ivalsaraj converting this umbrella into a focused implementation ticket.

Implementation scope: turn “conversation identity integrity for session rotation” into explicit invariants across stable `sessionKey`, changing `sessionId`, missing/pruned transcript files, `/new`, `/reset`, and active-vs-archived rows.

Acceptance criteria:
- regression tests for stable session-key rotation where old active rows cannot be silently reused for a new logical epoch
- active conversation lookup always prefers active rows and never reattaches archived rows
- no-anchor import behavior is bounded and diagnostic, especially for tiny new transcripts or missing old files
- current open candidate #757 is either fixed to cover the remaining P1 gaps or superseded by a narrower PR that covers these invariants

Keeping this open until the remaining concrete failure mode is confirmed or the invariant suite lands in a merged PR.

100yenadmin:
@ivalsaraj current read-only recheck says this umbrella should stay open as P1, not close from #757 alone.

What #757 fixed: merged commit `718e7998546aa2ba8ac1819b08dcf1ee04a5e6fd` covers the important subcase where a stable `sessionKey` resumes on a new runtime `sessionId` after the previously tracked transcript file was pruned/missing, preventing stale active-row replay across bootstrap/assemble/afterTurn.

What remains: current main still relies on transcript-file disappearance heuristics rather than explicit family/segment/runtime identity, and the adjacent dormant-DM shape in #769 is not covered with 95% confidence because the old real-history transcript can still exist while a tiny delivery/config transcript becomes active and no anchor is found.

Focused acceptance criteria for keeping this P1 open:
- explicit LCM family, segment, and runtime-session identity policy
- harden lifecycle/reconcile paths so session-file epoch changes do not silently split or reattach continuity
- regression for old long-lived transcript + new tiny system/delivery transcript + stable `sessionKey` + no-anchor path mismatch
- keep #757's pruned-transcript stale rollover behavior green

Reporter confirmation on the original repro is still useful, but current code plus adjacent live reports show the broader class is not fully closed.

100yenadmin:
[...]

Issue #489, opened 2026-04-24
Bootstrap/reconcile can re-import huge historical messages and create multi-million-token conversations


Adam-Researchh:
## Summary

`lossless-claw` 0.9.2 can re-import already-ingested historical transcript messages during bootstrap/reconcile, creating multi-million-token conversations. In the observed case, the same old bootstrap/media messages were appended again with new sequence numbers and current timestamps, causing later assembly to approach the token budget and provider requests to fail with `context_length_exceeded`.

## Environment

- OpenClaw: 2026.4.22
- lossless-claw: 0.9.2
- Runtime: macOS arm64, Node 25.8.0
- DB: SQLite `lcm.db`
- Session type: main long-lived agent session with channel sessions and TUI sessions
- `proactiveThresholdCompactionMode=deferred`
- summary model: observed both local fallback model and Haiku after config change

## What happened

A long-lived active conversation (`agent:main:main`) accumulated duplicate historical messages after bootstrap/reconcile. The old conversation was later archived/replaced by a new active conversation, but the archived conversation remained extremely large.

Depersonalized DB snapshot:

```text
conversation_id  session_key       active  msgs  toks     max_seq
146              agent:main:main   0       6776  4168945  6846
177              agent:main:main   1       11    15622    11
```

The largest duplicated messages were not new user turns; they were historical payloads reinserted later:

```text
conversation_id  seq   role  token_count  created_at           preview
146              5046  user  59845        2026-04-23 20:05:28  [Bootstrap pending] Please read BOOTSTRAP.md...
146              5047  user  59845        2026-04-23 20:05:33  [Bootstrap pending] Please read BOOTSTRAP.md...
146              6258  user  59845        2026-04-24 22:05:10  [Bootstrap pending] Please read BOOTSTRAP.md...
146              5961  user  45472        2026-04-24 21:21:01  [media attached: ...jpg ...]
146              5962  user  45472        2026-04-24 21:21:10  [media attached: ...jpg ...]
146              6789  user  45472        2026-04-24 22:05:10  [media attached: ...jpg ...]
146              5684  user  38004        2026-04-24 02:07:55  [media attached: ...jpg ...]
146              5707  user  38004        2026-04-24 02:10:16  [media attached: ...jpg ...]
146              6610  user  38004        2026-04-24 22:05:10  [media attached: ...jpg ...]
```

Gateway logs showed this immediately before context overflow recovery:

```text
[...]

PR #901, opened 2026-06-16
fix: canonicalize OpenClaw inbound metadata for message identity

SYU8384:
## Summary

This fixes an LCM context-continuity failure when OpenClaw user messages include injected inbound metadata blocks such as `Conversation info (untrusted metadata)` and `Sender (untrusted metadata)`.

LCM previously used raw message content for message identity and bootstrap checkpoint hashes. OpenClaw metadata can contain volatile fields such as message IDs and timestamps, so the same logical user text could hash differently between the persisted DB frontier and the session JSONL frontier. That could make append-only bootstrap/reconcile miss the frontier and leave a conversation stuck or force slower reconciliation.

## Concrete impact

In real OpenClaw + LCM sessions, this can show up as apparent context loss after resuming a conversation. The same logical user turn can carry different injected OpenClaw message IDs or timestamps, so LCM may fail to recognize the persisted conversation frontier. In the observed failure mode, the context from the last conversation was not carried forward even though the raw transcript content still existed.

Depending on the import path, this can leave `last_processed_offset` stuck, make append-only bootstrap miss the checkpoint, or force slower reconciliation/re-import behavior.

This patch canonicalizes only the identity/bootstrap hash input for recognized OpenClaw user-role metadata preambles. Stored message content remains raw and lossless. The canonicalizer also requires OpenClaw-shaped JSON object keys before stripping, so user-authored fenced JSON after similar headings remains part of identity.

## Related reports

I did not find an exact existing issue for this metadata-hash failure mode.

Related context:
- Martian-Engineering/lossless-claw#489: bootstrap/reconcile identity drift and duplicate historical imports.
- Martian-Engineering/lossless-claw#556: broader conversation identity integrity.
- openclaw/openclaw#93553: prior OpenClaw-side attempt closed because it stripped canonical persisted user messages. This PR uses the LCM-side boundary instead.

## Changes

- Add `canonicalizeOpenClawInboundMetadataIdentityContent(...)`.
- Require recognized OpenClaw metadata object keys before canonicalizing a preamble.
- Use it for:
  - message identity keys/hashes
  - bootstrap entry hashes
- Preserve raw `messages.content` exactly.
- Add regression coverage for:
  - volatile OpenClaw metadata hashing to the same identity
  - ordinary user-authored similar text not being stripped
  - raw stored content remaining lossless
  - append-only bootstrap advancing when only the metadata envelope changes

## Verification

- `npm test -- message-identity.test.ts message-signatures.test.ts engine-bootstrap.test.ts`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/.changeset/openclaw-inbound-metadata-identity.md
+++ b/.changeset/openclaw-inbound-metadata-identity.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Canonicalize OpenClaw inbound metadata when computing message identity and bootstrap hashes so volatile message metadata does not stall context continuity while raw transcript content stays lossless.
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -1,5 +1,7 @@
 import type { DatabaseSync } from "node:sqlite";
+import { createHash } from "node:crypto";
 import { getLcmDbFeatures } from "./features.js";
+import { canonicalizeOpenClawInboundMetadataIdentityContent } from "../openclaw-inbound-metadata.js";
 import { buildMessageIdentityHash } from "../store/message-identity.js";
 import { parseUtcTimestampOrNull } from "../store/parse-utc-timestamp.js";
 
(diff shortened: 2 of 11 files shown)


jalehman:
Thanks for the contribution, @SYU8384. This is landed now.

PR #911, opened 2026-06-18
Warn when LCM installs are pinned to exact versions

### LINKED ISSUE #910: Exact-version LCM installs stop following OpenClaw plugin update sync
## Summary

After a normal OpenClaw update, lossless-claw did not advance from `0.12.0` to the currently published `0.13.1`. The local behavior appears to be caused by LCM being tracked as an exact npm spec (`@martian-engineering/lossless-claw@0.12.0`) rather than a moving stable/latest spec.

This is probably consistent with OpenClaw's documented plugin updater semantics: exact versions and explicit tags are not rewritten during update sync. But from an LCM operator perspective this is a regression from the historical expectation that "OpenClaw update" also keeps LCM current enough to pick up memory-continuity fixes.

## Why this matters

LCM is not just a convenience plugin here; it is the selected long-run context engine. Missing a current LCM release can directly affect memory continuity after OpenClaw restarts/updates.

In this case, the host remained on `lossless-claw@0.12.0` while `0.13.1` was available. The next restored main session had an empty/new active LCM conversation instead of inheriting the prior summary chain. The visible assistant correctly said it only saw the new "you there" turn, the gateway interruption notice, and `?`.

The `0.13.1` release appears relevant to that class of issue because it includes fresh-transcript rollover / split-memory repair work, but the normal OpenClaw update path did not pick it up because the install was pinned.

## Environment

- OpenClaw: `2026.6.8 (844f405)`
- Install kind: global npm package, macOS LaunchAgent gateway
- lossless-claw active version: `0.12.0`
- latest published lossless-claw on npm: `0.13.1`
- selected context engine: `plugins.slots.contextEngine = "lossless-claw"`

The local LCM config itself was otherwise healthy:

- `contextThreshold: 0.9`
- `leafChunkTokens: 20000`
- `summaryModel: gpt-5.4-mini`
- fallback `gpt-5.4`
- `stubLargeToolPayloads: true`

## Evidence

Installed/runtime version:

```bash
openclaw --version
# OpenClaw 2026.6.8 (844f405)

openclaw plugins inspect lossless-claw --json
# plugin.version: 0.12.0
# install.spec: @martian-engineering/lossless-claw@0.12.0
# install.resolvedSpec: @martian-engineering/lossless-claw@0.12.0
```

Npm latest:

```bash
npm view @martian-engineering/lossless-claw version --json
# "0.13.1"
```

Plain plugin update treats the exact spec as current:

```bash
openclaw plugins update lossless-claw --dry-run
# lossless-claw is up to date (0.12.0).
```

Explicit moving spec would update:

```bash
openclaw plugins update @martian-engineering/lossless-claw@latest --dry-run
# Would update lossless-claw: 0.12.0 -> 0.13.1.
```

Core OpenClaw dry run does include plugin sync, but that sync follows the recorded plugin spec:

```bash
openclaw update --dry-run --json
# actions include:
# - Run plugin update sync after core update
```

OpenClaw docs also match this interpretation:

- `openclaw plugins update <id-or-npm-spec>` is the routine upgrade path for tracked npm plugins.
- npm installs without an exact version follow stable/latest compatibility resolution.
- exact versions and explicit tags remain strict / are not rewritten by channel sync.

## Separate but related noise

This host also still has the known OpenClaw stale metadata warning:

```text
Left plugin install index in place because shared SQLite state has conflicting plugin install metadata for: acpx, codex, gbrain, lossless-claw, tokenjuice
```

For this issue, that does not appear to be the decisive blocker. The active runtime/package metadata says `0.12.0`, and an explicit `@latest` dry run correctly detects `0.13.1`.

## Expected behavior

If LCM is installed as the selected context engine and the operator has not intentionally pinned a release for rollback/compliance, routine OpenClaw update/plugin sync should keep LCM on the latest compatible stable release.

At minimum, LCM should make it obvious when it is pinned in a way that prevents this.

## Suggested fix shape

One or more of:

1. Update LCM install/upgrade docs and release snippets so routine installs use a moving stable spec, e.g.

   ```bash
   openclaw plugins update @martian-engineering/lossless-claw@latest
   ```

   or an equivalent unpinned install/update form, reserving exact versions for explicit rollback/pin scenarios.

2. Add a `lossless doctor` / status warning:

   - detect `install.spec` matching `@martian-engineering/lossless-claw@<exact semver>`
   - compare against npm latest compatible stable
   - if newer exists and `lossless-claw` is the selected `contextEngine`, warn that OpenClaw plugin sync will not advance LCM because it is exact-pinned
   - print the safe repair command:

     ```bash
     openclaw plugins update @martian-engineering/lossless-claw@latest
     openclaw gateway restart
     ```

3. Optionally provide a narrow repair command, for example:

   ```bash
   lossless doctor repair --unpin-update-track
   ```

   or an equivalent OpenClaw-compatible operation that changes the tracked spec to `@latest` without touching unrelated OpenClaw doctor repairs.

4. If a published LCM release flow currently recommends exact-version installation after upgrades, add a follow-up step that restores the moving update track after verification.

## Non-goal

This issue is not asking OpenClaw to override deliberate exact pins. The ask is for lossless-claw to prevent accidental permanent pins in the common long-run-memory deployment path, or at least surface them loudly before operators miss continuity fixes.

100yenadmin:
## Summary

This PR adds an LCM-side guardrail for the update-track failure described in #910.

Historically, LCM followed OpenClaw plugin update sync during `openclaw update`. After an exact-version install or force reinstall such as:

```bash
openclaw plugins install @martian-engineering/lossless-claw@0.12.0 --force
```

OpenClaw records the exact package spec. From that point forward, plugin update sync can correctly-but-surprisingly keep resolving the same exact version, so `openclaw plugins update lossless-claw --dry-run` can report “up to date” even when npm has a newer LCM release. Explicitly moving the install track back to the npm dist-tag works:

```bash
openclaw plugins update @martian-engineering/lossless-claw@latest
```

## Long-Term Shape

The intended long-term default for routine LCM installs should be the moving stable track:

```bash
openclaw plugins install @martian-engineering/lossless-claw@latest
```

Exact versions are still useful, but they should mean “pin this version for rollback, reproducible canary testing, or incident containment.” They should not be the documented steady-state install shape for users who expect `openclaw update` to keep LCM current.

This PR treats `@latest` as the correct durable LCM-side answer and makes exact pins visible when they are likely to surprise users.

## What Changed

- `/lossless` status now warns when OpenClaw metadata shows `lossless-claw` installed from an exact npm package spec such as `@martian-engineering/lossless-claw@0.12.0`.
- `/lossless doctor` shows the same warning, including the repair command:

```bash
openclaw plugins update @martian-engineering/lossless-claw@latest
```

- Command handling now receives the plugin registration OpenClaw config snapshot as a fallback metadata source, so the warning can still fire when the command context itself does not include install metadata.
- README and configuration docs now document `@latest` as the routine install/update track and frame exact versions as rollback/canary pins.
- Regression tests cover exact-version warnings, moving `@latest` silence, and fallback config detection.

## Correctness Boundary

This is the right LCM-side fix if OpenClaw’s current behavior is intentional: exact npm specs are pins, and moving specs/dist-tags are update tracks.

The deeper platform-level fix, if desired, belongs in OpenClaw: make plugin update track state first-class, separate “desired spec” from “resolved installed version,” or have OpenClaw warn/repair exact official-plugin specs during `openclaw update` / `openclaw doctor`. LCM should not silently rewrite its own plugin install record from inside the plugin runtime.

## User Impact

- Existing users pinned to an exact LCM version get a visible warning instead of silently missing future LCM releases.
- New installs following the docs land on the moving stable track.
- Users who intentionally pin for rollback or canary testing can keep doing that; the warning is informational and includes the command to return to the stable track.

## Test Plan

- `npm exec vitest -- run test/lcm-command.test.ts`
- `npm run typecheck`
- `npm run build`
- GitHub CI: `test`, `plugin-inspector`, `smoke-latest-openclaw`

Fixes #910

### DIFF
--- a/.changeset/tall-update-track-warning.md
+++ b/.changeset/tall-update-track-warning.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Warn from `/lossless` status and doctor when OpenClaw metadata shows lossless-claw installed from an exact npm version, and document `@latest` as the routine install/update track.
--- a/README.md
+++ b/README.md
@@ -73,13 +73,19 @@ The bundled skill focuses on configuration, diagnostics, architecture, and recal
 Use OpenClaw's plugin installer (recommended):
 
 ```bash
-openclaw plugins install @martian-engineering/lossless-claw
+openclaw plugins install @martian-engineering/lossless-claw@latest
 ```
 
 If you're running from a local OpenClaw checkout, use:
 
 ```bash
-pnpm openclaw plugins install @martian-engineering/lossless-claw
+pnpm openclaw plugins install @martian-engineering/lossless-claw@latest
+```
+
+Use exact versions only for rollback or reproducible canary testing. OpenClaw records an exact install spec such as `@martian-engineering/lossless-claw@0.12.0` as a pinned update track, so OpenClaw plugin update sync will keep that version until you move back to the stable track:
+
+```bash
+openclaw plugins update @martian-engineering/lossless-claw@latest
 ```
 
 For local plugin development, build your working copy first, then link it instead of copying files:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -141,13 +141,19 @@ Notes on the example:
 Install with OpenClaw's plugin installer:
 
 ```bash
-openclaw plugins install @martian-engineering/lossless-claw
+openclaw plugins install @martian-engineering/lossless-claw@latest
 ```
 
 If you are running from a local OpenClaw checkout:
 
 ```bash
-pnpm openclaw plugins install @martian-engineering/lossless-claw
+pnpm openclaw plugins install @martian-engineering/lossless-claw@latest
+```
+
+Use exact versions only for rollback or reproducible canary testing. OpenClaw treats an exact install spec such as `@martian-engineering/lossless-claw@0.12.0` as pinned, so plugin update sync will not follow newer LCM releases until you return to the moving track:
+
+```bash
+openclaw plugins update @martian-engineering/lossless-claw@latest
 ```
 
 For local plugin development, link a working copy:
--- a/src/plugin/index.ts
+++ b/src/plugin/index.ts
@@ -1621,6 +1621,7 @@ function wirePluginHandlers(
   api: ContextEngineCapableOpenClawPluginApi,
   deps: LcmDependencies,
   shared: SharedLcmInit,
+  openClawConfig?: unknown,
 ): void {
   api.on("before_reset", async (event, ctx) => {
     await (await shared.waitForEngine()).handleBeforeReset({
@@ -1696,6 +1697,7 @@ function wirePluginHandlers(
     createLcmCommand({
       db: shared.waitForDatabase,
       config: deps.config,
+      openClawConfig,
       deps,
       getLcm: shared.waitForEngine,
     }),
@@ -1906,7 +1908,7 @@ const lcmPlugin = {
           logStartupMaintenanceSchedulingError,
         );
       }
-      wirePluginHandlers(api, deps, existingInit);
+      wirePluginHandlers(api, deps, existingInit, registrationConfig.openClawConfig);
       return;
     }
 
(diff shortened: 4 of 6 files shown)


jalehman:

PR #915, opened 2026-06-18
fix(dedup): align OpenClaw decorated and bare inbound copies by body

### LINKED ISSUE #912: OpenClaw inbound dedup: identity normalization covers only 2 of ~10 untrusted-metadata block types, so bare vs decorated copies persist as duplicate rows
## Summary

For OpenClaw channel turns, each inbound message is persisted as two `messages` rows: a bare transcript copy (the raw body) and a decorated model-facing copy (injected untrusted-metadata blocks plus the body). These should collapse to one row. They do not, so the assembler (which replays stored rows verbatim) renders the turn to the model two or more times across Slack, Telegram, and webchat.

The dedup that decides whether to persist the live decorated copy is the in-memory signature `messageIdentity(role, content)` (used by the after-turn covered-frontier alignment in `batch-dedup.ts`, and by the replay dedup). It compares the raw content: `role + NUL + content`. The decorated runtime copy's raw content never equals the bare transcript row's, so the alignment treats it as an unflushed new message and ingests it as a second row (with `transcript_entry_id` NULL).

## Reproduction (stock v0.13.1)

Send a message on any channel (Slack, Telegram, webchat). In `lcm.db`, the turn appears as two `user` rows: one bare (with a transcript entry id), one decorated (with `transcript_entry_id` NULL). On Telegram it is especially visible because the `Conversation context` block appears once there is conversation history, so most turns after the first persist as a duplicate pair. Sanitized shape of the decorated row:

````
```json
{ "chat_id": "<id>", "sender_id": "<id>", "sender": "<name>", "timestamp": "<ts>", "inbound_event_kind": "user_request" }
```

Sender (untrusted metadata):
```json
{ "label": "<name> (<id>)", "id": "<id>", "name": "<name>" }
```

Conversation context (untrusted, chronological, selected for current message):
#<id> <ts> <name>: <prior message text>
#<id> <ts> <name>: <prior message text>

<the actual user body>
````

The bare copy is just `<the actual user body>`. Webchat shows a related flavor: the decorated copy is prefixed with a `[Wkdy YYYY-MM-DD HH:MM GMT+N]` timestamp that the bare copy lacks.

## Why this is distinct from #901

#901 (`canonicalizeOpenClawInboundMetadataIdentityContent`) normalizes the input to the persisted `identity_hash`, and it deliberately keeps the stable metadata keys (`chat_id`, `sender`, and so on) so that different chats with the same text stay distinct. That behavior is correct and should stay.

But the persisted hash is not what gates this dedup. The after-turn path that decides whether to ingest the live runtime copy aligns it against the persisted transcript tail using the in-memory `messageIdentity(role, content)`, which is built on the raw content. So even with #901 in place, the decorated raw content does not equal the bare raw content, the alignment misses, and the duplicate row persists. Fixing this in the persisted canonicalizer would either be ineffective (the gate is the in-memory signature, not the stored hash) or would require stripping the stable keys, which would regress #901's cross-chat distinction.

## Proposed fix

Make the in-memory identity decoration-aware, without touching the persisted hash.

In `messageIdentity(role, content)`, for user content, reduce the content to its bare body before forming the signature, using a structural extraction rather than a per-label list (the label family is large and grows):

1. Strip a leading `[Wkdy YYYY-MM-DD HH:MM GMT+N]` timestamp.
2. Split the content on blank lines and drop any segment whose first line matches an untrusted-metadata header (`^.{1,80}\(untrusted[^)]*\):$`), keeping the remaining segments (the body).

This is the in-memory dedup and alignment signature only. The persisted `identity_hash` and #901's canonicalizer are unchanged, so cross-chat distinction is preserved. The extraction is a no-op for already-bare content and for non-user roles, so only the decorated user copy is affected. With it, the after-turn alignment recognizes the decorated runtime copy as the bare transcript row that reconcile already persisted and drops it, so exactly one row persists per turn.

Verified against real rows: it collapses Slack, Telegram (including the `Conversation context` block), and webchat decorated copies onto their bare counterparts; the full test suite passes; #901's identity tests are unaffected.

This supersedes #890, which normalized only the bare-brace `Conversation info` form in the persisted normalizer and predated #901.

## Environment

- lossless-claw v0.13.1 (stock), OpenClaw 2026.6.1, Linux, Node 24

gorkem2020:
Fixes #912.

OpenClaw delivers each channel turn in two forms: a live model-facing copy decorated with injected untrusted-metadata blocks (Conversation info, Sender, Conversation context, and the rest of the family) and the bare transcript copy. After-turn dedup decides whether to persist the live runtime copy by aligning it against the persisted transcript tail with the in-memory signature `messageIdentity(role, content)`, which compared the raw content. The decorated copy's raw content never matched the bare transcript row, so the alignment treated it as unflushed and ingested it as a second row (`transcript_entry_id` NULL). Because the assembler replays stored rows verbatim, every channel turn rendered to the model twice across Slack, Telegram, and webchat.

This makes `messageIdentity` strip the inbound decoration for user content via `extractOpenClawInboundBody` before forming the in-memory signature, so the decorated and bare forms align and only one row persists.

Key points:

- In-memory dedup signature only. The persisted identity hash (`buildMessageIdentityHash`, and `canonicalizeOpenClawInboundMetadataIdentityContent` from #901) is unchanged, so #901's cross-chat distinction is preserved and its identity tests still pass.
- `extractOpenClawInboundBody` is structural: strip a leading channel timestamp, then drop untrusted-metadata-header segments. It is robust to the full block family and to new block types, and it is a no-op for already-bare content and for non-user roles, so only the decorated user copy is affected.
- Tests: `messageIdentity` normalization units, an `alignRuntimeBatchAgainstCoveredFrontier` integration test, and `extractOpenClawInboundBody` units. Full suite green.

Verified against real stored rows from a live deployment: the change collapses Slack, Telegram (including the `Conversation context` block), and webchat decorated copies onto their bare counterparts, and post-deploy channel turns persist a single row per turn.

Supersedes #890, which normalized only the bare-brace `Conversation info` form in the persisted normalizer and predated #901.

### DIFF
--- a/src/message-signatures.ts
+++ b/src/message-signatures.ts
@@ -5,7 +5,10 @@
  */
 import { buildMessageParts, toStoredMessage, type StoredMessage } from "./message-content.js";
 import type { AgentMessage } from "./openclaw-bridge.js";
-import { canonicalizeOpenClawInboundMetadataIdentityContent } from "./openclaw-inbound-metadata.js";
+import {
+  canonicalizeOpenClawInboundMetadataIdentityContent,
+  extractOpenClawInboundBody,
+} from "./openclaw-inbound-metadata.js";
 import type { CreateMessagePartInput } from "./store/conversation-store.js";
 import { extractToolResultIdForPairing } from "./tool-pairing.js";
 import { extractBootstrapMessageCandidate } from "./transcript.js";
@@ -36,7 +39,18 @@ export function readBootstrapMessageFromJsonLine(line: string | null): AgentMess
 }
 
 export function messageIdentity(role: string, content: string): string {
-  return `${role}\u0000${content}`;
+  // In-memory identity for dedup / replay detection / after-turn alignment.
+  // OpenClaw delivers each channel turn in two forms - the live model-facing
+  // copy decorated with injected untrusted-metadata blocks, and the bare
+  // transcript copy - which are the SAME logical message and must align so the
+  // covered-frontier dedup does not persist a duplicate row (issue #912).
+  // Strip the inbound decoration for user content before forming the identity.
+  // This is NOT the persisted identity hash (buildMessageIdentityHash /
+  // canonicalizeOpenClawInboundMetadataIdentityContent / #901); the chat-aware
+  // persisted identity is unchanged, so #901 cross-chat distinction holds.
+  // extractOpenClawInboundBody is a no-op for already-bare content and for
+  // non-user roles, so only the decorated user copy is affected.
+  return `${role}\u0000${extractOpenClawInboundBody(role, content)}`;
 }
 
 export function isBootstrapReplayCandidateMessage(message: AgentMessage): boolean {
--- a/src/openclaw-inbound-metadata.ts
+++ b/src/openclaw-inbound-metadata.ts
@@ -194,3 +194,43 @@ function getKnownKeysForHeading(heading: string): Set<string> | undefined {
   }
   return undefined;
 }
+
+const OPENCLAW_INBOUND_TIMESTAMP_PREFIX_RE =
+  /^\s*\[[A-Za-z]{3}\s+\d{4}-\d{2}-\d{2}[^\]]*GMT[^\]]*\]\s*/;
+
+const OPENCLAW_UNTRUSTED_BLOCK_HEADER_RE = /^.{1,80}\(untrusted[^)]*\):\s*$/;
+
+/**
+ * Extract the raw user body from an OpenClaw inbound message by dropping every
+ * injected untrusted-metadata block (Conversation info, Sender, Conversation
+ * context, Reply chain/target, Forwarded message context, Thread starter,
+ * Location, ... — the whole family) plus any leading channel timestamp.
+ *
+ * Unlike `canonicalizeOpenClawInboundMetadataIdentityContent` (which preserves
+ * stable metadata so two genuinely different messages keep distinct identity
+ * hashes), this reduces an inbound turn to its body alone. It is used ONLY to
+ * recognize that the bare transcript copy and the decorated model-facing copy
+ * are the same logical turn during flush-lag adoption — it never feeds the
+ * message identity hash, so the chat-aware identity design is unaffected.
+ *
+ * Generic by construction: the builder emits each block as a
+ * "<Label> (untrusted...):" header followed by a fenced JSON body or a list,
+ * with blocks and the body separated by blank lines. Dropping the
+ * header-led segments survives new block types. The one edge is a list entry
+ * that itself contains a blank line; that is safe-by-default (the bodies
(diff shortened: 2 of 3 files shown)


100yenadmin:
Maintainer review: passed for merge-authority review.

**Summary**
I now rate this PR as 10/10 mergeable for the #912 P1 path. I did not merge it.

**What changed during review**
The first adversarial pass found a real false-dedup blocker: the original body-only extractor could collapse user-authored metadata-looking text, timestamp-looking text, or leading whitespace before in-memory alignment. I pushed follow-up commit `9b33a707c6759f576ff2efdb89c113690b7b1637` to make extraction fail closed unless a valid OpenClaw inbound metadata preamble is present, and to preserve user-authored metadata-looking bodies even inside a real decorated OpenClaw turn.

**Evidence checked**
- Current head: `9b33a707c6759f576ff2efdb89c113690b7b1637`
- GitHub merge state: `MERGEABLE` / `CLEAN`
- GitHub checks at current head: `test`, `plugin-inspector`, and `smoke-latest-openclaw` all passed
- Review threads: none unresolved
- Local focused validation from `/Volumes/LEXAR/repos/lossless-claw-pr915-fix`: `npm test -- test/openclaw-inbound-dedup.test.ts` passed 20 tests, `npm run typecheck` passed, and `git diff --check` was clean
- Adversarial review: four-lane initial sweep found the extractor false-positive blocker; after repair, blockers-only rereview cleared the decorated-body regression at current head

**Next step**
@jalehman this is ready for merge-authority review when convenient. Keeping issue #912 open until the PR is merged and the post-merge/release check confirms decorated and bare inbound copies persist as one row.

jalehman:
I ran autoreview on this PR and it found a real remaining blocker. A full OpenClaw-decorated user message can itself be quoted by the user, and that quoted text is currently indistinguishable from an injected decorated runtime copy. Because `messageIdentity` would collapse both to the trailing body, this can still false-dedup user-authored content.

I locally tested a hardening commit that addresses several narrower extractor false positives, but I did not push it because the remaining issue needs a trusted host-side marker or contract from OpenClaw rather than another text-shape heuristic. Focused local proof on that partial hardening was green (`test/openclaw-inbound-dedup.test.ts`, `test/message-identity.test.ts`, typecheck, diff check), but final autoreview still reported the quoted-decorated-message blocker.

gorkem2020:
Closing in favor of #927, which fixes #912 with a structural approach instead of identity normalization.

The body-strip approach here cannot safely tell host-added decoration from user-typed timestamp or metadata looking text without a host-side trusted marker, which was the open blocker. #927 sidesteps that: it recognizes the decorated copy as the bare body's line-aligned trailing text (fail-closed containment) and gates the collapse on recognized OpenClaw decoration, so it never collapses arbitrary user text and needs no host-side marker. Thanks for the review here, it directly shaped that design.

PR #923, opened 2026-06-20
fix: recover afterTurn from transient transcript ENOENT

### LINKED ISSUE #916: v0.13.1: lcm plugin self-reload triggers persistent afterTurn skip with false ENOENT for existing session file
## Version
- lossless-claw `0.13.1` (installed via `openclaw plugins install '@martian-engineering/lossless-claw@0.13.1' --force`)
- openclaw `2026.6.8` (844f405)
- Node 26.0.0 on macOS (Apple Silicon)

## Summary
After upgrading to v0.13.1 and running fine for ~3 hours, the lcm plugin appears to **internally self-reload** (without a gateway restart) and afterwards enters a persistent state where `afterTurn` persistence is silently skipped. Bridge workers continue to respond but their assistant turns never reach `lcm.db.messages`. Only a full `launchctl kickstart` of the gateway recovers it.

### Gateway lifecycle (no restarts during the stall window)
```
http server listening  18:52:54   (manual kickstart)
http server listening  22:36:00   (manual kickstart after stall observed)
```
No `received SIGTERM` / `received SIGUSR1` / `http server listening` events between 18:52:54 and 22:36:00.

### lcm plugin restarted itself twice within 1 minute
```
19:55:34  [lcm] Legacy state migration warnings: ...
19:55:36  [lcm] Engine initialized for db=~/.openclaw/lcm.db duration=1242ms
19:55:36  [lcm] Plugin loaded (enabled=true, threshold=0.75, ...)
19:55:36  [lcm] startup totalTokens recovery updated 4 session entries
19:55:38  [lcm] afterTurn: session file missing; skipping transcript reconcile full reread; could not stat/read transcript; allowing live afterTurn persistence and seeding...
19:55:38  [lcm] afterTurn: bootstrap checkpoint refresh failed for session=b0cf0fbd-... sessionKey=agent:mini-bal:main: ENOENT: no such file or directory
19:55:38  [lcm] auto-rotate: phase=runtime action=warn sessionId=b0cf0fbd-... sessionFile=~/.openclaw/agents/mini-bal/sessions/b0cf0fbd-...

19:56:31  [model-fetch] error provider=claude-bridge ... AbortError
19:56:31  [lcm] afterTurn: session file missing; skipping transcript reconcile full reread; preserving existing checkpoint (offset=0) conversation=8511 reason=append-only...
19:56:31  [lcm] dedup: prefix-mismatch, storedCount=1 batchLen=1, no overlap found — ingesting full batch
19:56:31  [lcm] afterTurn: bootstrap checkpoint refresh failed for session=b0cf0fbd-...: ENOENT
19:56:31  embedded run failover decision (LLM request failed.)

19:56:35  [lcm] Legacy state migration warnings: ...
19:56:35  [lcm] Engine initialized duration=31ms
19:56:35  [lcm] Plugin loaded
```

After 19:56:35 the lcm plugin emits **zero further log lines** until the next manual gateway kickstart.

### The "missing" session file actually exists
```
$ ls -la ~/.openclaw/agents/mini-bal/sessions/b0cf0fbd-b570-459b-a5ad-4fd52a488c7d.jsonl
-rw-------  1 user staff  17685 Jun 18 22:27 b0cf0fbd-...jsonl
```
The file was being written continuously (workers were still processing). Yet lcm logs ENOENT.

### Result: silent persistence skip
`messages.MAX(created_at)` for the active DM conversation froze at the moment of the first self-reload:
- DM session conversation (active=1): `MAX(created_at) = 13:28:02` (lasted 9 hours)
- main session conversation: `MAX(created_at) = 10:56:36` (lasted 12 hours)
- Bridge workers kept handling tasks and writing to session `.jsonl` files; lcm.db.messages had no new rows

After manual `launchctl kickstart -k gui/$UID/ai.openclaw.gateway`, the `startup totalTokens recovery` re-ran and a test message inserted successfully, persistence resumed.

## Related (already-fixed via v0.13.1)
PRs #903, #906, #901 covered the "rollover archive + empty replacement" case observed earlier the same day. v0.13.1 successfully recovered that case via `placeholder-checkpoint-recovery`. The behavior reported here is a **different** symptom and appears not to be covered.

## Hypothesis
1. Something inside lcm triggers `Engine initialized` to run again without gateway restart (a plugin-level self-reload).
2. During this internal re-init, `stat()` on the active session jsonl returns ENOENT (false negative — perhaps a path resolution / cache-invalidation race, or a check before agents.runtime publishes session file path).
3. The ENOENT path takes the "preserving existing checkpoint (offset=0)" branch, which (combined with `auto-rotate action=warn`) seems to mark the session as not eligible for `afterTurn` persistence going forward.
4. Subsequent turns continue to flush to the `.jsonl` but never re-attempt the bootstrap refresh, so messages never enter `lcm.db`.

## What would help
- A repro / unit covering "lcm plugin re-init while session file is held open by writer" path.
- If the ENOENT branch is supposed to allow live afterTurn persistence (the log says "allowing live afterTurn persistence and seeding..."), confirm the next call actually re-reads the jsonl and isn't gated by a stale flag.
- Surface the self-reload trigger in logs (`Engine initialized` arriving mid-life is currently indistinguishable from gateway startup).

Happy to run probes or share more session data privately if useful.

zerone0x:
Fixes #916

## Summary
- Retry a transient ENOENT once before afterTurn treats an active transcript as missing.
- Reuse the successful stat result when refreshing slow-path checkpoints to avoid a second stat race.
- Add a regression test covering a plugin reload/session registry race where the transcript appears during the retry window.

## Verification
- `npm test -- --run test/engine-fidelity.test.ts -t "transient ENOENT"`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/.changeset/soft-owls-reload.md
+++ b/.changeset/soft-owls-reload.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Retry transient ENOENT while reconciling afterTurn transcripts so plugin reload races do not preserve offset-0 checkpoints and skip durable catchup for active session files.
--- a/src/transcript-reconciler.ts
+++ b/src/transcript-reconciler.ts
@@ -68,6 +68,16 @@ const FLUSH_LAG_ADOPTION_TAIL_WINDOW = 16;
 // rather than scan an arbitrarily large real DB tail (#649 no-proof-no-advance).
 const NON_ANCHORING_FRONTIER_SCAN_LIMIT = 32;
 
+// A hot plugin/runtime reload can race the host session registry and produce a
+// transient ENOENT for a transcript that is still being written. Retry once
+// before taking the sticky missing-file branch so active sessions do not spend a
+// turn preserving an offset-0 checkpoint and skipping durable transcript catchup.
+const TRANSIENT_SESSION_FILE_MISSING_RETRY_MS = 25;
+
+function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 // No-anchor recovery reasons that can proceed from checkpoint/frontier metadata
 // instead of a content anchor. Block them if raw IDs are already owned by another
 // active conversation.
@@ -2072,6 +2082,7 @@ export class TranscriptReconciler {
         await this.refreshBootstrapState({
           conversationId: conversation.conversationId,
           sessionFile: params.sessionFile,
+          fileStats: sessionFileState,
         });
         rememberSlowReadState();
       } else {
@@ -2189,6 +2200,7 @@ export class TranscriptReconciler {
     await this.refreshBootstrapState({
       conversationId: conversation.conversationId,
       sessionFile: params.sessionFile,
+      fileStats: sessionFileState,
     });
     rememberSlowReadState();
     this.host.deps.log.warn(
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for the focused patch. This is pointed at the right P1 issue, and the changeset plus regression coverage are useful.

I think we should hold before merging because the core fix is a timed `ENOENT` retry. The linked report says the session file existed and the later failure mode was persistent after plugin self-reload, so the current test proves only the narrower "file appears within 25ms" case. Could you either add proof that this retry covers the observed self-reload/stale-state path, or adjust the fix/test to exercise the existing-file/reinit case more directly?

The other part of the patch, reusing the successful stat result for checkpoint refresh, looks sensible. The blocking question is whether one 25ms retry is the right recovery contract for #916 rather than masking a stale session registry/reinit state bug.

jalehman:
Thanks for the focused patch. This is aimed at the right P1, and the changeset plus regression coverage are useful, but I don’t think this PR should proceed in its current shape.

The linked report and Eva’s triage describe a persistent post-reinit `afterTurn` skip where the session file actually exists. This PR proves a narrower case: `stat()` returns ENOENT, the file appears within 25ms, and a retry succeeds. That does not establish that plugin reinitialization cannot leave stale session-file state that keeps durable persistence skipped for an existing transcript.

The `stat()` reuse part looks reasonable, but the core fix is a timed retry in a sensitive persistence path. We need either proof that this covers the observed self-reload/existing-file failure, or a revised fix/test that exercises that reinit path directly. Closing this one for now; a follow-up with that proof would be welcome.

PR #926, opened 2026-06-21
fix: preserve the live current turn across all channels (structural, plugin-agnostic)

### LINKED ISSUE #925: Context engine drops live-turn-only content (per-turn decoration and memory/context blocks) by reconstructing the current turn from the bare store
## Summary

When lossless-claw is the contextEngine, the current user turn sent to the model is reconstructed from the persisted store, which holds only the bare message body. Anything the host attaches to the live turn but does not persist (the per-turn metadata decoration, and any memory or context plugin blocks) is therefore dropped from the assembled prompt.

A re-append safety net exists (the volatile-live-input coverage added in [#688](https://github.com/Martian-Engineering/lossless-claw/pull/688)), but it recognizes only a fixed set of decoration shapes: a leading `Conversation info (untrusted metadata)` / `Sender (untrusted metadata)` preamble, or a leading channel timestamp. On a channel whose live turn does not start with one of those shapes, the gate does not fire and the bare turn reaches the model.

This is observable on the OpenClaw dashboard webchat: the live turn places content before the body, so it does not begin with the recognized preamble, and the model receives the bare turn with the decoration (and any memory blocks) removed. The impact is largest when a memory or context plugin is installed, because the model then loses the recalled context for that turn even though recall ran successfully.

## Scope

Confirmed on webchat, Telegram, and Slack with lossless-claw as the contextEngine. The same reconstruct-from-bare-store path runs for every turn; whether the loss is visible depends on whether the host re-injects the decoration after the context engine returns. On Slack it does, so the turn survives; on webchat it does not, so the bare turn reaches the model.

Reproducible on pristine upstream with no local patches: stock v0.13.1 sends a fully bare current turn on webchat.

## When it was introduced

This is a regression from [#688](https://github.com/Martian-Engineering/lossless-claw/pull/688) ("preserve unpersisted volatile live input in LCM assembly"), first released in v0.11.0. That change routed the current turn through the new volatile-live-input coverage gate, which recognizes only the shapes listed above. Releases before it (verified on v0.9.4) delivered the decorated current turn and did not exhibit this.

## Reproduction

1. Run OpenClaw with lossless-claw as the contextEngine.
2. Send a normal message from a channel whose live turn does not begin with the `Conversation info` / `Sender` preamble (the dashboard webchat).
3. Inspect the actual model request: the current user turn is the bare body, and the per-turn content the host attached to the live turn is absent. With a memory or context plugin installed, its injected blocks are absent too.
4. Disable lossless-claw and repeat: the host sends the live turn directly and the content is present. Re-enable and it regresses. This isolates the loss to the context engine reconstructing the current turn from the bare store.

## Mechanism (current source)

- `assembler.assemble` rebuilds the current turn from `getContextItems` (the bare stored body). No live content is used for the current turn.
- The only path that re-adds live content is `appendUncoveredVolatileLiveInputsWithinBudget`, gated by `isVolatileLiveInputMessage` / `hasOpenClawInboundDecoration`, which rely on `extractOpenClawInboundBody` recognizing the preamble or leading-timestamp shapes. A live turn that prepends other content (so it does not start with a recognized shape) is not detected, so it is neither re-appended nor used to supersede the bare row.

## Proposed fix (structural, plugin-agnostic)

Recognition by decoration shape cannot be made complete: different deployments attach different per-turn content (various memory and context plugins, host metadata), so any allowlist of shapes is incomplete by construction.

Instead, treat the current turn structurally. It is the last live user message, and it should supersede the matching bare assembled row(s) whenever the live content structurally contains the bare stored body (equality, or a line-aligned trailing segment, allowing an optional leading channel timestamp). The containment check is the fail-closed guard: a live turn that does not contain the bare body supersedes nothing, so genuinely distinct turns are never collapsed. This needs no knowledge of what the decoration looks like and works for any attached content, including a no-plugin setup. It is assembly-time only and persists nothing.

A reference implementation with tests is available and can be offered as a PR.

## Related

Same class as [#554](https://github.com/Martian-Engineering/lossless-claw/issues/554) and [#708](https://github.com/Martian-Engineering/lossless-claw/issues/708) (the context engine dropping content from the model input; disabling lossless-claw restores it). It also interacts with the inbound-dedup work in [#912](https://github.com/Martian-Engineering/lossless-claw/issues/912): the bare-versus-decorated persistence behavior described there is a separate, persistence-layer concern, and this report is strictly about the assembly-time drop, which it does not address.

gorkem2020:
## What

Fixes #925. The current user turn (the last live user message) now supersedes the bare assembled row(s) that `assemble()` reconstructs from the store, whenever the live content structurally contains the bare stored body. The model again receives the decorated current turn (per-turn metadata and any memory or context plugin blocks) on every channel, not just the ones whose decoration shape was previously recognized.

## Why

`assembler.assemble` rebuilds the current turn from the bare persisted body. The only re-append path (`appendUncoveredVolatileLiveInputsWithinBudget`) recognized the live turn only by a fixed set of decoration shapes (a `Conversation info` / `Sender` preamble, or a leading channel timestamp). A live turn that prepends other content (for example the dashboard webchat, where memory or context blocks come before the body) is not recognized, so the bare turn reaches the model. See #925 for the full mechanism and the v0.11.0 regression provenance.

## Approach (structural, plugin-agnostic, fail-closed)

- `liveContentContainsBareBody`: the structural primitive. True when the bare stored body equals the live content or is its line-aligned trailing segment (allowing an optional leading channel timestamp). The line boundary is the fail-closed guard against mid-line coincidences.
- `resolveStructuralCurrentTurnLiveIndex`: identifies the current turn as the last live user message, used only when it structurally contains a bare assembled row.
- `resolveStructuralCurrentTurnSupersededIndexes`: the current turn supersedes every matching bare row, collapsing redundant copies onto the single live copy.
- `stripLeadingOpenClawInboundTimestamp`: the only structurally known volatile prefix, used to align a bare row persisted with or without the channel timestamp.

No decoration or plugin-tag allowlist anywhere. It works for any attached content, including a no-plugin setup, and the structural containment subsumes the previous preamble recognition.

## Tests

New test file `test/live-turn-structural-supersede.test.ts` (12 cases) covering: the timestamp strip helper, the containment primitive (exact, timestamped trailing line, `[timestamp] body` row, and the unrelated / empty / mid-line fail-closed cases), the no-preamble (memory-first) supersede, the no-plugin `[timestamp] body` collapse, the fail-closed distinct-turn case, and an `engine.assemble` integration with the webchat shape. Written test-first (red, then green). Full suite green (1360 passing); typecheck and build clean.

## Notes

- Assembly-time only. Nothing is persisted.
- This broadens the volatile-live-input append to run for normal turns, so the existing `engine.ts` "appended unpersisted volatile live input" log fires more frequently. Its level is left unchanged for maintainers to decide.
- Strictly the assembly-time drop. The separate persistence-layer dedup behavior tracked in #912 is out of scope here.

### DIFF
--- a/.changeset/polite-lamps-travel.md
+++ b/.changeset/polite-lamps-travel.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Preserve the decorated live current turn across channels while collapsing only the matching current-turn store duplicates. A same-turn structural supersede is now accounted separately from budget eviction, so an already-budgeted prompt-recall cue is no longer dropped when the current turn is superseded.
+Preserve the decorated live current turn across channels by recognizing it structurally and appending it when the stored transcript only has the bare body. Ambiguous same-body stored rows are kept until a stable turn identity exists, so live-current-turn recovery remains lossless.
--- a/.changeset/polite-lamps-travel.md
+++ b/.changeset/polite-lamps-travel.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Preserve the decorated live current turn across channels while collapsing only the matching current-turn store duplicates.
+Preserve the decorated live current turn across channels while collapsing only the matching current-turn store duplicates. A same-turn structural supersede is now accounted separately from budget eviction, so an already-budgeted prompt-recall cue is no longer dropped when the current turn is superseded.
--- a/.changeset/polite-lamps-travel.md
+++ b/.changeset/polite-lamps-travel.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve the decorated live current turn across channels while collapsing only the matching current-turn store duplicates.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3746,7 +3746,7 @@ export class LcmContextEngine implements ContextEngine {
       }
       if (volatileLiveInputAppend.appendedMessages > 0) {
         this.deps.log.warn(
-          `[lcm] assemble: appended unpersisted volatile live input conversation=${conversation.conversationId} ${sessionLabel} appendedMessages=${volatileLiveInputAppend.appendedMessages} appendedTokens=${volatileLiveInputAppend.appendedTokens} evictedMessages=${volatileLiveInputAppend.evictedMessages} evictedTokens=${volatileLiveInputAppend.evictedTokens} supersededMessages=${volatileLiveInputAppend.supersededMessages} supersededTokens=${volatileLiveInputAppend.supersededTokens} overBudget=${volatileLiveInputAppend.overBudget}`,
+          `[lcm] assemble: appended unpersisted volatile live input conversation=${conversation.conversationId} ${sessionLabel} appendedMessages=${volatileLiveInputAppend.appendedMessages} appendedTokens=${volatileLiveInputAppend.appendedTokens} evictedMessages=${volatileLiveInputAppend.evictedMessages} evictedTokens=${volatileLiveInputAppend.evictedTokens} overBudget=${volatileLiveInputAppend.overBudget}`,
         );
       }
 
@@ -3800,7 +3800,7 @@ export class LcmContextEngine implements ContextEngine {
         : undefined;
       const summaryContextItems = contextItems.filter((item) => item.itemType === "summary").length;
       const volatileLiveInputLog = volatileLiveInputAppend.appendedMessages > 0
-        ? ` volatileLiveInputsAppended=${volatileLiveInputAppend.appendedMessages} volatileLiveInputEvicted=${volatileLiveInputAppend.evictedMessages} volatileLiveInputSuperseded=${volatileLiveInputAppend.supersededMessages} volatileLiveInputOverBudget=${volatileLiveInputAppend.overBudget}`
+        ? ` volatileLiveInputsAppended=${volatileLiveInputAppend.appendedMessages} volatileLiveInputEvicted=${volatileLiveInputAppend.evictedMessages} volatileLiveInputOverBudget=${volatileLiveInputAppend.overBudget}`
         : "";
       const promptRecallLog = budgetedPromptRecallCue
         ? ` promptRecallMatches=${budgetedPromptRecallCue.matchedMessages}`
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -3746,7 +3746,7 @@ export class LcmContextEngine implements ContextEngine {
       }
       if (volatileLiveInputAppend.appendedMessages > 0) {
         this.deps.log.warn(
-          `[lcm] assemble: appended unpersisted volatile live input conversation=${conversation.conversationId} ${sessionLabel} appendedMessages=${volatileLiveInputAppend.appendedMessages} appendedTokens=${volatileLiveInputAppend.appendedTokens} evictedMessages=${volatileLiveInputAppend.evictedMessages} evictedTokens=${volatileLiveInputAppend.evictedTokens} overBudget=${volatileLiveInputAppend.overBudget}`,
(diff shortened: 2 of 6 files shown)


jalehman:
Thanks for the careful writeup and the plugin-agnostic direction here. I think the problem is real and the structural containment primitive looks like the right building block, but I’d hold this PR until the supersede scope is tightened.

The current implementation appears to scan all assembled user rows and drop every row whose content is line-aligned-contained in the live current turn. That can collapse older repeated user turns with the same body as the current turn, not just the bare current-turn duplicate. For example, if an earlier user turn and the current user turn are both `yes`, the current live decorated copy would structurally contain both assembled rows and both could be superseded.

Could you add a regression for "current turn repeats an earlier user body" and adjust the supersede logic so it only replaces the current-turn row(s), or another bounded current-tail/coverage slot, rather than every historical matching row? Once that guard is in place, this looks like a good candidate for autoreview.

gorkem2020:
Thanks, that is a fair catch and I have tightened the supersede scope in `232bd77`.

The supersede is now bounded to the current turn's inbound tail: the trailing contiguous run of user rows, scanning back from the end and stopping at the first non-user (assistant or tool) message. The store double-write this collapses only ever materializes at first delivery of the turn, before any assistant reply is persisted, so every bare and `[timestamp]` face of the current turn lives in that trailing run on all channels (webchat, Telegram, Slack). A same-body user row sitting behind an assistant message is therefore recognized as a genuinely earlier turn and preserved, which is exactly your two `yes` example.

One detail worth flagging: the scan stops at the first non-user message, not at the first non-matching user row. That is deliberate. On the covered path the exact-equal decorated survivor sits at the tail end as a non-matching user row, and stopping there would shield the bare duplicates behind it and re-open the double-write. Skipping non-matching user rows while breaking only on the turn boundary keeps both the uncovered and covered paths correct.

Regressions added:
- unit: an earlier user turn whose body equals the current turn body survives (the bare row before the assistant is no longer collapsed).
- integration: the same shape driven through the real ingest-then-assemble store path (both `yes` turns survive, only the current one carries the live decoration).

Both fail on the previous broad scan and pass on the bound. Full suite green (1362 tests).

Known boundary, unchanged by this PR: two identical-body user messages delivered back to back with no assistant reply between them are still collapsed onto one, since at the store level they are indistinguishable from a single turn's duplicate faces. That was already the prior behavior within the tail, so this change neither introduces nor regresses it. Happy to add a stricter rule there separately if you would like.

jalehman:
Thanks for the revisions here. This is still the right direction for #925, and the repeated same-body turn guard looks much better now. I am holding it before autoreview/merge for one remaining integration issue: the current structural same-turn replacement path reports superseded bare rows as `evictedMessages`, and `engine.assemble` treats any `volatileLiveInputAppend.evictedMessages > 0` as a signal to drop an already-budgeted prompt recall cue and rerun assembly. Same-turn supersede is not real budget/context eviction, so a successful current-turn replacement can unnecessarily remove prompt-recall context.

Could you separate same-turn supersede/replacement accounting from actual budget eviction accounting, or adjust assembly so prompt recall is only dropped for real context pressure? A focused regression where prompt recall survives structural current-turn supersede when the cue fits the budget would cover the edge.

gorkem2020:
Thanks, this is fixed in [`5d98e4b`](https://github.com/Martian-Engineering/lossless-claw/pull/926/commits/5d98e4bb5c69e00118fc80c2da858251ef8d3d05).

I took the accounting-separation path you preferred. `appendUncoveredVolatileLiveInputsWithinBudget` now returns `supersededMessages` and `supersededTokens` as fields distinct from `evictedMessages` / `evictedTokens`. The same-turn supersede count is no longer folded into `evictedMessages` at the return boundary, so `evictedMessages` / `evictedTokens` now reflect only real budget-loop evictions.

The `engine.assemble` cue-drop guard is unchanged in text, so it still drops the prompt-recall cue under genuine pressure (`overBudget`, or a real budget eviction) and no longer drops it for a pure current-turn replacement. The two assemble telemetry lines now also surface the superseded counts.

Regression added in `test/engine-assemble.test.ts`: a budgeted prompt-recall cue survives a structural current-turn supersede when the cue fits the budget. It fails on the previous fused accounting (the supersede pushed `evictedMessages` above zero, which nulled the cue and re-ran assembly) and passes on the split. The existing unit assertion that expected `evictedMessages > 0` on a pure supersede now expects `supersededMessages > 0` with `evictedMessages === 0`, matching the corrected semantics. Full suite green (1364).

jalehman:

PR #927, opened 2026-06-21
fix(dedup): collapse the inbound-turn store double-write structurally

### LINKED ISSUE #912: OpenClaw inbound dedup: identity normalization covers only 2 of ~10 untrusted-metadata block types, so bare vs decorated copies persist as duplicate rows

gorkem2020:
## What

Fixes [#912](https://github.com/Martian-Engineering/lossless-claw/issues/912). When lossless-claw is the contextEngine, each inbound user turn is persisted twice: the transcript reconcile stores the bare body (with its entry id), then the afterTurn runtime batch stores a per-turn decorated copy of the same turn (a `Conversation info (untrusted metadata)` wrapper, or a leading channel timestamp). Their message identities differ, so the covered-frontier alignment treats the decorated copy as a new message and persists it as a second row. The result is two rows per turn in the store, and the duplicate replays in assembled history.

This collapses the decorated runtime copy onto the bare persisted row structurally, leaving one row per turn.

## Why this approach (relation to [#915](https://github.com/Martian-Engineering/lossless-claw/pull/915))

[#915](https://github.com/Martian-Engineering/lossless-claw/pull/915) addressed the same issue by normalizing message identity (stripping the inbound decoration so the decorated and bare copies hash equally). Review concluded that text-shape stripping cannot safely distinguish host-added decoration from a user who types timestamp or metadata looking text, so it would need a host-side trusted marker from OpenClaw before it is safe.

This PR avoids that blocker. Rather than stripping decoration to equate identities, it recognizes the decorated copy structurally: the bare persisted body must be the runtime copy's line-aligned trailing user text. The containment is fail-closed (a turn that does not line-align-contain the bare body collapses nothing), and a decoration gate restricts the collapse to copies that are the bare body wrapped in recognized OpenClaw decoration (the copy reduces to the bare body once a leading channel timestamp is stripped, or it carries an `(untrusted metadata)` block). Arbitrary leading user text is never collapsed, so a distinct turn whose trailing line happens to equal a short earlier body is preserved rather than lost. No decoration shape is hard-coded into the matcher, and no host-side marker is required. This is intended to supersede [#915](https://github.com/Martian-Engineering/lossless-claw/pull/915).

## Approach

- `runtimeRowCoversPersistedFrontierRow` (in `BatchDeduplicator`): a runtime batch row covers a persisted frontier row when their identities match, or (user to user) when the persisted bare body is the runtime row's line-aligned trailing text via `liveContentContainsBareBody`, gated on the runtime copy being recognized decoration of that body.
- It replaces the plain identity comparison inside the existing positional alignment loop of `alignRuntimeBatchAgainstCoveredFrontier`, so it only ever matches the current turn against the just-reconciled frontier tail.

## Relationship to [#926](https://github.com/Martian-Engineering/lossless-claw/pull/926)

This depends on [#926](https://github.com/Martian-Engineering/lossless-claw/pull/926) (it reuses the `liveContentContainsBareBody` primitive added there). Until #926 lands, this PR is opened against `main` and so its first commit is #926's commit; the change owned by this PR is the two dedup commits on top of it. I will rebase to drop the #926 commit once #926 merges.

## Tests

New integration tests in `test/transcript-entry-id.test.ts`: the timestamp and Conversation-info decorated copies each collapse onto the single bare row; a genuinely distinct turn that shares a trailing word is kept; and a distinct multi-line turn whose trailing line equals a short prior body is kept (no false-collapse). Full suite green, typecheck and build clean.

## Scope

Assembly-time behavior is unchanged. This covers the covered-frontier afterTurn path (`alignRuntimeBatchAgainstCoveredFrontier`); the sibling `deduplicateAfterTurnBatch` (used only on degraded transcript returns) keeps the plain identity check and can be extended the same way if it is shown to produce duplicates.


gorkem2020:
Follow-up commit `08f09fd`: extended the fix to the degraded (non-covered transcript) `afterTurn` path.

The original fix here taught `alignRuntimeBatchAgainstCoveredFrontier` (the covered-frontier path) to collapse a decorated runtime copy onto the bare persisted row. Its sibling `deduplicateAfterTurnBatch` (the path taken when the transcript reconcile does not cover the frontier, e.g. a missing or unreadable session file) still compared rows with raw `messageIdentity`, so the same decorated-vs-bare double-write could occur there.

This commit reuses the same `runtimeRowCoversPersistedFrontierRow` helper (and its fail-closed decoration gate) at the two boundary comparisons in `deduplicateAfterTurnBatch`, so both paths now behave identically. A genuinely distinct turn whose trailing line merely equals a short prior body is still kept (no false-collapse).

Added `test/f1-degraded-double-write.test.ts`: a deterministic repro that forces the degraded path via a missing session file, plus a no-false-collapse guard. Full suite green (1365 tests).

jalehman:
Thanks for the careful rework here. I think this is a good fit for #912 and the direction is better than #915 because it keeps the change local to afterTurn batch alignment instead of changing global message identity.

I’d hold this before autoreview/merge for one safety hardening point: the current decoration gate still accepts any runtime content containing `"(untrusted metadata)"` once the persisted body is a line-aligned suffix. That looks broader than the PR description’s "recognized OpenClaw decoration" claim, and it can still collapse user-authored or quoted metadata-looking text whose trailing line matches a prior bare row. For Lossless, that is a silent data-loss shape.

Could you tighten that gate to validate an actual OpenClaw injected metadata block/header shape, or otherwise add a regression test showing forged/quoted `"(untrusted metadata)"` text is preserved? After that, this should also be rebased once #926 lands and should include the changeset/release-note question for the user-visible bug fix.

gorkem2020:
Force-pushed: replaced the sample identifiers in the test fixtures with synthetic placeholders, and squashed the three incremental double-write commits into one on top of the live-current-turn base. No behavior change; the diff is identical to the prior head aside from those fixture values. Full suite green (1366 tests).

jalehman:
Thanks for the rework. I still think this is a good fit for #912, but I would keep holding it before autoreview/merge.

Two safety issues remain:

1. `runtimeRowCoversPersistedFrontierRow` still treats any content containing `(untrusted metadata)` as recognized decoration once the persisted body is a line-aligned suffix. That does not close the prior concern: user-authored or quoted metadata-looking text can still be collapsed if it ends with a prior bare row.

2. The assembly-side structural supersede path appears broader: it can recognize the last live user message purely because it line-aligned-contains an assembled bare user row, with no decoration gate. That means a distinct current user message like `here is more context\nok` can supersede an earlier assembled `ok` row. The PR has no-false-collapse coverage for that shape in the store dedup path, but not for assembly supersede.

Please tighten the recognized-decoration predicate to validate actual OpenClaw-injected shapes, or otherwise make the structural paths fail closed for quoted/forged metadata and ordinary multiline user messages. I would also add regression tests for both the store path and the assembly supersede path.

Since this fixes user-visible duplicate context replay, it should also include a patch changeset before merge.

gorkem2020:
Thanks, both tightened in [`302363d`](https://github.com/Martian-Engineering/lossless-claw/pull/927/commits/302363d).

**Issue 1 (store path).** The `(untrusted metadata)` substring check is gone. Recognition now goes through a new `contentBeginsWithOpenClawInboundMetadataBlock`, which validates the actual injected shape: after an optional leading channel timestamp and `Delivery:` hint, the content must begin with a heading line equal to a known sentinel (`Conversation info (untrusted metadata):` or `Sender (untrusted metadata):`), immediately followed by a ` ```json ` fence whose body parses to a non-array object carrying at least one heading-specific key. Quoted or forged `(untrusted metadata)` prose does not reproduce that frame, so it is preserved.

**Issue 2 (assembly path).** The assembly supersede no longer relies on line-aligned containment alone. Both paths now share one gate, `liveContentIsRecognizedDecoratedBareBody`: it collapses only when the copy structurally contains the bare body AND carries recognized decoration, which is either the validated metadata block above, or the bare body appearing as a channel-timestamp-prefixed trailing line (the plugin-agnostic webchat case, where the channel always stamps the model-facing body), or the whole copy reducing to the bare body once a single leading timestamp is stripped. An ordinary multiline message like `here is more context\nok` has none of those, so it no longer supersedes an earlier `ok`.

**Regressions added:**
- store path: a turn quoting `(untrusted metadata)` as prose whose trailing line equals a prior bare row is kept (not collapsed).
- assembly path: a distinct multiline turn, and a quoted-metadata turn, do not supersede an earlier bare row.
- the existing positive collapses (a genuine metadata block, and the webchat memory-first `[timestamp] body` shape) still pass.

Changeset added (`patch`). Full suite green (1369 tests).

One note for transparency: this branch is still on its pre-#926 base by design (you asked to rebase once [#926](https://github.com/Martian-Engineering/lossless-claw/pull/926) lands), so I have not rebased it yet. A local typecheck on the branch surfaces three pre-existing `LogMock` test-type errors in files this PR does not touch; they are already resolved on `main` and will clear on the rebase. I will rebase and squash once #926 merges.

jalehman:
Thanks for the latest tightening in `302363d`; the store-path direction still looks like a good fit for #912, and the new metadata/quoted-text regressions address the prior #927 safety comments.

I would keep this held before autoreview because the branch is still stacked on an older #926 commit. Current #926 has since fixed maintainer-reviewed assembly issues around bounded current-turn supersede and separating same-turn supersede accounting from real budget eviction. This PR head still contains the older assembly-side behavior, so reviewing #927 now would mix the #912 fix with stale #926 code.

Could you rebase/restack this after #926 lands, or onto the current #926 head, so #927 contains only the #912 store-dedup delta plus the strict decoration gate? Once the branch is clean and checks run on the rebased head, this should be ready for autoreview.

gorkem2020:
Rebased onto current `main` (now carries the live-current-turn fix from [#926](https://github.com/Martian-Engineering/lossless-claw/pull/926) and the cross-conversation guard from [#932](https://github.com/Martian-Engineering/lossless-claw/pull/932)) and reconciled with the new `identity_hash` afterTurn dedup.

The two are complementary rather than overlapping: a decorated runtime copy and its bare persisted row canonicalize to different `identity_hash` values (the metadata-block normalizer only touches volatile fields inside a recognized block), so the identity_hash path never equates decorated-vs-bare. This PR's decoration gate is layered only at the no-identity-match boundary, so it can only collapse a store double-write that identity_hash already declined, and it can never suppress an identity_hash match. The strict recognized-decoration predicate is unchanged, so forged or quoted `(untrusted metadata)` text and ordinary multiline turns are still preserved.

Two small adaptations to your new helpers during the rebase: the afterTurn comparison reads the `MessageRecord` from `getLastMessages` (since `getRecentMessageIdentityHashes` returns `string[]`), and the oversized-batch path lazily fetches `getLastMessage()` on a hash mismatch instead of the removed `lastDbMessage`.

Full suite green (1436) and CI is clean. Regression coverage intact for both the store path and the assembly supersede path.

Optional follow-up, left as your call since you just introduced the primitive: decoration recognition could eventually fold into `matchStoredMessageToIncoming` as a distinct match-kind for a single coverage decision, rather than living as a separate gate.

jalehman:

PR #931, opened 2026-06-25
fix(reconcile): recover offset-0 placeholder sessions with un-addressed or room-event frontiers

gorkem2020:
## Problem

A conversation that is created (a `conversation_bootstrap_state` row written at offset 0 / size 0) but never written to until later gets permanently stuck in the afterTurn `transcript reconcile did not cover the transcript frontier` loop the first time it is activated, when its first persisted frontier is composed entirely of OpenClaw ambient inbound rows (room-event deliveries and un-addressed channel messages). The checkpoint never advances, so the context engine goes inert for that conversation (no compaction) and emits the warning on every subsequent turn. No data is lost and the agent still answers, but the conversation is frozen and the warning multiplies per active session. A manual rotate clears it, which confirms the no-anchor import path is the right recovery lever.

## Root cause

Two compounding gates in `src/transcript-reconciler.ts`:

1. The slow-path no-anchor recovery is gated on `reason === "checkpoint-missing"`. An idle-created session carries an all-zero placeholder `conversation_bootstrap_state` row, so a row exists and the reconcile reason is `append-only-ineligible`, never `checkpoint-missing`. The recovery gate stays closed, `allowNoAnchorImport` is false, no content anchor is found against the empty offset-0 tail, and the engine treats it as unsafe to advance and skips persistence on every turn.

2. Even with gate 1 relaxed, recovery is conditioned on `conversationFrontierIsEntirelyNonAnchoring()`, which only recognized a user row whose content starts with `Conversation info (untrusted metadata)`. Real ambient frontiers do not match that prefix: they are room-event deliveries (`[OpenClaw room event]` with `inbound_event_kind: room_event`) or un-addressed channel deliveries (a `Delivery:` prelude wrapping the metadata, `explicitly_mentioned_bot: false`, `mention_source: "none"`), where the metadata block is embedded after a prelude and is followed by a real-looking trailing body (chat-history context, a current-event line, or a system notification). So the predicate returned false for the production rows.

## Fix

- Extend the no-anchor recovery to also fire when the checkpoint is the all-zero placeholder (offset 0, nothing processed, nothing to lose), treating it like a missing checkpoint for recovery purposes. The existing `placeholder-checkpoint-recovery` reason is already on the guarded-import allowlists.
- Add `isOpenClawAmbientInboundRecord` in `src/openclaw-inbound-metadata.ts`, which decides a user row is ambient/non-anchoring from the injected inbound metadata rather than from its body: true only for `inbound_event_kind === "room_event"`, or a clearly un-addressed delivery (`explicitly_mentioned_bot === false` and `mention_source === "none"`). It reuses the existing prelude split, metadata-block regex, and record parser. `conversationFrontierIsEntirelyNonAnchoring` now accepts a row that is either the existing metadata-preamble shape or this ambient shape.

## Safety

Under-match is the safe direction. Any parse failure, a missing or unexpected flag, an addressed turn (`explicitly_mentioned_bot === true` or `mention_source !== "none"`), or a non-user role returns false, so a real directed turn is never reclassified as ambient regardless of its trailing body. Recovery stays bounded to offset 0 plus a frontier proven entirely non-anchoring (and the existing 32-row scan limit), so a genuinely divergent history with real content still fail-closes per [#649](https://github.com/Martian-Engineering/lossless-claw/issues/649). This preserves the contamination boundary that closed [#824](https://github.com/Martian-Engineering/lossless-claw/issues/824).

## Tests

`test/engine-after-turn.test.ts` adds coverage built from the real frontier shapes (a room-event delivery row, an un-addressed `Delivery:` user-request row, and an addressed control row). It asserts that both an offset-0 placeholder session and a checkpoint-missing session whose frontier is entirely ambient recover (rows imported, no frontier-not-covered warning, checkpoint advances), and that two negative bounds still freeze: a non-placeholder checkpoint (offset advanced past 0), and a frontier containing an addressed turn. Full suite green.

## Relationship to prior work

Sibling of the no-anchor recovery shipped in [#837](https://github.com/Martian-Engineering/lossless-claw/issues/837) and [#822](https://github.com/Martian-Engineering/lossless-claw/issues/822); this extends it to the all-zero placeholder checkpoint and to the room-event / un-addressed-delivery frontier shapes that the original `Conversation info` prefix check did not recognize.

### DIFF
--- a/.changeset/lucky-frontiers-recover.md
+++ b/.changeset/lucky-frontiers-recover.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Recover stuck afterTurn reconciliation for OpenClaw room-event and unaddressed Delivery frontiers so compaction can resume from placeholder or checkpoint-missing sessions.
--- a/src/openclaw-inbound-metadata.ts
+++ b/src/openclaw-inbound-metadata.ts
@@ -122,16 +122,17 @@ export function canonicalizeOpenClawInboundMetadataIdentityContent(
  * (untrusted metadata)" block is present (located through the same optional
  * "[OpenClaw room event]" header and "Delivery:" prelude the rest of this
  * module handles) AND the parsed metadata is either an explicit room event, or
- * a clearly un-addressed delivery (explicitly_mentioned_bot === false AND
- * mention_source === "none").
+ * a clearly un-addressed group delivery (is_group_chat === true AND
+ * explicitly_mentioned_bot === false AND mention_source === "none").
  *
  * SAFETY (#824 contamination zone): under-match is the safe direction. Any
  * parse failure, a missing/unexpected flag, an addressed turn
  * (explicitly_mentioned_bot === true or mention_source !== "none"), or a
- * non-user role returns false. The un-addressed case requires BOTH fields; if
- * mention_source is absent we do NOT treat the row as ambient unless the event
- * is an explicit room_event. A real directed turn is never misclassified as
- * ambient regardless of its trailing body.
+ * non-user role returns false. The un-addressed case requires the explicit
+ * group-chat flag plus BOTH mention fields; if any are absent we do NOT treat
+ * the row as ambient unless the event is an explicit room_event. A real
+ * directed turn is never misclassified as ambient regardless of its trailing
+ * body.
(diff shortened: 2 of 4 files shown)


jalehman:

PR #932, opened 2026-06-25
fix(reconcile): scope cross-conversation raw-id guard to an agent's own channel siblings

gorkem2020:
## Problem

A base channel session can get permanently stuck in the afterTurn `did not cover the transcript frontier` loop. Its no-anchor recovery import is blocked by the cross-conversation raw-id contamination guard because the candidate raw ids already exist in other active conversations, so the engine skips persistence every turn. The conversation never anchors or compacts (no data loss, but it goes inert and the warning repeats on every turn).

## Root cause

The "other active conversations" that trip the guard are the same agent's own sibling sessions on the same channel. A channel agent gets a base channel session plus a separate per-thread session for each thread it replies in (and active-memory variants). Those thread sessions are forks of the base session: they carry a `parentSession` reference and inherit its transcript prefix and entry-id chain, so their tool-call and block ids overlap the base session's legitimately. `countActiveCrossConversationRawIdMatches` counted those siblings as foreign owners and vetoed the import, treating an agent importing its own lineage as contamination.

This is a distinct failure from the placeholder/room-event frontier recovery in #931. There the recovery path was never reached; here it is reached but blocked by the guard. A base channel session with no bootstrap row reaches `checkpoint-missing-recovery`; one with an all-zero placeholder reaches `placeholder-checkpoint-recovery`. Both are vetoed by the same guard.

## Fix

Add `sessionKeyChannelScope`, which reduces a session key to its `agent:<id>:...:channel:<id>` scope, stripping `:thread:<ts>` and `:active-memory:<hash>` suffixes. The guard now excludes active conversations that share that scope (the base channel session, its thread forks, and active-memory variants) from the cross-conversation match. Genuinely foreign conversations (a different agent, a different channel, or a key with no channel segment) still block, preserving the #824 contamination boundary. A key shape that does not parse returns null and still blocks (fail-safe, under-match preferred).

## Tests

- `test/engine-after-turn.test.ts`: a no-anchor recovery whose only colliding raw ids belong to a same-agent same-channel thread sibling now imports the transcript (previously blocked); a same-agent different-channel collision still blocks.
- `test/session-patterns.test.ts`: unit coverage for the scope reduction (thread and active-memory suffixes stripped, no-channel keys return null) and the sibling predicate (different agent, different channel, and non-channel keys are not siblings).

Full suite green, typecheck clean.

### DIFF
--- a/.changeset/same-channel-sibling-recovery.md
+++ b/.changeset/same-channel-sibling-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Recover base channel sessions that were stuck in afterTurn reconciliation when their raw ids also appeared in same-agent same-channel thread or active-memory fork history.
--- a/src/session-patterns.ts
+++ b/src/session-patterns.ts
@@ -39,11 +39,8 @@ export function sessionKeyChannelScope(
   return match ? match[1] : null;
 }
 
-/** Whether two session keys are siblings of the same agent on the same channel. */
-export function sessionKeysAreSameAgentChannelSiblings(
-  a: string | null | undefined,
-  b: string | null | undefined,
-): boolean {
-  const scopeA = sessionKeyChannelScope(a);
-  return scopeA !== null && scopeA === sessionKeyChannelScope(b);
+/** Whether a session key names the base session for its agent and channel. */
+export function isBaseChannelSessionKey(sessionKey: string | null | undefined): boolean {
+  const scope = sessionKeyChannelScope(sessionKey);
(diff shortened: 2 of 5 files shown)


gorkem2020:
Rebased onto current main (now including [#931](https://github.com/Martian-Engineering/lossless-claw/pull/931)). The reconcile recovery region and the engine-after-turn tests merged with no conflict, so this now sits cleanly on top of the placeholder/room-event recovery and the direct-inbound is_group_chat hardening that landed in #931. Typecheck clean, full suite green (1366). The guard-scope change itself is unchanged from the prior head.

jalehman:

PR #935, opened 2026-06-27
fix(reconcile): dedup the decorated runtime row on never-ingested recovery

gorkem2020:
> Rebased onto `main` now that [#927](https://github.com/Martian-Engineering/lossless-claw/pull/927) has merged (it reuses `liveContentIsRecognizedDecoratedBareBody` from that change). This PR is now the single fix commit on top of `main`.

## Problem

On a brand-new conversation, the live `afterTurn` ingest can persist a turn as a decorated, still-unstamped row before the session jsonl is flushed. The subsequent never-ingested recovery (an all-zero placeholder checkpoint, or no checkpoint row at all) then re-imports the same turn's bare, id-bearing transcript row through `importNoAnchorEpoch`.

The bare and decorated faces of that turn differ on both `identity_hash` and `content` (the decorated face carries the injected inbound-metadata block and the channel timestamp), so the existing adopters all miss the pair:

- `hasMessageByTranscriptEntryId` and `adoptTranscriptEntryId` key on the entry id or an exact `identity_hash` + `content` match, which the decorated row does not satisfy.
- the replay-overlap analyzer keys on the raw message identity, which also differs.

The bare row therefore imports as a duplicate. And because the import is treated as a fresh, non-overlapping epoch, the checkpoint never advances off the placeholder, so the conversation re-enters recovery every turn and loses continuity.

### Where it bites

The trigger is host plus classification level, not channel specific. When an inbound is classified as a room event (ambient group or unaddressed), `suppressNextUserMessagePersistence` is set, so the user message is placed in the model prompt but never written to the jsonl as a runtime event. On the embedded context-engine host, which does not pre-flush the transcript before `afterTurn` (unlike the codex app-server host, which awaits a mirror first), the file `afterTurn` stats is then missing or empty for that turn, which seeds the placeholder and drives the re-import. Any deployment on the embedded host that starts new conversations from ambient or room-event inbounds is exposed.

## Change

Before importing a bare row on a never-ingested recovery, adopt a recent same-role unstamped row that is the recognized decorated face of the same turn, stamping it with the bare row's entry id instead of importing a duplicate.

- New `ConversationStore.listRecentUnstampedMessagesByRole`: returns the windowed (flush-lag tail) unstamped rows of a role, oldest-first, so a multi-turn recovery pairs each bare row with its own decorated twin chronologically rather than letting a short earlier turn adopt a later, longer row that merely shares a trailing line.
- New `TranscriptReconciler.adoptDecorationInvariantEntryId`: matches a candidate via the existing structural primitive `liveContentIsRecognizedDecoratedBareBody` (line-aligned containment plus recognized-decoration evidence, fail-closed) and restamps on a hit.
- Gated on the two never-ingested recovery reasons (`placeholder-checkpoint-recovery` and `checkpoint-missing-recovery`). On any other reason, or with no decorated twin in the window, behaviour is unchanged.

Adopting rather than skipping makes the import report overlap, so the caller advances the checkpoint and the conversation stops re-entering recovery, which also resolves the continuity loss.

It is complementary to the existing recovery guards ([#822](https://github.com/Martian-Engineering/lossless-claw/issues/822), [#824](https://github.com/Martian-Engineering/lossless-claw/issues/824), [#649](https://github.com/Martian-Engineering/lossless-claw/issues/649)) and to the afterTurn `identity_hash` dedup, none of which catch the decorated-versus-bare pair.

## Testing

- `npm run typecheck` clean.
- Full suite green (`npm test`).
- New `test/placeholder-recovery-double-write.test.ts` drives the real `afterTurn` recovery path with id-bearing transcripts (built through `SessionManager`, so entries carry genuine envelope ids) and a non-anchoring room-event frontier:
  - single-turn: the turn survives once as the decorated row, the assistant reply still imports, and the checkpoint advances off the placeholder (continuity).
  - multi-turn: two turns where the earlier body is the trailing line of the later one both survive exactly once (the oldest-first ordering regression).
  - checkpoint-missing: the same dedup on the `checkpoint-missing` recovery lane.

### DIFF
--- a/.changeset/never-ingested-decorated-recovery.md
+++ b/.changeset/never-ingested-decorated-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve conversation continuity when never-ingested recovery encounters an exact metadata-decorated runtime copy of a bare transcript row.
--- a/src/store/conversation-store.ts
+++ b/src/store/conversation-store.ts
@@ -1092,9 +1092,8 @@ export class ConversationStore {
    * entry id — the flush-lagged runtime rows a transcript catch-up may adopt.
    * Returns id + content so the caller can apply a decoration-invariant match
    * before stamping, rather than the exact identity_hash + content match the SQL
-   * adopters use. Ordered OLDEST-first so a multi-turn recovery pairs each bare
-   * row with its own decorated twin chronologically, instead of letting a short
-   * earlier turn adopt a later, longer row that merely shares a trailing line.
+   * adopters use. Ordered OLDEST-first so a multi-turn recovery consumes
+   * repeated equal bodies in transcript order.
    */
   async listRecentUnstampedMessagesByRole(
     conversationId: ConversationId,
--- a/src/store/conversation-store.ts
+++ b/src/store/conversation-store.ts
@@ -1088,6 +1088,40 @@ export class ConversationStore {
   }
 
   /**
+   * Tail rows (within the window) of the given role that carry no transcript
+   * entry id — the flush-lagged runtime rows a transcript catch-up may adopt.
+   * Returns id + content so the caller can apply a decoration-invariant match
+   * before stamping, rather than the exact identity_hash + content match the SQL
+   * adopters use. Ordered OLDEST-first so a multi-turn recovery pairs each bare
+   * row with its own decorated twin chronologically, instead of letting a short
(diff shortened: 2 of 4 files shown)


gorkem2020:
Gentle ping on this one. It has been open nine days with all checks green and no review activity yet, and it is the oldest of our open fixes here. The change is contained (dedup of the decorated runtime row on never-ingested recovery, with tests). Happy to rebase or adjust the approach if anything about it needs discussion.

jalehman:

PR #936, opened 2026-06-27
feat(engine): report deferred maintenance debt

christian-lallo:
## Summary

- expose `info.hasDeferredMaintenance(...)` from the Lossless Claw context engine
- report deferred maintenance only when conversation compaction maintenance is pending or running
- return false for ignored, stateless, and no-conversation sessions
- add lifecycle coverage for the no-debt and pending-debt states
- add a patch changeset

## What Problem This Solves

OpenClaw can now ask context engines whether a turn-triggered background maintenance pass has actual durable work to do. Lossless Claw already tracks that state in `CompactionMaintenanceStore`; this PR wires that existing debt state into the optional host hint.

`turnMaintenanceMode: "background"` remains the capability signal. `hasDeferredMaintenance(...)` is the debt signal.

## Behavior

The hint returns:

- `false` for ignored sessions
- `false` for stateless sessions
- `false` when no conversation exists for the session
- `true` only when the conversation maintenance record is `pending` or `running`

## Companion

OpenClaw host API/support: https://github.com/openclaw/openclaw/pull/97284

## Evidence

- `npx vitest run test/engine-lifecycle.test.ts`
- `npm run typecheck`
- `npm run build`
- Regression coverage verifies the hint resolves `false` before compaction debt is requested and `true` after `requestProactiveCompactionDebt(...)`.

## Proof (real run)

Captured from a live gateway running a companion OpenClaw host with this engine hint implementation, before and after deploying the host/plugin pair.

**Before** — on turns with no durable maintenance debt, the host queued a no-op deferred maintenance task on essentially every successful turn:

```
[context-engine] deferred turn maintenance queued taskId=<redacted> sessionKey=<redacted> lane=context-engine-turn-maintenance:<redacted>
```

These fired continuously throughout the day (peaking at ~21 in a single hour; last occurrence immediately before the deploy).

**After** — with this engine hint live (no compaction debt -> `hasDeferredMaintenance(...)` resolves `false` -> the companion host skips scheduling):

- **0** `deferred turn maintenance queued` lines across the **1,172** turn-level context-engine events logged since the deploy.
- The active conversation had no maintenance-debt row, so this implementation returned `false` from existing `CompactionMaintenanceStore` state rather than creating new bookkeeping.

Real pending/running maintenance debt still reports `true` and remains eligible for background scheduling.

### DIFF
--- a/.changeset/deferred-maintenance-hint.md
+++ b/.changeset/deferred-maintenance-hint.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Expose `hasDeferredMaintenance` from the OpenClaw context-engine metadata so hosts can skip background turn maintenance when no compaction debt is pending.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -340,6 +340,7 @@ export class LcmContextEngine implements ContextEngine {
       version: "0.1.0",
       ownsCompaction: migrationOk,
       turnMaintenanceMode: "background",
+      hasDeferredMaintenance: (params) => this.hasDeferredMaintenance(params),
       hostRequirements: {
         "agent-run": {
           requiredCapabilities: LOSSLESS_AGENT_RUN_REQUIRED_HOST_CAPABILITIES,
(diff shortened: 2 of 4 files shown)


jalehman:
Thanks. The Lossless-side implementation looks like the right shape: it uses the existing compaction-maintenance row and keeps the hint narrow to pending/running durable debt.

I’d like to hold this until the companion OpenClaw host API lands in a stable release. Since this PR’s user-visible behavior depends on the host calling `hasDeferredMaintenance(...)`, please verify the first stable OpenClaw version that supports that contract and update the compatibility/docs as needed before we do a full review.

jalehman:
Thanks for the PR. The Lossless-side shape still looks reasonable: it uses the existing compaction-maintenance row and keeps the hint narrow to pending/running durable debt.

I’m going to close this for now because the required OpenClaw host contract did not land. The companion OpenClaw PR was closed for inactivity, current OpenClaw main does not expose hasDeferredMaintenance(...), and there is no stable OpenClaw release to declare as the minimum supported host for this behavior.

If the host API lands later, this would be worth resubmitting with the matching stable OpenClaw compatibility metadata/docs update and refreshed tests against that released contract.

PR #937, opened 2026-06-27
fix(engine): log volatile live input append at debug unless over budget

gorkem2020:
## Problem

`LcmContextEngine.assemble` logs `[lcm] assemble: appended unpersisted volatile live input ...` at `warn`, gated only on `appendedMessages > 0`. The live current turn is appended as volatile input on essentially every turn, so this fires a `warn` on nearly every assemble. In normal operation the append is benign: `overBudget=false` and `evictedMessages=0`. The fields that would make the line actionable are never set, so it is per-turn noise at warn level.

## Fix

Log the benign append at `debug`, and keep `warn` only when the append actually went over budget (`overBudget`) or had to evict assembled messages (`evictedMessages > 0`). The message string is unchanged, so the genuine signal (real budget pressure or dropped assembled messages) stays at warn while the normal per-turn append drops to debug.

## Test

Adds `test/volatile-live-input-log-level.test.ts`, an `engine.assemble` integration test:

- a benign volatile append (large budget, `overBudget=false`, `evictedMessages=0`) logs at `debug` and not `warn`;
- an over-budget volatile append (tight budget) still logs at `warn` and not `debug`.

Verified failing before the change (the benign case logged at warn) and passing after. The full suite passes (1429 tests) and `tsc --noEmit` is clean.

### DIFF
--- a/.changeset/quiet-live-input-logs.md
+++ b/.changeset/quiet-live-input-logs.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Log benign volatile live-input appends at debug instead of warn, while preserving warnings when the append goes over budget or evicts assembled messages.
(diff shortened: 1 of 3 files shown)


jalehman:

PR #938, opened 2026-06-28
[codex] Repair unsafe conversation ids during migration

VontaJamal:
## Summary

Repair legacy LCM SQLite databases where `conversations.conversation_id` can exceed JavaScript's safe integer range before migration, startup recovery, or recall tools read those rows.

The migration now builds a SQLite-side remap table, updates all known `conversation_id` references, resets the `conversations` sqlite sequence, and avoids reading unsafe integer values into JS while repairing.

## Root Cause

`node:sqlite` throws when reading integer values larger than `Number.MAX_SAFE_INTEGER`. Existing safe-ID repair covered oversized `message_id` rows, but not oversized `conversation_id` rows, so startup/session hooks and `lcm_grep` could fail before Lossless could repair or use the DB.

## Validation

- `npm test -- test/migration.test.ts`
- `npm run typecheck`
- `npm run build`
- `npm test`
- `npm run plugin-inspector:ci -- --openclaw /tmp/lossless-openclaw-contract`

### DIFF
--- a/.changeset/safe-conversation-id-repair.md
+++ b/.changeset/safe-conversation-id-repair.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Repair legacy SQLite databases whose `conversation_id` values exceed JavaScript's safe integer range before migration, startup recovery, or recall tools read those rows.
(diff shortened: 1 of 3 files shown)


VontaJamal:
Closing per owner request.

PR #939, opened 2026-06-28
fix(dedup): cover whitespace-divergent faces of the same turn

gorkem2020:
## Problem

OpenClaw core builds the in-memory runtime message with the user's indentation collapsed (multi-space runs squashed to a single space), while the session transcript persists it verbatim. The two faces of one user turn then differ only in whitespace, so their `identity_hash` values differ and the decorated-bare containment check also fails on the whitespace. The afterTurn batch alignment no longer recognizes the runtime row as covering the persisted frontier row, and the collapsed twin is ingested as a second user row: a store double-write.

In practice this fires on any inbound carrying significant whitespace runs (pasted code, JSON, or indented config), so it is not a rare edge case.

## Fix

Add a third coverage arm to `runtimeRowCoversPersistedFrontierRow`: a whitespace-only normalized equality (user-role, frontier-positional, comparison only). Storage stays byte-verbatim and the persisted verbatim row is the survivor. The normalization is whitespace-only (no case folding), so turns that differ in real content are never collapsed.

## Tests

- `test/whitespace-divergence-double-write.test.ts` reproduces the covered-path double-write (two stored user rows before the fix, one after).
- A guard test asserts that two genuinely distinct turns differing beyond whitespace are never collapsed.

Full suite green, typecheck clean.

## Note

Sibling to [#927](https://github.com/Martian-Engineering/lossless-claw/pull/927) (a fix for a different divergence class), now merged. Rebased onto `main`; this PR is the single fix commit.

### DIFF
--- a/.changeset/whitespace-frontier-space-run-narrow.md
+++ b/.changeset/whitespace-frontier-space-run-narrow.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Recognize the specific OpenClaw runtime/transcript whitespace divergence, where core collapses runs of spaces in the runtime message to a single space while the transcript persists them verbatim, as one user turn during afterTurn frontier-coverage. This prevents a store double-write of the same turn without collapsing newlines, tabs, or leading and trailing whitespace, so two turns that differ in meaningful whitespace (line breaks or tab indentation) are never merged. Storage stays byte-verbatim; the persisted row is the survivor.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -29,6 +29,17 @@ import { buildMessageIdentityHash } from "./store/message-identity.js";
 import type { LargeFileRecord, SummaryStore } from "./store/summary-store.js";
 import type { LcmDependencies } from "./types.js";
 
+/**
+ * Collapse only the whitespace OpenClaw core actually rewrites on the runtime
+ * face: runs of spaces squashed to a single space. Newlines, tabs, and
+ * leading/trailing whitespace are preserved, so two turns that differ in
+ * meaningful whitespace (line breaks or tab indentation) are never treated as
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks, this is the right class of bug to fix: preserving the verbatim transcript row while avoiding a second whitespace-collapsed runtime copy fits Lossless well.

I’d like to hold before full review because the current normalization is broader than the problem statement. `content.replace(/\s+/g, " ").trim()` treats all whitespace-only differences as equivalent, including newlines/tabs/leading/trailing whitespace, and this helper is used by broader dedup/suffix paths, not only the covered frontier case. For pasted code/config/JSON, whitespace can be meaningful, so this risks collapsing two distinct user turns.

Can you narrow the comparison to the exact OpenClaw divergence shape we have evidence for, and add a guard proving semantically distinct whitespace-only content is preserved? This should also include a patch changeset because it changes persisted dedup behavior.

gorkem2020:
Thanks, that is a fair call and I have narrowed it accordingly.

- **Narrowed the normalization to the evidenced shape.** `normalizeFrontierWhitespace` now collapses only runs of spaces (`content.replace(/ +/g, " ")`), which is exactly what core rewrites on the runtime face: multi-space runs squashed to a single space, per the reproduction. It no longer touches newlines or tabs and no longer trims leading or trailing whitespace, so genuinely distinct multiline content (line breaks, tab indentation) can never be collapsed.
- **Added a guard for meaningful whitespace-only differences.** A new test asserts that two turns whose only difference is a line break versus a space are both preserved. They would have been wrongly merged under the previous `/\s+/`. The existing content-difference guard stays, and the original space-run reproduction still collapses to a single verbatim row.
- **Added the patch changeset.**

Full suite green (1461) and typecheck clean. The comparison stays user-role and frontier-positional, and storage is still byte-verbatim with the persisted row as the survivor.

jalehman:

PR #943, opened 2026-06-28
fix(batch-dedup): demote full-overlap fail-closed log to debug

gorkem2020:
### What

`alignRuntimeBatchAgainstCoveredFrontier` fails closed (returns `[]`) and logs at `warn` whenever the runtime batch overlaps persisted history without forming a cleanly alignable suffix. This splits that single `warn` by how much of the batch overlaps:

- **Full overlap** (`persistedIdentityOverlaps === batch.length`): every row in the batch is already persisted, so failing closed drops nothing new and the next reconcile has nothing to deliver. A benign dedup no-op. Demoted to `debug`.
- **Partial overlap** (`persistedIdentityOverlaps < batch.length`): the batch still contains genuinely new rows that get deferred to the reconcile. Stays at `warn`.

The fail-closed behavior is unchanged in both cases (both still `return []`). Only the log severity changes, and the full-overlap message is preserved verbatim at `debug` so it stays recoverable when debugging frontier alignment.

### Why

The guard is a deliberate fail-closed observability signal, and that intent is kept here. The change is purely signal-to-noise: in a healthy long-running session the full-overlap case fires repeatedly (afterTurn replay where the whole batch is already covered), and at `warn` it reads identically to the partial-overlap case, which is the one that actually defers new content. Separating them keeps the meaningful divergence loud while quieting the no-op.

### Tests

Adds `test/batch-dedup-frontier-loglevel.test.ts` (3 cases): full overlap (N == M) logs at `debug` not `warn`; partial overlap (N < M) stays at `warn`; zero overlap ingests the full batch with no overlap log. Full suite (1430) and `tsc --noEmit` pass.

### Note for maintainers

If you rely on the full-overlap case staying at `warn` (for instance to catch a frontier-tracking regression where every row is present but not suffix-alignable), let me know and I will drop the demotion or gate it behind a flag. Happy to match whatever observability contract you want here.

### DIFF
--- a/.changeset/quiet-frontier-overlap.md
+++ b/.changeset/quiet-frontier-overlap.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Demote fully redundant afterTurn frontier-overlap fail-closed logs from warn to debug while preserving warnings for partial or under-covered overlap batches.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -130,7 +130,7 @@ export class BatchDeduplicator {
       // Full overlap (N === batch.length) means every row is already persisted, so
       // failing closed drops nothing new: log at debug. A partial overlap still
       // defers genuinely new rows to the reconcile and stays at warn.
-      if (persistedIdentityOverlaps === batch.length) {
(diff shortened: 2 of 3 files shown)


jalehman:

PR #945, opened 2026-06-29
feat(tools): let lcm_grep search externalized file contents

mpz4life:
## Problem

`lcm_describe(id=file_xxx, expandFile=true)` inlines the original content of an already-externalized file, but caps the inlined bytes at `expandFileMaxBytes` (default 32KB). Its tool description told agents to use `lcm_grep` to search the full file when the output was truncated.

That guidance was misleading: `lcm_grep` only searched indexed conversation messages and summaries, not the contents of `large_files`. An agent following the hint would search the wrong scope and fail to find content that existed only in the externalized file.

## Solution

Teach `lcm_grep` to search externalized file contents:

- New `scope: "files"` option restricts the search to `large_files` rows.
- New optional `fileIds: ["file_xxx", ...]` parameter restricts the search to specific file IDs.
- Regex and full-text modes are both supported.
- Each match reports the file ID, line number, byte offset, matched text, and a contextual snippet.

`lcm_describe` now gives accurate guidance: when inlined content is truncated, use `lcm_grep(scope="files", fileIds=[file_xxx])` to search the full file.

## Changes

- `src/store/summary-store.ts`
  - Add `searchLargeFiles` with regex/full-text matching over file contents.
  - Return per-match results with `matchedText`, `lineNumber`, `byteOffset`, and `snippet`.
  - Cap each file read to 512KB and skip non-text MIME types.

- `src/retrieval.ts`
  - Extend `GrepInput.scope` with `"files"` and add `fileIds`/`largeFilesDir` fields.
  - Route `scope="files"` to `summaryStore.searchLargeFiles`.

- `src/tools/lcm-grep-tool.ts`
  - Expose `scope="files"` and `fileIds` in the tool schema and description.
  - Render a "Files" section in the output with line numbers and matched text.

- `src/tools/lcm-describe-tool.ts`
  - Restore `lcm_grep` guidance now that the tool can actually search file contents.

- `test/engine-ingest.test.ts`
  - Add integration test covering regex search, full-text search, and `fileIds` filtering on externalized tool outputs.

## Verification

```bash
pnpm test test/lcm-tools.test.ts test/engine-ingest.test.ts test/retrieval-sort.test.ts test/lcm-integration-retrieval.test.ts -- --run --reporter=dot
```

61 tests passed.

### DIFF
--- a/.changeset/wet-bags-create.md
+++ b/.changeset/wet-bags-create.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Teach `lcm_grep` to search the contents of externalized `large_files` rows via the new `scope="files"` option. Add an optional `fileIds` parameter to restrict the search to specific file IDs. Each match reports the file ID, line number, byte offset, matched text, and a contextual snippet. Update `lcm_describe` to give accurate guidance when inlined content is truncated. Honor `allConversations=true` for `scope="files"` by searching large files across all conversations.
+Teach `lcm_grep` to search the first 512,000 bytes of externalized `large_files` text rows via the new `scope="files"` option. Add an optional `fileIds` parameter to restrict the search to specific file IDs. Each match reports the file ID, line number, byte offset, matched text, and a contextual snippet. Update `lcm_describe` to give accurate bounded-search guidance when inlined content is truncated. Honor `allConversations=true` for `scope="files"` by searching large files across all conversations.
--- a/.changeset/wet-bags-create.md
+++ b/.changeset/wet-bags-create.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Teach `lcm_grep` to search the contents of externalized `large_files` rows via the new `scope="files"` option. Add an optional `fileIds` parameter to restrict the search to specific file IDs. Each match reports the file ID, line number, byte offset, matched text, and a contextual snippet. Update `lcm_describe` to give accurate guidance when inlined content is truncated. Honor `allConversations=true` for `scope="files"` by searching large files across all conversations.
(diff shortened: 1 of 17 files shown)


mpz4life:
@jalehman I think we should enable the lcm_grep tool to support querying key snippets from files via regex matching.

jalehman:
Thanks, this is a good direction. Searching externalized `large_files` content is the right fix for the misleading `lcm_describe` guidance, and the PR has a solid amount of test coverage around regex/full-text search, `fileIds`, MIME skipping, and line-range describe reads.

I’d like to hold before full review for two things:

1. This needs a patch changeset because it changes user-facing LCM tool behavior.
2. `scope="files"` should honor the existing `allConversations=true` contract. Right now the shared scope resolver returns all-conversation mode without concrete conversation IDs, but the new large-file search only loads files when `fileIds`, `conversationIds`, or `conversationId` are present, so `lcm_grep(scope="files", allConversations=true)` silently returns no file matches.

Once those are fixed, this looks worth full review.

mpz4life:
Hi @jalehman  — thanks for the review. Both items should now be addressed:
1. Changeset added: .changeset/wet-bags-create.md with a patch bump describing the new lcm_grep(scope="files") behavior.
2. allConversations=true honored for scope="files":
- Added allConversations to GrepInput and LargeFileSearchInput.
- Added SummaryStore.getAllLargeFiles() for a full-table scan ordered by created_at DESC.
- searchLargeFiles now falls back to getAllLargeFiles() when no explicit fileIds/conversationId/conversationIds are provided and allConversations is true.
- Wired conversationScope.allConversations through lcm-grep-tool.ts into retrieval.grep.
- Added a regression test covering cross-conversation file search with allConversations: true.
Pushed as two commits to this branch:
- a6ddee1 feat(tools): honor allConversations=true for lcm_grep scope=files
- 2972d9b chore(changeset): add patch changeset for lcm_grep files scope
All relevant tests pass (78 tests) and pnpm run typecheck is clean. Please take another look when you have a moment.

jalehman:

PR #946, opened 2026-06-29
fix(interceptor): do not re-externalize oversized lcm_describe tool results

mpz4life:
## Problem

`lcm_describe` is the drill-down tool an agent uses when it sees a compact `[LCM Tool Output: file_xxx | tool=… | N bytes]` reference in the conversation. Its whole purpose is to return the original, full content of that externalized file.

Because that content can be large, the large-file interceptor was treating the `lcm_describe` result itself as a new oversized tool output and wrapping it in a second `[LCM Tool Output: file_yyy]` stub. This defeated the tool:

- The agent calls `lcm_describe(id=file_xxx, expandFile=true)` expecting the original file content.
- It receives a stub pointing at `file_yyy` instead.
- It must then call `lcm_describe(id=file_yyy)` to read what it just asked for, wasting a turn and a model context window.

This is especially wasteful because `lcm_describe` output is already the answer to a previous externalization; re-externalizing it adds no new information and only creates indirection.

## Fix

In `interceptLargeToolResults`, detect `toolName === "lcm_describe"` before the threshold/externalization step and keep the original block inline. The content still flows through normal message-part storage, so it is not lost; it is simply not turned into another stub.

## Files Changed

- `src/large-file-interceptor.ts`  
  Skip externalization for blocks whose tool name is `lcm_describe`.

- `test/engine-ingest.test.ts`  
  Add regression test: ingesting an oversized `lcm_describe` tool result with `largeFileTokenThreshold=20` produces no `large_files` rows and no `toolOutputExternalized` / `externalizedFileId` metadata.

## Verification

```bash
pnpm test test/engine-ingest.test.ts -- --run --reporter=dot -t "keeps oversized lcm_describe tool results inline"
```

All 28 tests in `test/engine-ingest.test.ts` pass.

### DIFF
--- a/.changeset/lcm-describe-inline-results.md
+++ b/.changeset/lcm-describe-inline-results.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Keep oversized `lcm_describe` tool results inline instead of wrapping them in another externalized tool-output stub, so drilldown responses return the requested content directly.
--- a/src/large-file-interceptor.ts
+++ b/src/large-file-interceptor.ts
@@ -850,10 +850,7 @@ export class LargeFileInterceptor {
         safeString(record.id) ??
         topLevelToolCallId;
 
-      // The lcm_describe tool returns the original content of an already-
-      // externalized file. Wrapping that output in another externalized stub
-      // makes the drilldown useless and forces the agent to call lcm_describe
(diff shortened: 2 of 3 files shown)


mpz4life:
@jalehman ready to review

jalehman:
Great. Good catch. Will review today. New release coming soon.

jalehman:

PR #947, opened 2026-06-29
fix: relax sessionId guard for isolated cron checkpoint-missing recovery

ralf003:
## Summary
Isolated cron sessions share a durable sessionKey, but the checkpoint-missing recovery path in `transcript-reconciler.ts` guards on `conversation.sessionId === params.sessionId`. Since sessionId changes every run, when multiple cron jobs fire simultaneously, the last bootstrap overwrites the conversation's sessionId, causing earlier runners to fail the guard, enter a stuck loop, and eventually overflow context.

## Changes
- Export `isIsolatedCronSessionKey` from `src/tools/lcm-conversation-scope.ts` (was private)
- Import and add `isIsolatedCronSessionKey(conversation.sessionKey ?? undefined)` as an OR condition alongside the existing sessionId guard in `src/transcript-reconciler.ts`
- Add regression test in `test/isolated-cron-checkpoint-missing-race.test.ts`

## Safety
The helper returns `false` for all non-cron session keys, so the existing guard behavior is identical for non-cron paths.

Fixes: concurrent isolated cron `context_overflow`

### DIFF
--- a/.changeset/cron-sessionid-recovery.md
+++ b/.changeset/cron-sessionid-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Allow isolated cron sessions with a matching durable sessionKey to recover from a checkpoint-missing reconciliation state even when the conversation's sessionId has been overwritten by a newer cron run.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -2902,6 +2902,12 @@ export class LcmContextEngine implements ContextEngine {
         hasOverlap: false,
       };
     }
+    if (transcriptReconcileResult.blockedReason === "stale-isolated-cron-afterturn") {
+      this.deps.log.warn(
+        `[lcm] afterTurn: stale isolated cron callback skipped before persistence and compaction maintenance ${sessionLabel}`,
+      );
+      return;
+    }
     const transcriptReconcileUnsafeToAdvance =
       transcriptReconcileResult.blockedByImportCap ||
       (!transcriptReconcileResult.hasOverlap && transcriptReconcileResult.importedMessages === 0);
--- a/src/plugin/lcm-doctor-rollover-splits.ts
+++ b/src/plugin/lcm-doctor-rollover-splits.ts
@@ -1,5 +1,6 @@
 import type { DatabaseSync, SQLInputValue } from "node:sqlite";
 import { getFileBackedDatabasePath } from "../db/connection.js";
+import { isIsolatedCronSessionKey } from "../session-patterns.js";
 import { normalizeMessageContentForFullTextIndex } from "../store/conversation-store.js";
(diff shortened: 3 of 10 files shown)


jalehman:
Thanks for digging into this. The underlying failure mode is relevant and worth fixing, but I think this needs a stronger proof before we merge it.

The current patch treats an isolated cron `sessionKey` as an alternate binding identity for checkpoint-missing recovery, while the existing rollover code documents cron runtime UUID changes as isolated scheduled runs. That may still be the right exception for this specific recovery path, but the added test only checks the exported helper and does not exercise the reconciler branch that was changed.

Can you add a regression that reproduces the stale/mismatched `sessionId` checkpoint-missing case through `afterTurn`, and also proves we do not regress the existing cron isolation / fail-closed rollover behavior? This will also need a patch changeset because it changes recovery behavior.

ralf003:
## Test Update

Added reconciler-branch regression tests that exercise the guard condition through the engine.afterTurn flow:

1. **Cron recovery (main test)**: Bootstraps a cron conversation, deletes the bootstrap checkpoint, simulates a TOCTOU race (DB-level sessionId overwrite + no-op mock of rollover detector), calls afterTurn with the original sessionId, and verifies that `isIsolatedCronSessionKey` in the relaxed guard allows checkpoint-missing recovery.

2. **Non-cron fail-closed**: Same TOCTOU setup but with a non-cron sessionKey -- proves the fix does NOT weaken the guard for non-cron paths.

3. **Cron isolation preservation**: Verifies the existing rollover detector behavior (archives old conversation, creates new one) is not broken by the fix.

### Test results
```
 ✓ test/isolated-cron-checkpoint-missing-race.test.ts (5 tests)
   ✓ isIsolatedCronSessionKey export and behavior (2)
   ✓ isolated cron checkpoint-missing recovery (1)
   ✓ cron isolation / fail-closed rollover preservation (2)
```

Full test suite: 1424 passed, 8 failed (all pre-existing failures unrelated to this change).

### Files changed (this branch)
- `src/tools/lcm-conversation-scope.ts` -- export `isIsolatedCronSessionKey` (+1 line)
- `src/transcript-reconciler.ts` -- relaxed guard (+3 lines)
- `test/isolated-cron-checkpoint-missing-race.test.ts` -- regression tests (357 lines)
- `.changeset/cron-sessionid-recovery.md` -- patch changeset

ralf003:
@jalehman Gentle ping on this — it's been 7 days since I added the reconciler-branch regression tests (exercising the full afterTurn flow). Could you take another look at your convenience?

Summary:
- CI is all green (test, smoke-latest-openclaw, plugin-inspector)
- Mergeable: clean
- The tests now cover both the cron recovery path (main test) and the non-cron fail-closed path directly through the engine, addressing your concern about stronger proof

Thanks!

jalehman:
Thanks for the contribution, @ralf003. This is landed now.

PR #948, opened 2026-06-30
chore: codify LF line endings

### LINKED ISSUE #933: Codify LF line endings with .gitattributes?
## Local CRLF pollution observed at `main` HEAD `d404eba6`

A working tree of this repo at `main` HEAD `d404eba6c71df4fa387560ab1188a58c3bda6f4f` was dirty with 242 files showing as modified against HEAD, with `git diff --stat` reporting equal insertions and deletions (133,565 each).

Root cause: mass LF → CRLF re-encoding on disk. Every "modification" was the addition of a `\r` byte at every line terminator. Verified by `git diff --ignore-cr-at-eol --exit-code` returning exit 0 (no content differences once CR is normalized) and HEAD-vs-disk byte deltas matching exactly the per-line CR insertion count.

`file src/assembler.ts` reported "UTF-8 text, with CRLF line terminators" in the polluted working tree, while a sister worktree at a different HEAD on a different branch reported plain "UTF-8 text" with no CRLF. So the canonical state of `main` HEAD is LF — the pollution was local to the affected machine (likely a Windows-side editor or `core.autocrlf=true` setting that was later removed but not undone).

## Question

Would you accept a one-line PR adding `.gitattributes` with:

```gitattributes
* text=auto eol=lf
```

This would codify the LF convention that HEAD already follows, prevent accidental CRLF-only working tree churn on other platforms, and not change any committed content.

Happy to send the PR if useful. Ignore if not desired.

No external product details to share; this is general housekeeping.

dashitongzhi:
## Summary
- add `.gitattributes` rules to normalize text files to LF
- explicitly cover common source, docs, config, template, lockfile, and binary file types
- document the repository line-ending policy in the README development section

Fixes #933

## Test plan
- `git diff --check`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,44 @@
+# Normalize text files to LF in the repository.
+# Git will store these files with LF line endings.
+* text=auto eol=lf
+
+# Source files
+*.ts text eol=lf
+*.tsx text eol=lf
+*.js text eol=lf
+*.jsx text eol=lf
+*.mjs text eol=lf
+*.cjs text eol=lf
+*.go text eol=lf
+
+# Documentation and structured data
+*.md text eol=lf
+*.json text eol=lf
+*.jsonl text eol=lf
+*.yaml text eol=lf
+*.yml text eol=lf
+*.toml text eol=lf
+
+# Shell and config files
+*.sh text eol=lf
+*.bash text eol=lf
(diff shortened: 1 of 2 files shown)


jalehman:
Thanks for the contribution, @dashitongzhi. This is landed now.

PR #950, opened 2026-06-30
fix(interceptor): recover full file content when read tool truncates output

mpz4life:
## Problem

When OpenClaw's built-in `read` tool reads a file:

1. The base truncation cap (~8KB / 256 lines) cuts the output before it reaches the interceptor.
2. The adaptive paging wrapper (`createOpenClawReadTool`) further caps aggregated text at ~8KB (adjusted by context window).
3. The large-file interceptor externalizes the tool result — the **already-truncated** text is all that gets saved to `large_files`.
4. The agent calls `lcm_describe` to drill down — receives only the truncated fragment. **The full file content is permanently lost.**

## Solution

In `LargeFileInterceptor.interceptLargeToolResults`, when a `read` tool result contains a truncation marker (`[Read output capped at...]` or `[Truncated:...]`):

1. The interceptor looks up the tool call input via a `toolCallInputMap` (built from the ordered message batch) to get the original file path.
2. If the path is absolute and exists on disk, the file is re-read in full via `readFileSync`.
3. The full content replaces the truncated text before externalization.
4. Any subsequent `lcm_describe` read returns the complete file.

### Changes

- `tool-pairing.ts`: new `buildToolCallInputMap(messages)` — maps `toolCallId → input` from assistant `tool_use` blocks.
- `large-file-interceptor.ts`: `interceptLargeToolResults` now accepts optional `toolCallInputMap`; detects read truncation and recovers full content.
- `engine.ts`: both `ingest` and `assemble` paths build and pass the tool call input map to the interceptor.

## Verification

- `pnpm typecheck` passes.
- 186 related tests pass.

### DIFF
--- a/.changeset/read-truncation-recovery.md
+++ b/.changeset/read-truncation-recovery.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-When OpenClaw's built-in `read` tool returns truncated output, recover the full file content before externalizing the oversized tool result — but only for the live `assemble()` current turn. The truncated text is preserved on ingest, bootstrap, and replay paths so transcript fidelity is not compromised by current disk state. Fallback to the truncated fragment when the original path is missing, relative, or unreadable.
+When OpenClaw's built-in `read` tool returns truncated output, recover the full file content before externalizing the oversized tool result — but only for the live `assemble()` current turn. The truncated text is preserved on ingest, bootstrap, and replay paths so transcript fidelity is not compromised by current disk state. Fallback to the truncated fragment when the original path is missing, relative, unreadable, non-regular, or too large for bounded live recovery.
--- a/.changeset/read-truncation-recovery.md
+++ b/.changeset/read-truncation-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+When OpenClaw's built-in `read` tool returns truncated output, recover the full file content before externalizing the oversized tool result — but only for the live `assemble()` current turn. The truncated text is preserved on ingest, bootstrap, and replay paths so transcript fidelity is not compromised by current disk state. Fallback to the truncated fragment when the original path is missing, relative, or unreadable.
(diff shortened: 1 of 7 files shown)


jalehman:
Thanks for the patch. The problem is real and a good fit for Lossless: once truncated `read` output is externalized, `lcm_describe` cannot recover what was lost.

I think this needs revision before review, though. The current implementation re-reads the current host filesystem from the original tool input path. That is risky for transcript fidelity, especially through `ingestBatch`/replay/import paths, because a historical truncated read can be replaced with whatever is at that path now rather than what the tool returned at the time.

Could you tighten this to avoid historical rehydration from current disk state, add regression coverage for the recovery path through ingest/assemble and `lcm_describe`, and add a patch changeset? If re-reading remains part of the design, it should be explicitly limited/guarded to the live current-turn case and should document the fallback behavior when the path is missing, relative, or unreadable.

mpz4life:
@jalehman Thanks for the feedback. I've revised the change to scope read-tool truncation recovery strictly to the live current-turn assemble() path, so historical ingest/replay/bootstrap paths no longer rehydrate from current disk state.
Changes made:
1. Recovery only happens in assemble() live messages. ingest(), ingestBatch(), afterTurn(), bootstrap(), and replay/import paths no longer pass a toolCallInputMap, so truncated text is persisted verbatim.
2. buildToolCallInputMap now captures both tool name and input, since tool result blocks often omit the tool name and we need the paired tool_use name to identify read calls.
3. Recovery is gated by the presence of toolCallInputMap as the live current-turn signal. It only triggers when the effective tool name is read, a truncation marker is present, the original path is absolute, and the file is readable.
4. Fallback to truncated text when the path is missing, relative, or unreadable.
5. Fixed file-id hashing to use the recovered full content for getFileId, avoiding dedup drift.
6. Added regression tests in test/read-tool-truncation-recovery.test.ts:
- assemble() recovers the full file.
- afterTurn() ingest preserves the truncated text.
- bootstrap() does not rehydrate from current disk state.
- Relative/missing paths fall back to the truncated fragment.
7. Added a patch changeset.
Please review.

jalehman:
Thank you!

PR #951, opened 2026-06-30
feat(lcm): add programmatic control handlers

SYU8384:
## Summary
- exports programmatic LCM `status`, `doctor`, and `rotate` handlers
- shares the native command implementation instead of shelling out or parsing transcript text
- keeps handler outputs sanitized: no transcript content, database paths, credentials, provider debug, or local filesystem details

## Context
This is a small optional integration surface for developers who want to embed lossless-claw behind another app, gateway, admin tool, CI smoke, or control plane.

The existing native command behavior remains the source of truth. These handlers make the same behavior callable without scraping command text, shelling out, or exposing transcript/local runtime details.

The handlers are intentionally generic and are not tied to any downstream application.

## Validation
- `git fetch upstream main`
- `git diff --check`
- `npm run typecheck`
- `npm run build`
- `npm test -- lcm-command.test.ts`

Additional downstream smoke coverage verified these handlers through gateway control endpoints with status, doctor, rotate, post-rotate state, rate limiting, two-session isolation, and sanitized failure output.

### DIFF
--- a/.changeset/quiet-control-plane.md
+++ b/.changeset/quiet-control-plane.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Document the optional programmatic status/doctor/rotate control surface and keep status free of non-durable rotation timestamps.
--- a/README.md
+++ b/README.md
@@ -60,6 +60,18 @@ Not currently supported as root CLI commands:
 
 The bundled skill focuses on configuration, diagnostics, architecture, and recall-tool usage. Its reference set lives under `skills/lossless-claw/references/`.
 
+### Programmatic control
+
+Lossless-claw also exposes an optional host-facing context-engine control contract for OpenClaw gateways that support context-engine capabilities and control dispatch. The contract is intentionally smaller than the native slash command surface:
+
+- `status` returns whether an LCM conversation is active and the current stored message count.
+- `doctor` returns a bounded, sanitized warning list for summary-health issues.
+- `rotate` runs the same safe transcript rotation path as `/lossless rotate` and returns the post-rotate message count plus the timestamp for that successful rotate operation.
+
+Programmatic control never returns transcript text, local database paths, backup paths, credentials, provider debug, or shell output. `status` does not report `lastRotatedAt`; hosts that need durable rotation timestamps should persist that product state themselves after a successful `rotate` result.
+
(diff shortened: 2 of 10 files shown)


jalehman:
Thanks for this. The shape is a good fit: structured status/doctor/rotate handlers are a useful way to support downstream control planes without scraping `/lossless` output.

I’d like to hold this before deeper review for two issues:

1. This adds a new user/developer-facing programmatic surface, so it needs a patch changeset and docs. Because it appears to depend on a host-side context-engine control capability, please also verify the first stable OpenClaw release that supports that contract and update the compatibility metadata if needed.
2. `lastRotatedAt` is currently kept in a module-global map keyed only by `conversationId`. Conversation IDs are database-local, so multiple OpenClaw state dirs/profiles in one process can collide, and the value disappears on restart. Please either make this state durable/profile-scoped or remove it from the status contract.

Once those are addressed, this looks worth a full maintainer review.

SYU8384:
> Thanks for this. The shape is a good fit: structured status/doctor/rotate handlers are a useful way to support downstream control planes without scraping `/lossless` output.
> 
> I’d like to hold this before deeper review for two issues:
> 
> 1. This adds a new user/developer-facing programmatic surface, so it needs a patch changeset and docs. Because it appears to depend on a host-side context-engine control capability, please also verify the first stable OpenClaw release that supports that contract and update the compatibility metadata if needed.
> 2. `lastRotatedAt` is currently kept in a module-global map keyed only by `conversationId`. Conversation IDs are database-local, so multiple OpenClaw state dirs/profiles in one process can collide, and the value disappears on restart. Please either make this state durable/profile-scoped or remove it from the status contract.
> 
> Once those are addressed, this looks worth a full maintainer review.

Thanks, addressed in `945ca1e6959bfffb4b108e095a184b2a0c757513`.

Changes:
1. Added a patch changeset and docs for the optional programmatic `status` / `doctor` / `rotate` surface.
2. Removed `lastRotatedAt` from the `status` contract and deleted the module-global timestamp map. `rotate` still returns `lastRotatedAt` for the successful rotate operation.
3. Checked OpenClaw compatibility: the latest stable OpenClaw release, `v2026.6.11`, does not yet include the required host-side context-engine control endpoints. That support is proposed in `openclaw/openclaw#98060`, so I documented this as optional and host-capability-gated rather than raising the OpenClaw peer dependency or compatibility metadata.

CI is green on the updated branch.

jalehman:
Thanks for the contribution, @SYU8384. This is landed now.

PR #952, opened 2026-06-30
feat: migrate to sqlite and add pending summary compaction

jalehman:
## What

Migrates Lossless Claw onto the OpenClaw SQLite session/transcript runtime from openclaw/openclaw#98236, then adds pending-summary compaction on top of that projection-backed runtime.

## Why

Pending compaction needs a stable host-owned transcript projection so background LLM work can prepare summaries without exposing partial state or racing active JSONL files. OpenClaw #98236 moves session metadata and transcript events into SQLite, so this PR removes Lossless-owned active transcript file maintenance and uses the host projection as the runtime boundary.

## Changes

- Read visible SQLite transcript projection
- Remove active transcript file maintenance
- Add verified transcript anchor trust
- Add pending summary DAG tables
- Prepare pending summaries under leases
- Publish ready frontiers atomically
- Route background debt through pending flow
- Route manual compact through pending flow
- Preserve lifecycle rebind/reset semantics
- Update docs, config, changesets

## Testing

- `npm run typecheck`
- `./node_modules/.bin/vitest run test/pending-summary-worker.test.ts test/pending-summary-store.test.ts test/pending-summary-coordinator.test.ts test/pending-summary-planner.test.ts test/pending-summary-publisher.test.ts test/pending-summary-compaction-proof.test.ts test/engine-pending-e2e.test.ts test/engine-after-turn.test.ts test/engine-maintenance.test.ts test/engine-lifecycle.test.ts test/sqlite-bootstrap-projection.test.ts test/transcript-anchor-audit.test.ts test/transcript-anchor-trust-store.test.ts test/migration.test.ts test/lcm-command.test.ts test/lcm-tools.test.ts test/index-runtime-llm-complete.test.ts test/plugin-config-registration.test.ts test/package-metadata.test.ts test/migrate-sessions.test.ts`
- `$HOME/.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main`
- `git diff --check`

## Dependency

Depends on openclaw/openclaw#98236 for the host-owned SQLite session/transcript runtime and visible transcript projection API.


islandpreneur007:
Real-world reproduction that this PR's direction addresses — reporting it here rather than as a separate issue, since it looks like exactly the boundary this migration moves.

**Environment**
- lossless-claw `0.11.3` (also tested `0.13.2` and `0.14.0`)
- OpenClaw core `2026.7.2-beta.2`
- Codex app-server harness, single main agent, Node 26, Linux (systemd-managed gateway)

**Symptom**

On `2026.7.2-beta.2`, the host supplies a virtual SQLite session locator rather than a real path — `sessionFile` arrives as a `sqlite:` marker, with the actual identity carried in `sessionTarget`. Core bootstrap still inspects `sessionFile` directly, so it raises:

```
failed during bootstrap: ENOENT: stat 'sqlite:main:<uuid>:/.../sessions/sessions.json'
```

The engine is then quarantined for the process and the host silently selects `legacy`.

**What makes it hard to notice**

It fails **once per process, then goes silent** ("quarantined for this process"). Every health surface stays green — healthz, `rpc.ok`, version string — and `lcm.db` keeps looking alive because the legacy path continues touching metadata rows. We read the silence as health for roughly 19 hours. This is the same detection gap described in #971, arriving via a different cause.

**Version testing**

We rehearsed on disposable clones of a SQLite-consistent copy of our live `lcm.db`:

| version | result |
|---|---|
| 0.11.3 | ENOENT → quarantined → legacy |
| 0.13.2 | identical |
| 0.14.0 | identical |

Every stable tarball from 0.11.3 through 0.14.0 retains the direct session-file inspection in core bootstrap. Pass condition was **compaction ownership on a real second turn**, not "loaded" and not "no error on startup" — first-turn startup and plugin-loaded are both false greens here.

`migrate-sessions` is not applicable: it is a dry-run-by-default JSONL backfill (0 eligible files for us) and neither reads the host's shared session SQLite nor translates the `sqlite:` marker.

**Compatibility datapoint for this PR**

0.14.0 opened a clone of our live `lcm.db` cleanly: 0 conversations and 0 messages deleted, summaries stable at 124, 8 user-message identity hashes repaired, one migration marker added. So the DB-side upgrade path looks safe in our shape, independent of the locator issue.

**Independently reproduced** on a second, unrelated deployment (different host, different agent count) — there it affected 4 agents and ran ~8h before detection, so blast radius scales with agent count.

Happy to test a branch build against this shape and report back if that would be useful.

PR #953, opened 2026-06-30
fix: skip summary thinking for Ollama

### LINKED ISSUE #944: Ollama extended thinking models place entire summary in reasoning block, resulting in empty summary after parsing
```markdown
## Problem

When using `enableSummaryThinking: true` with Ollama's extended thinking models (e.g., `qwen3.5:9b`, `qwen3.6:27b`), the summarizer sends `reasoningIfSupported: "low"` to the API. Models with extended thinking support return the summary content entirely within a `type: "reasoning"` block, leaving the `type: "text"` block empty.

The response parsing logic (`collectTextLikeFields`) skips all blocks where `isReasoningLikeType(value.type)` is true (summarize.ts L351), resulting in an empty summary string after collecting all content entries.

## Root Cause

**summarize.ts L1612-1614:**
```typescript
...(params.deps.config.enableSummaryThinking !== false
  ? ({ reasoningIfSupported: "low" } as const)
  : {}),
```

This enables extended thinking mode for the summarizer model. When Ollama's qwen3.x models receive this flag, they structure the response as:

```json
{
  "content": [
    { "type": "reasoning", "text": "<~4265 tokens of actual summary content>" },
    { "type": "text", "text": "" }
  ]
}
```

**summarize.ts L351:**
```typescript
if (isReasoningLikeType(value.type) || isReasoningLikeType(value.rawType)) {
  return; // reasoning blocks are skipped entirely
}
```

Since `type: "reasoning"` matches `isReasoningLikeType`, the entire summary content in the reasoning block is discarded. The text block is empty, so the final collected summary is `""`.

## Impact

- First attempt returns empty → envelope fallback runs → retry with doubled maxTokens → still produces reasoning-only output → eventually falls through to truncation fallback (original conversation text used as summary)
- Each compaction cycle wastes ~2 API calls and significant token budget before hitting the deterministic fallback
- For models that consistently produce reasoning-only output, `enableSummaryThinking` is effectively useless

## Environment

- lossless-claw version: latest (commit 0681006)
- Provider: Ollama (remote)
- Model: qwen3.5:9b (and likely affects other extended-thinking-capable models)
- `enableSummaryThinking`: true (default)

## Suggested Fix Options

**Option A — Extract reasoning block text as fallback when text is empty:**
When `collectTextLikeFields` collects no text from non-reasoning blocks, fall back to extracting content from reasoning blocks before declaring the summary empty. This preserves the model's output quality while handling structural separation.

**Option B — Add a configurable system prompt injection point:**
Allow users to append custom instructions to `LCM_SUMMARIZER_SYSTEM_PROMPT` via config or environment variable (e.g., `summarySystemPromptAppend`). This enables instructing thinking-capable models to re-state their summary in the text block after reasoning.

**Option C — Document and default disable for Ollama:**
Add a provider/model-aware heuristic that disables `reasoningIfSupported` for known providers (Ollama extended thinking) where the structural separation causes parsing issues, or document this limitation prominently.

## Workaround

Set `"enableSummaryThinking": false` in the lossless-claw plugin config to disable the flag and force models to return content directly in text blocks.

cxbAsDev:
## Problem

When `enableSummaryThinking: true` is used with Ollama extended-thinking models, the summarizer can request low reasoning via `reasoningIfSupported`. Those models may then return the usable summary inside a `type:"reasoning"` block while leaving the `type:"text"` block empty. Lossless intentionally treats typed reasoning/thinking blocks as private diagnostics, so the summary normalizes to empty and falls through to retry/fallback behavior.

Closes #944.

## Fix

Do not request summary-thinking controls for resolved Ollama summarizer candidates. This keeps Ollama summaries on the normal text path without reclassifying typed reasoning blocks as persisted summary content. Reasoning-only and mixed reasoning/empty-text responses remain private and fall back instead of being promoted.

## Testing

- `npm test -- --run test/summarize.test.ts -t "Ollama|reasoning blocks|thinking-only"`
- `npm test -- --run test/summarize.test.ts`
- `npm run typecheck`
- `npm test`
- `autoreview --mode branch --base origin/main` clean

### DIFF
--- a/.changeset/reasoning-block-text-fallback.md
+++ b/.changeset/reasoning-block-text-fallback.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Extract summary content from typed reasoning blocks when a text-type block is also present in the response, so Ollama extended-thinking models that place the entire summary inside a `type:"reasoning"` block produce usable summaries instead of falling through to the truncation fallback. Reasoning-only responses without an accompanying text block remain treated as private diagnostics.
+Avoid requesting summary-thinking controls for Ollama summarizer calls, so extended-thinking models return usable text summaries without promoting typed reasoning blocks into persisted summary content.
--- a/.changeset/reasoning-block-text-fallback.md
+++ b/.changeset/reasoning-block-text-fallback.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Extract summary content from typed reasoning blocks when a text-type block is also present in the response, so Ollama extended-thinking models that place the entire summary inside a `type:"reasoning"` block produce usable summaries instead of falling through to the truncation fallback. Reasoning-only responses without an accompanying text block remain treated as private diagnostics.
--- a/src/summarize.ts
+++ b/src/summarize.ts
@@ -337,10 +337,14 @@ function shouldAppendDirectTextField(key: string): boolean {
 }
 
 /** Collect text payloads from common provider response shapes. */
-function collectTextLikeFields(value: unknown, out: string[]): void {
+function collectTextLikeFields(
+  value: unknown,
+  out: string[],
+  skipReasoning = true,
+): void {
   if (Array.isArray(value)) {
     for (const entry of value) {
-      collectTextLikeFields(entry, out);
+      collectTextLikeFields(entry, out, skipReasoning);
     }
     return;
   }
@@ -348,7 +352,10 @@ function collectTextLikeFields(value: unknown, out: string[]): void {
     return;
   }
 
-  if (isReasoningLikeType(value.type) || isReasoningLikeType(value.rawType)) {
+  if (
+    skipReasoning &&
+    (isReasoningLikeType(value.type) || isReasoningLikeType(value.rawType))
+  ) {
     return;
   }
 
@@ -367,7 +374,7 @@ function collectTextLikeFields(value: unknown, out: string[]): void {
         }
         continue;
       }
-      collectTextLikeFields(nested, out);
+      collectTextLikeFields(nested, out, skipReasoning);
     }
   }
 }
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for tracking this down. The underlying problem is real and worth fixing, but I think this solution is too broad as written.

Lossless currently treats typed `reasoning` / `thinking` payloads as diagnostic/private rather than summary text, and this PR changes the existing redaction test so `PRIVATE_TYPED_REASONING_TRACE` becomes the persisted summary. That could leak model-internal reasoning into stored summaries/context for providers where the reasoning block is not actually the user-visible answer.

Can you narrow the fix so it addresses the Ollama extended-thinking shape without generally reclassifying typed reasoning blocks as summary content? A provider/model-gated path, an explicit adapter-level signal, or another shape-specific distinction would be safer. This will also need a patch changeset if we merge it because it changes summarizer behavior.

cxbAsDev:
Thanks for the review — narrowed the fix per your feedback.

**What changed:**
The fallback now only activates when the response contains **both** reasoning-type AND text-type blocks (`blockTypeSet.has("text")`). This targets the Ollama extended-thinking shape specifically:

```json
// Ollama qwen3.x response → fallback triggers (has "text" block)
{ "content": [
    { "type": "reasoning", "text": "<actual summary>" },
    { "type": "text", "text": "" }
] }

// Private reasoning → still treated as diagnostic (no "text" block)
{ "type": "reasoning", "summary": [{ "type": "summary_text", "text": "..." }] }
```

The existing "redacts typed reasoning blocks from diagnostic previews" test is fully restored — `PRIVATE_TYPED_REASONING_TRACE` is never persisted as summary text.

Also added a filter for empty-string chunks from empty text blocks (`chunks.every(c => !c.trim())`) and a patch changeset (`reasoning-block-text-fallback.md`).

jalehman:

Issue #754, opened 2026-05-27
Bootstrap guard incorrectly triggers conversation replacement when session file is renamed to .reset. suffix


feel-think:
Problem

When OpenClaw restarts or a /new is issued, OpenClaw may assign a new sessionId (new UUID) to the session and rename the old session JSONL file to {original}.reset.{timestamp} for archival. LCM's bootstrap guard (in engine.ts, around line 5788 in v0.11.2 / line 3994 in v0.9.2) detects the sessionId change and the different file path, then calls stat(trackedSessionFile) to verify the old file still exists. Since the old file has been renamed, stat returns ENOENT, causing the guard to treat this as a "missed /reset" and call applySessionReplacement, which archives the old conversation and creates a new one — losing conversation continuity.

Root cause

The bootstrap guard has four trigger conditions:

1. Same sessionKey (stable identity for the chat/thread)
2. Different sessionId (new UUID after restart or /new)
3. transcriptRotated = true (old file path != new file path, because UUID is embedded in the filename)
4. trackedSessionFileMissing = true (stat(trackedSessionFile) returns ENOENT because the file was renamed)

When all four are true, the guard calls applySessionReplacement. The guard does NOT check whether a .reset. backup of the file exists in the same directory. If a .reset. file is found, the file was merely archived by OpenClaw, not truly lost — the guard should not fire.

Also worth noting: getOrCreateConversation (called immediately after the guard) already handles sessionId changes correctly — it finds the existing conversation by sessionKey and updates the stored sessionId in place. The guard is solving a problem that does not need solving in this scenario.

Affected versions

• v0.9.2 (confirmed in production)
• v0.11.2 (confirmed by source code review — identical guard logic)

Note: v0.11.2 added "restart" and "shutdown" to the handleSessionEnd skip list (PR #657), but this does not affect the bootstrap guard, which operates during message processing, not during session lifecycle events.

Reproduction steps

1. Configure a non-main agent (e.g., a chat agent)
2. Restart the OpenClaw gateway (or trigger an abrupt restart via OOM kill)
3. The session is recreated with a new UUID and the old session JSONL file is renamed to .reset.
4. Send a message to that agent
5. Observe that LCM creates a new conversation instead of reusing the existing one

Observed impact

• Main agent: lower probability (more persistent sessions)
• Sub-agents (e.g., chat agent): higher probability (sessions are more likely to be recreated with new UUIDs on restart)

[...]

PR #954, opened 2026-06-30
fix(rollover): demote happy-path ambiguous-rollover rebind logs from warn

gorkem2020:
## Problem

The ambiguous session-key rollover recovery path logs its happy path at warn. A routine soft reset (`/new`) that the guard recovers cleanly emits a burst of warn lines (the per-phase preserve restatement, the successful rebind, the not-provably-fresh decision, the new-epoch import, a benign frontier skip), even though nothing went wrong. That noise drowns out the warn level for the cases that actually need attention.

## Fix

Level the recovery logs off the freshness disposition carried out of the rebind attempt:

- The successful fresh-transcript rebind and the not-provably-fresh decision log at info.
- The assemble pass's per-phase preserve restatement logs at debug, since it defers freshness judgment to the next bootstrap or afterTurn pass.
- The bootstrap and afterTurn preserve log keys off the verdict: a transient or unjudgeable freshness state (no usable timestamps, delivery-only traffic, nothing comparable) is a pending state that the next turn re-evaluates and logs at debug, while a conflicting verdict (identity overlap, or candidate entries predating persistence) is a genuine freeze and stays at warn.
- The rebind-failed and freshness-check exception paths, the no-anchor import-cap aborts, and every non-rollover unsafe-to-advance frontier skip stay at warn.

The freeze and no-merge protection is unchanged throughout. Only log levels move.

## Tests

Adds `test/rebind-loglevel.test.ts` and updates the rotation, after-turn, fidelity, and maintenance suites to assert the new levels: debug or info on the benign recovery path, warn retained on conflict, freeze, and rebind failure. Includes a changeset.

## Note

Companion to the conversation-preservation fix #955, which fixes the underlying #754 destruction bug. The two are independent; whichever merges second needs a trivial rebase.

### DIFF
--- a/.changeset/quiet-ambiguous-rollover-rebind-logs.md
+++ b/.changeset/quiet-ambiguous-rollover-rebind-logs.md
@@ -1,0 +1,7 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Demote the happy-path ambiguous session-key rollover recovery logs from warn to info or debug, context-aware. The successful fresh-transcript rebind and the not-provably-fresh decision now log at info; the assemble pass's per-phase preserve restatement logs at debug; the fresh-rebind new-epoch import logs at info, and the afterTurn "frontier not covered" line logs at debug only on the benign ambiguous-rollover path.
+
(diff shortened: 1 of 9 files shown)


jalehman:

Issue #262, opened 2026-04-04
/new (session reset) should not carry over compacted summaries


chungjchris:
## Summary

`/new` (or `/reset`) creates a fresh session but all compacted LCM summaries are injected into the new session, defeating the purpose of starting fresh.

## Root Cause

`getOrCreateConversation()` in `conversation-store.ts` keys on `sessionKey` (e.g. `agent:main:telegram:direct:<userId>`), not `sessionId`. When `/new` fires:

1. A new `sessionId` (UUID) is generated
2. The `sessionKey` stays the same (it is channel-scoped, permanent)
3. LCM looks up by `sessionKey` first, finds the existing conversation, updates the `sessionId` column, and returns the **same conversation** with all its summaries

This means `/new` can never produce a truly empty context for any persistent channel (Telegram DM, Slack DM, etc.). All historical compacted summaries (~43k+ tokens in my case) are always re-injected.

## Expected Behavior

`/new` should start with zero LCM context. The summaries should still exist in the database (lossless), but should not be assembled into the new session's context unless the user explicitly requests recall via tools (lcm_grep, lcm_expand_query, etc.).

## Suggested Fix

On session rollover (new `sessionId` for existing `sessionKey`):
- Create a new conversation record (or mark a rollover boundary on the existing one)
- Do not carry forward context items from the previous conversation into the assembler
- Keep the old conversation and its summaries accessible via LCM tools for cross-conversation search

## Environment

- lossless-claw installed from git (Martian-Engineering/lossless-claw)
- OpenClaw 2026.3.31 (213a704)
- Channel: Telegram DM (persistent sessionKey)

PR #955, opened 2026-06-30
fix(rollover): preserve conversation on soft reset via archive-sibling probe

### LINKED ISSUE #754: Bootstrap guard incorrectly triggers conversation replacement when session file is renamed to .reset. suffix

gorkem2020:
Fixes #754.

## Problem

On a soft reset (`/new`), the host renames the active session file to a `.reset.` suffix and starts a fresh session under the same session key. The bootstrap rollover guard sees the tracked transcript as missing and treats it as an ambiguous session-key rollover. With no signal that the rename was deliberate, the guard archives the live conversation and mints a fresh empty one, so the conversation is replaced and its active context is lost.

## Fix

Before treating the vanished transcript as an anomaly, probe for an archived sibling of the tracked session file. If a `.reset.` or `.deleted.` sibling exists, the rename was a deliberate operator reset, not a crash or a genuine key collision, so the guard stands down and rebinds the existing conversation to the new session instead of archiving it. The conversation is preserved and continues under the new session id.

When no archived sibling exists, behavior is unchanged, so genuine rollover and crash-recovery paths keep their existing protection.

## Tests

Adds `test/soft-reset-new-sibling-rebind.test.ts` covering: a deliberate `.reset.` sibling rebinds in place; a foreign reused key with no sibling does not rebind; and the identity-overlap conflict case still preserves rather than merges. Includes a changeset.

## Alignment with documented behavior

This restores a contract the docs already state. From `docs/configuration.md`:

> `/new` keeps the active LCM conversation and prunes active context according to `newSessionRetainDepth`
> `/reset` archives the active conversation row and creates a fresh active row for the same stable `sessionKey`

The destruction described above breaks that documented `/new` contract by archiving and replacing the conversation. This change makes the guard match the docs. It does not alter context pruning (still governed by `newSessionRetainDepth`), so the "start `/new` with a pruned context" intent from #262 is unchanged; only the conversation row is preserved rather than stranded.

## Note

This touches `src/session-rollover.ts`, which also moves in the companion log-level cleanup PR #954. The two are independent; whichever merges second needs a trivial rebase.

### DIFF
--- a/.changeset/preserve-conversation-on-soft-reset.md
+++ b/.changeset/preserve-conversation-on-soft-reset.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve the active conversation on a soft reset (/new). When the session file is renamed to a .reset. or .deleted. sibling and a fresh session starts under the same session key, the bootstrap rollover guard now probes for the archived sibling, recognizes the rename as a deliberate reset, and rebinds the existing conversation to the new session instead of archiving it and minting an empty one. Genuine rollover and crash-recovery paths (no archived sibling) are unchanged.
--- a/.changeset/preserve-conversation-on-soft-reset.md
+++ b/.changeset/preserve-conversation-on-soft-reset.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Preserve the active conversation on a soft reset (/new). When the session file is renamed to a .reset. or .deleted. sibling and a fresh session starts under the same session key, the bootstrap rollover guard now probes for the archived sibling, recognizes the rename as a deliberate reset, and rebinds the existing conversation to the new session instead of archiving it and minting an empty one. Genuine rollover and crash-recovery paths (no archived sibling) are unchanged.
+Preserve the active conversation on a soft reset (/new). When the session file is renamed to a .reset. sibling and a fresh session starts under the same session key, the bootstrap rollover guard now probes for the archived sibling, recognizes the rename as a deliberate reset, and rebinds the existing conversation to the new session instead of archiving it and minting an empty one. Genuine rollover and crash-recovery paths (no archived sibling) are unchanged.
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -413,6 +413,9 @@ function ensureConversationBootstrapStateEpochColumns(db: DatabaseSync): void {
   if (!columns.some((col) => col.name === "last_processed_entry_id")) {
     db.exec(`ALTER TABLE conversation_bootstrap_state ADD COLUMN last_processed_entry_id TEXT`);
   }
+  if (!columns.some((col) => col.name === "soft_reset_pruned_at")) {
+    db.exec(`ALTER TABLE conversation_bootstrap_state ADD COLUMN soft_reset_pruned_at TEXT`);
+  }
 }
 
 function backfillMessageIdentityHashes(
@@ -1190,6 +1193,7 @@ export function runLcmMigrations(
       last_processed_entry_id TEXT,
       fork_bounded INTEGER NOT NULL DEFAULT 0,
       fork_source_message_count INTEGER NOT NULL DEFAULT 0,
+      soft_reset_pruned_at TEXT,
       updated_at TEXT NOT NULL DEFAULT (datetime('now'))
     );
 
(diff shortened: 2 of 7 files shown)


gorkem2020:
The companion log-level PR [#954](https://github.com/Martian-Engineering/lossless-claw/pull/954) merged, and this PR was closed in the same moment, but the fix itself did not land: the archive-sibling probe and `test/soft-reset-new-sibling-rebind.test.ts` are not on main, so the [#754](https://github.com/Martian-Engineering/lossless-claw/issues/754) destruction path is still live. Was the close intentional?

The branch is now rebased onto current main (the only adjustment was aligning one log-level assertion with #954) and is green, 45/45 across this suite and the adjacent rollover suites. If the close was just the pair being treated as one, reopening picks it up as-is; happy to open a fresh PR instead if that is easier.

**Update (July 4):** I tried to reopen this from my side and GitHub refuses (the close came from the maintainer side and the branch received a post-close update, either of which blocks an author reopen). Since the close landed in the same second as the [#954](https://github.com/Martian-Engineering/lossless-claw/pull/954) merge, I am assuming it was an accidental batch action rather than a rejection. If reopening works from your side, the branch is current and includes your narrowing commit, so it is in your requested shape. If it is blocked for you too, I am happy to open a fresh PR carrying the same commits and linking back here for the review history. For what it is worth, this fix has been running in our production deployment since June 29 with zero unintended rollovers, while the [#754](https://github.com/Martian-Engineering/lossless-claw/issues/754) destruction path remains live on main. And happy to drop it entirely if the close was intentional.

PR #957, opened 2026-07-01
fix(doctor): never auto-restore a deliberately archived conversation

gorkem2020:
## Problem

The rollover-split doctor (`/lossless doctor apply rollover-splits`) merges archived "stranded" conversation rows back into the active conversation for a session key, recovering conversations that an abrupt crash or session rollover accidentally split off. Its classifier (`classifySessionKeyGroup`) keys only on `active`, `archived_at`, chronology, and stranded-data counts.

But a deliberate operator `/reset` also archives the active conversation (and starts a fresh one), and it archives through the exact same `applySessionReplacement` to `archiveConversation` funnel as a crash rollover. The `reason` is passed in but never persisted, and no column records why a row was archived. So the doctor cannot tell a deliberate `/reset` from a crash, and it resurrects conversations the user intentionally wiped, re-merging their messages back into the active context.

## Fix

Persist the archive provenance. A new nullable column `conversations.archive_cause` is written at the single archive funnel with a normalized cause:

`manual-reset | session-deleted | session-end | rollover-fallback | cron-rotation`

The doctor then excludes deliberately-archived rows from its merge `sources`:

```ts
export const DELIBERATE_ARCHIVE_CAUSES = new Set(["manual-reset"]);
// sources filter:
row.active === 0 && hasStrandedData(row) && !isDeliberateArchive(row)
```

The exclusion is per-row, not per-group: a session key holding both a deliberate `manual-reset` row and a genuine `rollover-fallback` row still merges the rollover row into the active target while leaving the reset row untouched. If the filter empties the source set, the classifier already returns no finding, so the normal "reset then fresh successor" shape produces no doctor noise.

## Conservative scope

Ships with `manual-reset` as the only skipped cause. A `/reset` is unambiguous operator intent. `session-deleted` is recorded but left merge-eligible for now, because it depends on the host emitting its delete reason only for deliberate deletes (not automated tombstone or cleanup pruning); promoting it is a one-line follow-up once that contract is confirmed. Every other cause stays merge-eligible, so the safety-critical direction is preserved: the doctor never withholds recovery of a genuine broken split.

## Backward compatibility

The column is additive and nullable (no default, no index, no backfill). Existing archived rows read NULL, which is treated as unknown provenance and stays merge-eligible (today's behavior). The only behavioral change is the doctor declining to merge rows positively marked as a deliberate reset, which can only be rows archived after this change. A `/reset` archived before deploy can still be over-restored at most once; the gap self-heals forward as new resets carry the marker.

The archive write uses `archive_cause = COALESCE(archive_cause, ?)`, so the first recorded cause is retained if a conversation were ever archived twice. In practice the funnel only re-archives an active conversation and a reset cannot re-touch an already-archived row, so this is the protective direction and does not arise in normal operation.

## Host-contract dependency

The deliberate-vs-incidental distinction relies on the host emitting its lifecycle reason strings (`reset`, `deleted`) only for genuine operator actions. A real `/reset` surfaces as **both** a `before_reset(reason=reset)` and a `session_end(reason=reset)`, so both lifecycle handlers map the reset reason to `manual-reset`; the `COALESCE` write keeps the cause stable whichever event closes the conversation first. Incidental `session_end` reasons (`idle`, `daily`, `compaction`) stay `session-end` and remain merge-eligible, and `new`/`restart`/`shutdown`/`unknown` never archive at all (they hit the handler's early-return guard). Those strings are centralized into named constants at the mapping site with a comment, and producer-mapping tests assert the mapping for both lifecycle events so a host-side rename cannot silently change which archives are skipped.

## Tests

- `test/doctor-archive-cause-skip.test.ts`: a `manual-reset` archive appears in neither the safe nor needs-review scan and is left unmerged after apply; a `rollover-fallback` split still merges; a legacy NULL row still merges; a mixed group merges only the rollover row.
- `test/archive-cause-write.test.ts`: the real `/reset`, `session_end`, and rollover and cron paths persist the expected cause.
- `test/migration.test.ts`: the column is added idempotently to an existing DB; legacy rows read NULL; a second migration run is a no-op.

No public type or API changes. The full suite passes (1468 tests) and the typecheck is clean.

### DIFF
--- a/.changeset/doctor-skip-deliberate-reset.md
+++ b/.changeset/doctor-skip-deliberate-reset.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Stop the rollover-split doctor from auto-restoring conversations a user deliberately wiped with `/reset`. A new nullable `archive_cause` column records why each conversation was archived, written at the single archive funnel for both the `before_reset` and `session_end` lifecycle events a `/reset` surfaces. The doctor excludes deliberate causes (`manual-reset`) from its merge sources, so a `/reset` archive is never re-merged into the active conversation. Incidental archives (`rollover-fallback`, `cron-rotation`, `session-end`, idle/daily/compaction) and legacy NULL-cause rows stay merge-eligible, so genuine crash and rollover splits are still recovered.
--- a/src/db/migration.ts
+++ b/src/db/migration.ts
@@ -1067,6 +1067,7 @@ export function runLcmMigrations(
       session_key TEXT,
       active INTEGER NOT NULL DEFAULT 1,
       archived_at TEXT,
+      archive_cause TEXT,
       title TEXT,
       bootstrapped_at TEXT,
       created_at TEXT NOT NULL DEFAULT (datetime('now')),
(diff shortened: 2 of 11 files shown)


jalehman:

PR #958, opened 2026-07-02
fix(doctor-contract): honor explicit provider in fallbackProviders model refs

### LINKED ISSUE #853: doctor-contract toModelRef ignores the explicit fallbackProviders provider when model contains a slash, causing a false "Missing allowedModels entries" override-policy warning
### Summary

On startup, lossless-claw 0.12.0 emits a runtime LLM model override policy warning claiming a configured fallback summary model is missing from `plugins.entries.lossless-claw.llm.allowedModels`, and recommends `openclaw doctor --fix`. It is a false positive. The fallback model is already allowlisted in the exact form the runtime requests at call time. The divergence is in the doctor/contract helper `toModelRef`: for a `fallbackProviders` entry that supplies `provider` and `model` as separate fields, where `model` is itself a provider-namespaced id (for example an OpenRouter id such as `openai/gpt-4.1-mini`), `toModelRef` ignores the explicit `provider` field and instead treats the substring before the first slash in `model` as the provider. The runtime composes the same ref as `${provider}/${model}`, so the two paths disagree.

Unlike #701, the runtime allowlist check here actually passes (the configured ref matches what the runtime requests), so there is no runtime denial. Only the doctor/contract validation is wrong.

### Versions

- lossless-claw 0.12.0
- OpenClaw 2026.6.1
- Node 24
- Summary route via OpenRouter

### Configuration that triggers it

```json
{
  "plugins": { "entries": { "lossless-claw": {
    "llm": {
      "allowModelOverride": true,
      "allowedModels": [
        "openrouter/deepseek/deepseek-v4-flash:exacto",
        "openrouter/openai/gpt-4.1-mini"
      ]
    },
    "config": {
      "summaryModel": "openrouter/deepseek/deepseek-v4-flash:exacto",
      "fallbackProviders": [
        { "provider": "openrouter", "model": "openai/gpt-4.1-mini" }
      ]
    }
  } } }
}
```

### Observed

Startup advisory:

```
[lcm] Runtime LLM model override policy may block configured Lossless summary models. Run "openclaw doctor --fix" to repair plugins.entries.lossless-claw.llm. Required config path: plugins.entries.lossless-claw.llm.allowModelOverride and plugins.entries.lossless-claw.llm.allowedModels. Missing allowedModels entries: fallbackProviders=openai/gpt-4.1-mini.
```

In the same startup, the runtime logs the fallback with the full provider-qualified ref:

```
[lcm] Fallback providers: openrouter/openai/gpt-4.1-mini
```

No `model override is not allowlisted` error ever follows. Only the advisory fires.

### Root cause

`doctor-contract-api.js`, `toModelRef(provider, model)`:

```js
const slash = modelId.indexOf("/");
if (slash > 0 && slash < modelId.length - 1) {
  const directProvider = modelId.slice(0, slash).trim();
  const directModel = modelId.slice(slash + 1).trim();
  return directProvider && directModel ? `${directProvider}/${directModel}` : undefined;
}
const providerId = readString(provider);
return providerId ? `${providerId}/${modelId}` : undefined;
```

For `{ provider: "openrouter", model: "openai/gpt-4.1-mini" }`, `model` contains a slash, so the first branch runs: it derives `directProvider = "openai"`, returns `openai/gpt-4.1-mini`, and never consults the explicit `provider: "openrouter"`. The contract then reports `openai/gpt-4.1-mini` as the required allowlist entry, which is absent (the configured allowlist uses the 3-part `openrouter/openai/gpt-4.1-mini`).

This only affects `fallbackProviders` entries, where `provider` and `model` are separate fields and `model` is itself a provider/model id. A single-string `summaryModel` such as `openrouter/deepseek/deepseek-v4-flash:exacto` is unaffected, because there the leading segment is correctly kept as the provider.

By contrast, the runtime builds the fallback override ref as `${provider}/${model}` (yielding `openrouter/openai/gpt-4.1-mini`), which matches the configured allowlist, so the enforced check passes. Only the doctor/contract validation path mis-derives the ref.

### Impact

1. A persistent false-positive override-policy warning on every startup for any OpenRouter-namespaced fallback model.
2. `openclaw doctor --fix` would inject the mis-derived 2-part ref `openai/gpt-4.1-mini` into `llm.allowedModels`. That entry does not match the 3-part ref the runtime requests, so the allowlist gains a no-op entry while the original (correct) entry is the one actually used. The recommended remediation does not address the real (non-existent) problem and adds a misleading entry.

### Expected

`toModelRef` should honor an explicitly provided `provider` for `fallbackProviders` entries and compose the ref the same way the runtime does (`${provider}/${model}`). When the configured `allowedModels` already contains that ref, no warning should fire, and `doctor --fix` should suggest the same provider-qualified ref the runtime enforces.

### Suggested fix

In `toModelRef` (or wherever the contract derives `fallbackProviders` refs), prefer the explicit `provider` field when present instead of re-parsing `model` for a provider segment, or reuse the same ref builder the runtime already uses for `fallbackProviders`, so the validation path and the runtime path cannot diverge.

### Related

- #738 (closed): bare `LCM_SUMMARY_MODEL` ignoring configured `summaryProvider`. Same family of provider-aware model-ref resolution, but the opposite trigger (a bare model with no provider, rather than an explicit provider being dropped).
- #701 (closed): `doctor --fix` missing the `subagent` policy for `expansionModel`. Same doctor/override-policy area, but a coverage gap plus a real runtime denial, whereas here the runtime check passes and only the validator is wrong.

mpz4life:
When a `fallbackProviders` entry supplied an explicit `provider` (e.g. `openrouter`) and a `model` containing a slash (e.g. `anthropic/claude-sonnet-4-6`), `toModelRef()` ignored the explicit provider and parsed the provider from the model string. This produced a modelRef that did not match the configured override policy, causing false "Missing allowedModels entries" warnings from `openclaw doctor`.

Now `toModelRef()` prefers the explicit `provider` argument when present and only falls back to parsing provider/model from a slashed model string when no provider is given.

Fixes #853.

### DIFF
--- a/.changeset/blue-signs-swim.md
+++ b/.changeset/blue-signs-swim.md
@@ -1,0 +1,5 @@
+---
+"lossless-claw": patch
+---
+
+Fix `doctor-contract` model reference generation to honor the explicit `provider` field in `fallbackProviders` when the model value also contains a slash, preventing false "Missing allowedModels entries" override-policy warnings.
--- a/doctor-contract-api.js
+++ b/doctor-contract-api.js
@@ -44,14 +44,10 @@ function readSubagentPolicy(cfg) {
   return readModelOverridePolicy(cfg, "subagent");
 }
 
-function toModelRef(provider, model) {
+function toProviderPrefixedModelRef(provider, model) {
   const modelId = readString(model);
   if (!modelId) {
     return undefined;
-  }
-  const providerId = readString(provider);
-  if (providerId) {
-    return `${providerId}/${modelId}`;
   }
   const slash = modelId.indexOf("/");
   if (slash > 0 && slash < modelId.length - 1) {
@@ -59,7 +55,14 @@ function toModelRef(provider, model) {
     const directModel = modelId.slice(slash + 1).trim();
     return directProvider && directModel ? `${directProvider}/${directModel}` : undefined;
   }
-  return undefined;
+  const providerId = readString(provider);
+  return providerId ? `${providerId}/${modelId}` : undefined;
+}
+
+function toFallbackModelRef(provider, model) {
+  const providerId = readString(provider);
+  const modelId = readString(model);
+  return providerId && modelId ? `${providerId}/${modelId}` : undefined;
 }
 
 function uniqueModelRefs(modelRefs) {
@@ -88,7 +91,7 @@ function collectLosslessRuntimeLlmModelRefs(cfg) {
     if (!modelId) {
       return;
     }
-    const modelRef = toModelRef(provider, modelId);
+    const modelRef = toProviderPrefixedModelRef(provider, modelId);
     if (modelRef) {
       modelRefs.push({ field, modelRef, configPath });
       return;
@@ -124,7 +127,7 @@ function collectLosslessRuntimeLlmModelRefs(cfg) {
         });
         continue;
       }
-      const modelRef = toModelRef(fallback.provider, fallback.model);
+      const modelRef = toFallbackModelRef(fallback.provider, fallback.model);
       if (modelRef) {
         modelRefs.push({
           field: "fallbackProviders",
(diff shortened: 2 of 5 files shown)


jalehman:

PR #959, opened 2026-07-02
fix: delegate compact() to runtime for sessions excluded via ignoreSessionPatterns

spiral-cmd:
# PR draft (not yet opened — for review)

**Title:** fix: delegate compact() to runtime for sessions excluded via ignoreSessionPatterns

**Body:**

## What

When a session matches `ignoreSessionPatterns`, `compact()`'s exclusion branch currently returns a bare refusal:

```js
return { ok: true, compacted: false, reason: "session excluded" };
```

This PR changes that to delegate to OpenClaw's own runtime compaction bridge instead:

```js
return await delegateCompactionToRuntime(params);
```

## Why

`delegateCompactionToRuntime` (`openclaw/plugin-sdk`) is documented for exactly this case:

> "Delegate a context-engine compaction request to OpenClaw's built-in runtime compaction path. This is the same bridge used by the legacy context engine. Third-party engines can call it from their own `compact()` implementations when they do not own the compaction algorithm but still need `/compact` and overflow recovery to use the stock runtime behavior."

It's the entire implementation of OpenClaw's own default (no-plugin) engine — one line, `return await delegateCompactionToRuntime(params)`.

Without this, OpenClaw core's preflight guard (which checks raw transcript byte size independently of token usage, via `agents.defaults.compaction.maxActiveTranscriptBytes`) hard-fails the turn whenever an excluded session's transcript crosses that threshold, because its reason classifier only treats three specific strings (`below_threshold`, `no_compactable_entries`, `already_compacted_recently`) as safe to skip. `"session excluded"` isn't one of them, so it surfaces to the end user as "Context is too large and auto-compaction could not recover this turn" — even when token usage is a small fraction of the model's actual context window. The manual `/compact` command hits the same result but is handled more leniently on that side (labeled "skipped" rather than erroring), so it doesn't crash, but it also doesn't do anything.

This is a real, reproducible issue for any user who excludes Telegram (or any other) sessions via `ignoreSessionPatterns` for an active, long-running conversation.

## Scope

Limited to the `compact()` method's exclusion branch only. `bootstrap()`, `ingest()`/`ingestBatch()`, and `assemble()` already handle the excluded case correctly:
- `bootstrap()`/`ingest()`: correctly declining to track excluded sessions, no framework bridge exists or is needed there.
- `assemble()`: already a correct pass-through (`{ messages: params.messages, estimatedTokens: 0 }`).

Only `compact()`'s refusal has an externally-visible side effect (blocking overflow recovery).

## Known limitation

Sessions compacted this way get a flat runtime summary instead of the plugin's DAG-based summarization — no `lcm_grep`/`lcm_expand` recall afterward. This is consistent with what "excluded from LCM" already implies, and is strictly better than the current behavior (hard failure / unbounded transcript growth).

## Operational note for anyone using `ignoreSessionPatterns`

If a session is later removed from the exclusion list after having been compacted this way, its transcript will contain a runtime-generated summary block that the plugin's own ingestion has never produced. Depending on how `bootstrap()`/ingestion handles unexpected content at the start of a transcript, this could need a one-time reset (fresh session, or clearing that session's rows) rather than transparent re-adoption. Flagging this as something worth documenting or hardening against, not something this PR attempts to solve.

## Testing

Verified against a real, live, oversized excluded session (Telegram group session, ~7.5MB transcript, `ignoreSessionPatterns`-excluded): before this change, `/compact` returned `Compaction skipped: session excluded` and did nothing; after, it returned a real compacted result (`~190k → ~53k` tokens, transcript rotated with a genuine LLM-generated summary present in the new file), and the preflight hard-fail no longer occurs on subsequent turns.

## Note on the import

This is the first direct runtime import from `openclaw/plugin-sdk` in `engine.ts` — everywhere else in this file goes through the `LcmDependencies` injection pattern (`this.deps.*`) instead of importing directly from `openclaw`. `openclaw` isn't a real devDependency in this repo, so `npm run typecheck` can't resolve `openclaw/plugin-sdk` locally (`Cannot find module`), even though it resolves correctly at runtime inside an actual OpenClaw install (verified — that's how the fix was tested). If the preferred convention here is to route this through `LcmDependencies` instead (e.g. injecting `delegateCompactionToRuntime` alongside `complete`/`callGateway`/etc.), happy to rework it that way — didn't want to guess at the wiring in `src/plugin/index.ts` without more context on the intended pattern.

### DIFF
--- a/.changeset/ignored-session-runtime-compact.md
+++ b/.changeset/ignored-session-runtime-compact.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Delegate ignored-session compaction to OpenClaw's runtime compaction path when the host exposes it, so sessions excluded from LCM can still recover from raw transcript pressure.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -356,7 +356,7 @@ LCM_EXPANSION_MODEL=openai/gpt-5.4-mini
 - `*` matches any characters except `:`
 - `**` matches anything, including `:`
 
-Cron scheduler keys (`agent:<agent>:cron:<job>...`) are isolated automatically when a new runtime `sessionId` reuses the same `sessionKey`. Configure `ignoreSessionPatterns` for cron only when the run should bypass LCM entirely; leave cron sessions included when they need in-run compaction.
+Cron scheduler keys (`agent:<agent>:cron:<job>...`) are isolated automatically when a new runtime `sessionId` reuses the same `sessionKey`. Configure `ignoreSessionPatterns` for cron only when the run should bypass LCM entirely; leave cron sessions included when they need in-run compaction. When OpenClaw exposes its runtime compaction delegate, `/compact` and overflow recovery for ignored sessions fall back to OpenClaw's built-in compaction path instead of LCM's summary DAG. Older hosts that do not expose that delegate keep the previous safe skip behavior.
 
 Example:
 
--- a/skills/lossless-claw/references/config.md
+++ b/skills/lossless-claw/references/config.md
@@ -548,6 +548,7 @@ Why it matters:
 - keeps low-value automation or noisy sessions out of the DB
 - useful for excluding certain agent lanes or ephemeral traffic entirely
 - cron scheduler keys are already isolated per runtime run, so ignore them only when they should bypass LCM compaction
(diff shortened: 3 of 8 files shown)


jalehman:
Thanks for the contribution, @spiral-cmd. This is landed now.

PR #960, opened 2026-07-02
fix(assemble): demote happy-path fork-suffix append log from warn

gorkem2020:
## Summary

The per-assemble `[lcm] assemble: appended fork-bounded live suffix ...` line logs at warn unconditionally whenever the append attaches messages. For thread-fork sessions (for example Slack threads, where every thread is a fork of the base conversation) this fires on essentially every assemble, so routine healthy operation produces a steady stream of warns. On a live multi-agent gateway, one afternoon of threaded chatting produced dozens of these, every one of them reporting `evictedMessages=0 evictedTokens=0 overBudget=false`.

This change logs the healthy append at debug and keeps warn for the states that actually need operator attention: the append evicted messages or ran over budget. The level choice lives in a small exported helper (`forkBoundedLiveSuffixAppendLogLevel`) next to `appendForkBoundedLiveSuffixWithinBudget`, and the engine call site uses the existing indexed-log idiom. Assembly behavior is unchanged throughout; only the log level moves.

Same spirit as [#954](https://github.com/Martian-Engineering/lossless-claw/pull/954): happy-path recovery/maintenance chatter should not sit at warn.

## Tests

Adds `test/fork-suffix-loglevel.test.ts` covering the three cases (healthy append at debug; warn retained on eviction; warn retained on over-budget). Full suite green: 1485/1485 (`npx vitest run --dir test`), `tsc --noEmit` clean. Includes a changeset.

### DIFF
--- a/.changeset/demote-fork-suffix-append-log.md
+++ b/.changeset/demote-fork-suffix-append-log.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Demote the per-assemble "appended fork-bounded live suffix" log from warn to debug on the healthy path. Thread-fork sessions append this suffix on essentially every assemble, so the line was warn-level noise in routine operation. The log stays at warn when the append evicted messages or ran over budget, the states that need operator attention. Assembly behavior is unchanged; only the log level moves.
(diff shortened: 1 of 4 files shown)


jalehman:

Issue #965, opened 2026-07-04
Bug: same-turn collapse no longer recognizes untrusted-metadata-decorated copies, so every inbound turn reaches the model twice


gorkem2020:
### Summary

Since the "require genuine decoration before collapsing same-turn duplicates" hardening (shipped in [c70e4df](https://github.com/Martian-Engineering/lossless-claw/commit/c70e4df)), the same-turn collapse gate no longer accepts an OpenClaw untrusted-metadata block as decoration evidence. On hosts where inbound messages arrive wrapped in that standard metadata block without a leading channel timestamp (the default OpenClaw shape), the decorated runtime copy of each inbound user turn never collapses onto its bare persisted row. The result, verified on a live deployment via request-level logging: the model-facing prompt contains exactly persisted_rows + 1 byte-identical consecutive copies of every real inbound user turn, on every call, and the batch path can additionally double-write the store. On a busy multi-agent channel lane this compounded to a 139k-token prompt where roughly half the history was duplicate copies.

### Root cause

The gate `liveContentIsRecognizedDecoratedBareBody` ([src/live-coverage.ts](https://github.com/Martian-Engineering/lossless-claw/blob/main/src/live-coverage.ts)) and the frontier matcher in [src/batch-dedup.ts](https://github.com/Martian-Engineering/lossless-claw/blob/main/src/batch-dedup.ts) both dropped the `contentBeginsWithOpenClawInboundMetadataBlock(...)` arm, on the stated rationale that a structured metadata frame is user-forgeable and must not count as trusted decoration. That reasoning is sound for the containment case it defends against, but it also removed the only recognition path for the most common legitimate shape. With the gate returning false, the collapse never fires, and the coverage signature comparison cannot help because it keys on raw content, where the decorated and bare copies differ.

### Repro

1. Run lossless-claw under OpenClaw with a channel whose inbound messages carry the standard untrusted-metadata block and no leading channel timestamp.
2. Exchange a few turns, then inspect the assembled request payload with any request logger.
3. Every real inbound user turn appears twice (or stored_rows + 1 times when store-level duplicates exist for other reasons), byte-identical and consecutive.

### Fix

[...]

PR #967, opened 2026-07-04
fix(dedup): collapse a metadata-wrapped runtime copy onto its bare persisted row by model-facing body equality

gorkem2020:
### What this fixes

On hosts whose inbound messages arrive wrapped in the standard OpenClaw untrusted-metadata block without a leading channel timestamp, the same-turn collapse in the afterTurn batch path no longer recognizes the wrapped runtime copy as a face of the bare persisted row, so the copy is written as a second row and replayed to the model on every subsequent call. Live measurement on a busy multi-agent channel: every real inbound user turn reached the model twice (persisted rows + 1), compounding to a 139k-token prompt where roughly half the history was duplicates. Root cause analysis and reproduction details are in the companion issue: https://github.com/Martian-Engineering/lossless-claw/issues/965.

### Why the gate misses

The "require genuine decoration before collapsing same-turn duplicates" hardening removed `contentBeginsWithOpenClawInboundMetadataBlock(...)` from the collapse gate on the sound reasoning that a structured metadata frame is user-forgeable text. That removal also deleted the only recognition path for the most common legitimate shape, and the coverage signature cannot help because it keys on raw content, where the wrapped and bare faces differ.

### The fix

A new shared helper (`openClawInboundBodiesMatch` in `openclaw-inbound-metadata.ts`) strips a structurally validated leading metadata block plus any leading channel timestamp from both candidates and compares the remaining full bodies byte for byte. The batch frontier matcher (`runtimeRowCoversPersistedFrontierRow`) collapses on a match.

This consciously revisits the keep-both behavior the hardening pinned, and the reversal is stated in the commit body: full-body equality after a structural strip drops no user content (only the non-anchoring wrapper), while the defended forgery (a frame concealing a different turn) necessarily fails the equality and keeps the existing fail-closed path. The comparison also only runs inside the same-turn frontier window, so a cross-turn forged duplicate never reaches it.

Scoping note: the collapse is deliberately NOT routed through the shared assembly recognizer (`liveContentIsRecognizedDecoratedBareBody`). That recognizer serves the assembly path with the opposite polarity (identifying a decorated current turn to append), and unifying the two verifiably regressed the deliberate bare-plus-decorated current-turn behavior pinned by the structural-supersede tests. The batch double-write is the source of the echo, and the batch matcher is where the fix lives; the strip-and-compare logic stays in one helper so the two sites cannot drift.

### Tests

Three new tests: the wrapped-runtime versus bare-persisted pair collapses to one row; a distinct turn whose body quotes a metadata-looking frame with a different body is NOT collapsed; two genuinely repeated identical user turns as separate frontier turns are both retained. Two existing tests that pinned the keep-both behavior for this exact shape are flipped to expect collapse, each with a justification comment. Full suite green.

### DIFF
--- a/.changeset/same-turn-collapse-frontier.md
+++ b/.changeset/same-turn-collapse-frontier.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Collapse metadata-wrapped OpenClaw runtime copies onto their bare persisted rows only when the covered transcript frontier proves same-turn alignment. Degraded after-turn dedup remains timestamp-gated, so repeated short user messages wrapped in forgeable metadata-shaped text are preserved instead of silently collapsed.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -81,6 +81,7 @@ export class BatchDeduplicator {
     persistedContent: string,
     batchRole: string,
     batchContent: string,
+    options?: { allowUntimestampedInboundBodyMatch?: boolean },
   ): boolean {
     if (
       messageIdentity(persistedRole, persistedContent) ===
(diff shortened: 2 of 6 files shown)


gorkem2020:
A review thread on [#974](https://github.com/Martian-Engineering/lossless-claw/pull/974) surfaced a default-choice question that belongs to this PR's mechanism rather than to the recap widening: when the degraded (non-frontier) dedup path meets two rows with byte-equal full bodies and no timestamp evidence, should it collapse by default (current behavior of this PR, pinned by the two keep-to-collapse test flips) or stay keep-both, restricting the no-timestamp `openClawInboundBodiesMatch` collapse to the covered-frontier alignment path where positional evidence is strongest? Details and a sketched ~15-line guard in the thread: [#974 discussion](https://github.com/Martian-Engineering/lossless-claw/pull/974#discussion_r3543783201). Since the bytes alone cannot distinguish a store double-write from two genuine repeats, this is a policy default; the repo's lossless standard arguably favors keep-both on the unproven path. Happy to implement either resolution here.

jalehman:

Issue #966, opened 2026-07-04
Bug: reconcile ingestion accepts re-delivered inbound events under fresh transcript entry ids, storing 2 to 5 duplicate rows per event


gorkem2020:
### Summary

When the host re-delivers the same logical inbound event under fresh transcript entry uuids (observed during provider-stall retry and failover storms: idle-watchdog retries plus rate-limit retries re-append the event to the transcript), the reconcile ingestion path stores one row per delivery. A live deployment accumulated clusters of 2 to 5 rows with byte-identical content, identical identity_hash, identical to-the-second created_at (the frozen source timestamp), but a distinct transcript_entry_id per row, at non-consecutive seq values (other traffic landed between the passes). The duplicated rows then replay into every assembled prompt for the rest of the conversation.

### Root cause

Every dedup layer on the ingest path keys on transcript entry id identity and treats a proven-absent id as proven-new content:

- The exact-idempotency check in `ingestSingle` ([src/engine.ts](https://github.com/Martian-Engineering/lossless-claw/blob/main/src/engine.ts)) only rejects entry ids that are already persisted; a re-delivery carries a fresh id.
- `adoptTranscriptEntryId` ([src/store/conversation-store.ts](https://github.com/Martian-Engineering/lossless-claw/blob/main/src/store/conversation-store.ts)) collapses onto existing rows only where transcript_entry_id is NULL, so once copy 1 stamps the row, copies 2..n insert.
- The entry-id reconcile path passes `skipReplayTimestampFloodGuard: true` ("entry-id-verified imports are exact"), bypassing the same-second replay flood guard that protects the live path.
- The content-based no-anchor overlap guard is explicitly disabled when all entries bear ids ([src/transcript-reconciler.ts](https://github.com/Martian-Engineering/lossless-claw/blob/main/src/transcript-reconciler.ts)).
- The identity_hash index is non-unique, and the schema uniques (conversation_id+seq, conversation_id+transcript_entry_id) cannot catch cross-id replays.

### The discriminating signal (verified on live data)

Each re-appended transcript entry carries a fresh envelope timestamp, but the inner message timestamp (the millisecond-epoch source-event stamp the store later truncates into created_at) is frozen and identical across every copy of the same logical event. Genuinely repeated messages are distinct source events and carry distinct millisecond stamps. So replay detection does not need content heuristics or a schema change: the full-precision inner timestamp plus content identity identifies a re-append exactly.

### Repro (deterministic, no concurrency needed)

[...]

PR #968, opened 2026-07-04
fix(reconcile): guard entry-id imports against retry re-appends via source-identity twin detection

gorkem2020:
### What this fixes

When the host's retry or failover machinery re-appends the same logical inbound event to the transcript under a fresh entry uuid (observed during a provider-stall storm: idle-watchdog retries plus rate-limit retries), the entry-id reconcile path imports one row per re-append. A live deployment accumulated clusters of 2 to 5 rows with byte-identical content, identical identity_hash, identical second-truncated created_at, distinct transcript_entry_id per row, at non-consecutive seq (other traffic landed between passes). Details and the full guard-by-guard analysis are in the companion issue: https://github.com/Martian-Engineering/lossless-claw/issues/966.

### The discriminating signal

Each re-appended copy carries a fresh envelope timestamp but an identical inner millisecond source timestamp (the value later truncated into created_at). Genuine repeats are distinct source events with distinct millisecond stamps. Verified on live data.

### How

Twin detection on the entry-id reconcile path: a candidate entry with a proven-absent id is skipped when an already-persisted entry exists with a different id, the same role, the same canonical content, and an identical inner millisecond timestamp. Full-tail routes (bootstrap and full re-read afterTurn) check the parsed tail directly. The append-only fast path stays fast via a two-tier gate: an indexed store lookup on (conversation_id, role, identity_hash, created_at second) filters normal traffic for free, and only a hit triggers a bounded full leaf-path re-read to compare full millisecond precision. Fail-open when the inner timestamp is missing on either side. No schema migration.

### Tests

New deterministic twin tests: a re-appended twin is skipped (exactly one row persists); an identical message with a 1ms-different source timestamp is imported (both rows kept); a candidate without an inner timestamp imports unchanged (fail-open). The existing keep-repeats invariants (repeated id-less prefix messages, byte-identical tool-loop pairs under fresh ids, repeated content under a fresh id while the original is live) all remain green, since those fixtures carry distinct millisecond stamps. Full suite green.

### DIFF
--- a/.changeset/replay-twin-entry-id-guard.md
+++ b/.changeset/replay-twin-entry-id-guard.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Skip replayed transcript entries that OpenClaw re-appends under fresh entry ids when the persisted row and new entry share the same canonical identity and full-precision inner source timestamp.
--- a/src/store/conversation-store.ts
+++ b/src/store/conversation-store.ts
@@ -873,6 +873,38 @@ export class ConversationStore {
   }
 
   /**
+   * Tier-1 replay-twin gate for the append-only reconcile path: does this
+   * conversation already hold a row with the same role, identity hash, and
+   * created_at SECOND as the candidate? An indexed point lookup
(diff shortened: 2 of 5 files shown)


jalehman:

PR #970, opened 2026-07-05
fix(doctor): back up database before scoped summary-repair apply

gorkem2020:
## Problem

The scoped summary-repair `doctor apply` path, [`applyScopedDoctorRepair`](https://github.com/gorkem2020/lossless-claw/blob/fix/summary-repair-apply-backup/src/plugin/lcm-doctor-apply.ts#L50-L58), mutates `lcm.db` inside a `withDatabaseTransaction` call with no database backup taken first. The sibling rollover-split repair path, [`applyRolloverSplitRepair`](https://github.com/gorkem2020/lossless-claw/blob/fix/summary-repair-apply-backup/src/plugin/lcm-doctor-rollover-splits.ts#L680-L718), always takes an unconditional backup via [`createLcmDatabaseBackup`](https://github.com/gorkem2020/lossless-claw/blob/fix/summary-repair-apply-backup/src/plugin/lcm-doctor-rollover-splits.ts#L704-L712) before it mutates anything. The two `doctor apply` subcommands ended up with different safety guarantees even though both perform the kind of automatic, in-place repair that [AGENTS.md](https://github.com/Martian-Engineering/lossless-claw/blob/main/AGENTS.md) asks maintenance/repair code to treat carefully ("lossless means lossless").

## Fix

Route `applyScopedDoctorRepair` through the same `createLcmDatabaseBackup` call, in the same spot rollover-splits uses it (right before the mutating transaction, only once at least one summary is queued for rewrite) and with the same failure handling: return `{ kind: "unavailable", reason: ... }` if a backup path can't be resolved, instead of proceeding to mutate. See [lcm-doctor-apply.ts#L147-L157](https://github.com/gorkem2020/lossless-claw/blob/fix/summary-repair-apply-backup/src/plugin/lcm-doctor-apply.ts#L147-L157).

The no-op case is unchanged: when nothing needs repairing (`repairedSummaryIds.length === 0`), the function still returns `applied` with zero counts and never touches the backup helper, the same way rollover-splits skips its own backup when there are no safe lanes to merge.

## Test

Added a focused test asserting a backup file is created before doctor apply mutates any summary: [engine-maintenance.test.ts#L1639-L1690](https://github.com/gorkem2020/lossless-claw/blob/fix/summary-repair-apply-backup/test/engine-maintenance.test.ts#L1639-L1690).

Red/green proof, the test was written and run against the unmodified source first, then again after the fix:

```
# before the fix (source unmodified)
$ npx vitest run test/engine-maintenance.test.ts -t "creates a database backup before mutating summaries"
✗ creates a database backup before mutating summaries
  AssertionError: expected 0 to be greater than 0

# after the fix
$ npx vitest run test/engine-maintenance.test.ts -t "creates a database backup before mutating summaries"
✓ creates a database backup before mutating summaries
```

No regressions in the surrounding suite:

```
$ npx vitest run test/engine-maintenance.test.ts test/engine-fidelity.test.ts test/engine-after-turn.test.ts test/doctor-archive-cause-skip.test.ts
Test Files  4 passed (4)
     Tests  146 passed (146)

$ npx vitest run test/lcm-command.test.ts
Test Files  1 passed (1)
     Tests  63 passed (63)
```

`npm run typecheck` and `npm run build` both pass clean.

## Release note

This changes user-facing behavior: an automatic repair path now backs up the database first, and can newly report `unavailable` if a backup path can't be resolved (for example, a non-file-backed/in-memory database). Per [RELEASING.md](https://github.com/Martian-Engineering/lossless-claw/blob/main/RELEASING.md), that should get a changeset before merge. I have not added one myself since this repo's stated convention is not to expect external contributors to run the Changesets workflow; flagging it here for a maintainer to add (`patch`, matching "fixes, docs-visible behavior changes" in the bump guidance).

### DIFF
--- a/.changeset/scoped-doctor-repair-backup.md
+++ b/.changeset/scoped-doctor-repair-backup.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Create and report a SQLite backup before scoped doctor summary repair writes, matching the backup-first safety behavior used by rollover-split repair.
--- a/src/plugin/lcm-command.ts
+++ b/src/plugin/lcm-command.ts
@@ -2920,6 +2920,7 @@ async function buildDoctorApplyText(params: {
       buildStatLine("truncated-marker summaries", formatNumber(stats.truncated)),
       buildStatLine("fallback-marker summaries", formatNumber(stats.fallback)),
       buildStatLine("emergency-fallback summaries", formatNumber(stats.emergency)),
+      buildStatLine("backup path", result.backupPath ?? "skipped (no writes)"),
       buildStatLine("repaired summaries", formatNumber(result.repaired)),
       buildStatLine("unchanged summaries", formatNumber(result.unchanged)),
       buildStatLine("skipped summaries", formatNumber(result.skipped.length)),
--- a/src/plugin/lcm-doctor-apply.ts
+++ b/src/plugin/lcm-doctor-apply.ts
@@ -1,5 +1,5 @@
 import type { DatabaseSync } from "node:sqlite";
-import { withDatabaseTransaction } from "../transaction-mutex.js";
+import { DatabaseTransactionTimeoutError, withExclusiveDatabaseLock } from "../transaction-mutex.js";
 import { formatTimestamp } from "../compaction.js";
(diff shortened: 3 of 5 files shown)


jalehman:

PR #979, opened 2026-07-10
fix(tools): lcm_describe accepts full [LCM Tool Output: ...] reference strings

### LINKED ISSUE #895: lcm_describe should accept full LCM Tool Output reference strings
## Problem

`lcm_describe` supports drilldown for `file_xxx` IDs, and the tool description tells agents to use it when they see an `[LCM Tool Output: file_xxx | ...]` reference.

However, the natural thing for an agent to pass may be the whole visible reference string, for example:

```text
[LCM Tool Output: file_abc123 | tool=Bash | 42,000 bytes]
```

Rather than extracting only:

```text
file_abc123
```

The retrieval path appears to route by ID prefix (`file_`, `sum_`, etc.), so passing the full bracketed reference is likely to miss unless every agent remembers to parse the ID first.

## Expected behavior

`lcm_describe` should accept full LCM reference strings and extract the embedded `file_xxx` ID when unambiguous.

## Acceptance criteria

- `lcm_describe(id="file_xxx", expandFile=true)` continues to work.
- `lcm_describe(id="[LCM Tool Output: file_xxx | tool=... | N bytes]", expandFile=true)` resolves the same file.
- The parser is conservative: if zero or multiple `file_xxx` IDs are present, return a clear validation error instead of guessing.
- Tests cover `[LCM Tool Output: ...]`, `[LCM File: ...]` if applicable, and malformed/multiple-ID strings.
- Tool help mentions that either a raw `file_xxx` ID or a copied LCM reference string is accepted.

cxbAsDev:
Closes #895.

 now extracts a bare `file_xxx` or `sum_xxx` ID from a full reference string such as `[LCM Tool Output: file_xxx | tool=... | N bytes]`, so agents can paste the reference directly instead of manually isolating the ID.

### DIFF
--- a/.changeset/lcm-describe-reference-string.md
+++ b/.changeset/lcm-describe-reference-string.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+`lcm_describe` now accepts a full `[LCM Tool Output: file_xxx | ...]` reference string as `id`, extracting the embedded `file_xxx` or `sum_xxx` ID automatically. Bare IDs continue to work, and unrecognized inputs return a clearer error hint.
--- a/.changeset/lcm-describe-reference-string.md
+++ b/.changeset/lcm-describe-reference-string.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-`lcm_describe` now accepts a full `[LCM Tool Output: file_xxx | ...]` reference string as `id`, extracting the embedded `file_xxx` or `sum_xxx` ID automatically. Bare IDs continue to work, and unrecognized inputs return a clearer error hint.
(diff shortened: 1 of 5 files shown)


jalehman:
Maintainer triage note: this is a real and project-appropriate fix, but the parser is currently broader than #895 asks for.

The issue acceptance criteria call for conservative parsing. The current implementation extracts the first `sum_` / `file_` token anywhere in the argument, which can silently select the wrong record when input contains multiple IDs or a malformed partial ID.

Suggested narrow revision:

- Accept bare IDs and recognized copied reference forms.
- Reject ambiguous/malformed input with clear errors, especially zero IDs and multiple IDs.
- Add tests for multiple IDs, malformed/disallowed-character bare IDs, and `[LCM File: ...]` if that copied form is intended to be supported.

I would not close this PR; the direction is right, but it needs the conservative parser behavior before review/merge.

jalehman:

Issue #883, opened 2026-06-11
independentLogFile isn't a valid config value


DigitalCharon:
I'm trying to debug issues that occur when I try to use lossless-claw. So I tried to set a log file based on what I saw in the [config reference](https://github.com/Martian-Engineering/lossless-claw/blob/main/docs/configuration.md). 

```json
  "independentLogFile": {
    "enabled": true,
    "file": "/.openclaw/lcm-logs/log.log",
    "maxFileBytes": 104857600
  },
``` 

However, I get this error:
```shell
plugins.entries.lossless-claw.config: invalid config: must not have additional properties: "independentLogFile"
```

PR #980, opened 2026-07-10
fix(log): make independent log file redaction tests deterministic

cxbAsDev:
Relates to #883.

The independentLogFile config key is already declared in openclaw.plugin.json and accepted by src/db/config.ts. This change fixes test flakiness: when openclaw is installed in the test environment, createRequire loads the host redactor and the fallback-redaction tests see host-redacted output instead of the expected [REDACTED] markers. The tests now reset or inject the redactor explicitly so the fallback path is deterministic.


jalehman:

PR #981, opened 2026-07-10
feat(doctor): allow doctor apply to target a specific conversation id

### LINKED ISSUE #917: doctor apply cannot safely repair a non-current conversation
## Summary
The native `/lossless doctor apply` / `/lcm doctor apply` command only repairs the **current conversation**. There is no user-facing way on the native command surface to target a different conversation ID, archived conversation, or prior conversation from another lane.

That would be fine if the native command were the only repair path. But the current UX is confusing because:

- the native command can hit safety preflights on large/live conversations
- operators may reasonably assume they should switch to another lane and run the same command there
- doing so repairs the *other lane's current conversation*, not the intended target
- meanwhile the repo also ships an external maintenance tool (`lcm-tui`) that **can** repair a named conversation ID offline

So the real gap is not “no repair path exists anywhere.” The real gap is that the **native repair workflow has no parity with the external maintenance tool and no clear targeting story for non-current conversations**.

### Native command surface
The native command help/code makes this current-conversation scope explicit:
- help text says: **"Repair broken summaries in the current conversation."**
- `doctor apply` parsing accepts only optional `confirm-offline`-style overrides
- there is no user-facing `conversationId` / `sessionKey` / `--conversation` targeting on the native command path

### External maintenance path
The repository also documents a separate tool:
- `lcm-tui repair <conversation_id> --apply`
- `lcm-tui doctor <conversation_id> --apply`

That means offline repair of a previous conversation **does** exist, but only through the separate `lcm-tui` maintenance binary, not through the normal `/lossless` chat-command workflow.

## Why this is still a problem
Today the behavior is split across two worlds:

1. **Native `/lossless` repair**
   - easy to reach from chat
   - scoped to current conversation only
   - no explicit target selector

2. **External `lcm-tui` repair**
   - can target a specific conversation ID
   - works for prior/non-current conversations
   - is not part of the normal native command workflow

That split is easy to misunderstand and creates bad operator ergonomics, especially when a user is trying to repair the conversation they were just told is too large/hot to handle inline.

## Expected behavior
At least one of these should exist:

1. **Parity in native commands**
   - `doctor apply` can target a conversation ID/session key from another lane
   - example shape: `doctor apply --conversation 472`

2. **Clear native handoff**
   - native `/lossless doctor` explicitly tells the operator that non-current/offline repair must be done with `lcm-tui repair <conversation_id> --apply`
   - ideally with exact conversation ID surfaced in the output

3. **Unified maintenance UX**
   - native repair command delegates or bridges to the same offline repair machinery used by `lcm-tui`

## Actual behavior
- native `doctor apply` repairs only the current conversation
- there is no native target flag for another conversation
- separate `lcm-tui` tooling can repair prior conversations by ID, but that capability is not available through the normal native command workflow

## Impact
This is primarily a **workflow / parity / discoverability** issue:
- operators can incorrectly conclude that old conversations are unrepairable
- operators can waste time trying to use `confirm-offline` from the wrong lane
- repair capability exists, but not where users of the native command surface expect it

## Classification
Best classification now:
- **feature gap / workflow bug** in the native repair UX
- specifically: missing native targeting parity and poor discoverability of the `lcm-tui` offline repair path

cxbAsDev:
Closes #917.

## What
Adds native conversation-id targeting for summary repair through `/lossless doctor apply <conversation-id> confirm-offline`.

## Why
The native command repairs only the active conversation, while offline tooling can repair a named conversation. Explicit targeting gives operators the same repair scope without risking a rewrite in the wrong active lane.

## Changes
- Parse one positive conversation id
- Require exact targeted offline confirmation
- Preserve large and hot preflight checks
- Keep current-conversation behavior unchanged
- Document authorization and lane isolation
- Add a minor release changeset

## Testing
- `npm test -- --run test/lcm-command.test.ts`
- `npm run typecheck`
- `npm run build`
- Synthetic current-main merge: 1,821 tests pass
- Hosted test, inspector, and smoke checks pass

### DIFF
--- a/.changeset/doctor-apply-conversation-id.md
+++ b/.changeset/doctor-apply-conversation-id.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": minor
+---
+
+`/lossless doctor apply` now accepts an optional conversation id: `doctor apply [<conversation-id>] [confirm-offline]`. This lets operators repair a specific conversation without needing that session to be the current active one. The existing current-conversation behavior is unchanged when no id is provided.
--- a/.changeset/doctor-apply-conversation-id.md
+++ b/.changeset/doctor-apply-conversation-id.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": minor
 ---
 
-`/lossless doctor apply` now accepts an optional conversation id: `doctor apply [<conversation-id>] [confirm-offline]`. This lets operators repair a specific conversation without needing that session to be the current active one. The existing current-conversation behavior is unchanged when no id is provided.
+`/lossless doctor apply` can now repair a specific conversation with `doctor apply <conversation-id> confirm-offline`. Targeted repair is limited to authorized OpenClaw command senders and requires the explicit offline confirmation after the target's active channel path is isolated. The existing current-conversation behavior is unchanged when no id is provided.
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ The plugin now ships a bundled `lossless-claw` skill plus a small plugin command
 - `/lossless backup` creates a timestamped backup of the current LCM SQLite database
 - `/lossless rotate` rewrites the active session transcript into a compact tail-preserving form without changing the live OpenClaw session identity or current LCM conversation
 - `/lossless doctor` scans for broken or truncated summaries
+- `/lossless doctor apply` repairs broken summaries in the current conversation after the normal safety preflight
+- `/lossless doctor apply <conversation-id> confirm-offline` repairs a specific conversation after its active channel path has been paused or moved away; targeted repair is restricted to authorized OpenClaw command senders and always requires the explicit offline confirmation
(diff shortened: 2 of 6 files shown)


jalehman:
Maintainer triage note: this is the right feature lane for #917, but there are a few targeted correctness fixes before deeper review.

Please preserve the selected conversation ID in the large/hot confirm-offline guidance. Right now the targeted safety-preflight path can tell the operator to run `doctor apply confirm-offline`, which drops the requested conversation ID and can send them back to the current conversation. The guidance should keep the target, e.g. `doctor apply <conversation-id> confirm-offline`.

Two smaller follow-ups from the same pass:

- Add regression coverage for nonexistent target IDs and duplicate numeric IDs.
- Rename the ID-targeted resolver provenance from `source: "session_id"` to a truthful source such as `conversation_id`.

I would not close this PR; it looks like a good fit once those narrow issues are fixed.

jalehman:

PR #991, opened 2026-07-11
fix(rollover): scope identity-overlap freshness to session generation and warn once

gorkem2020:
## Problem

When a session rolls over (typed /new or /reset) and the first message of the new session is byte-identical to a message already in persisted history (a repeated short probe like "ping" across rapid same-day rollovers), the freshness evaluation classifies the rollover as ambiguous: the identity-overlap gate treats any non-heartbeat content appearing exactly once as lineage-discriminating. Two consequences:

1. The conversation takes the preserve path instead of a clean rotation, replaying prior context (observed live as duplicate replies re-answering older messages).
2. The ambiguity re-warns on every subsequent turn, in both the bootstrap and afterTurn phases, with nothing to resolve it: two WARN lines per message, indefinitely.

## Fix

Two narrow changes in session-rollover.ts, both fail-closed:

1. **Deliberate-rollover awareness.** `findAmbiguousSessionKeyRuntimeRollover` now threads its existing deliberate-reset evidence (the `softResetPrunedAt` marker and the `.reset.` archive sibling) into the freshness evaluation. The identity-overlap gate bypasses a single-occurrence match only when that evidence is present AND the overlapping content is trivial (8 trimmed characters or fewer, mirroring the existing heartbeat-noise exclusion). Substantial-content overlap still blocks rotation even with reset evidence, preserving the foreign-reused-key defense covered by the existing sibling-rebind tests.
2. **Once-only warning.** The detector memoizes the last warned freshness reason per session generation (sessionKey + old session id + new session id), shared across bootstrap and afterTurn. Repeats of the same reason demote to debug; a changed reason warns once more.

## Tests

New suite `test/rollover-identity-scope.test.ts` (4 tests, 3 genuinely red against the pre-fix code via a tests-first commit):

- trivial identical content + deliberate-reset evidence rotates cleanly with zero warns
- same-generation ambiguity warns once across repeated bootstrap evaluations
- bootstrap and afterTurn share the memo (one warn total)
- no-evidence control: behavior unchanged (fail-closed guard; passes before and after)

Full suite 97 files / 1718 tests green; typecheck and build clean. Changeset included (patch) since warn-level behavior changes.


gorkem2020:
All three P3s addressed at head [`65e89f3`](https://github.com/Martian-Engineering/lossless-claw/pull/991/commits/65e89f3):

1. **Cross-phase once-only test completeness:** fixed. Added the matching notProvablyFreshWarns assertion alongside the existing preservingWarns one, so both once-only warn sites are proven once-only across bootstrap + afterTurn.
2. **8-char trivial floor:** took the lighter option, not a redesign. Two new boundary tests pin the threshold from both directions (8 chars bypasses, 9 chars fails closed), so any future change to the constant is a deliberate review point, plus a docstring addendum acknowledging the residual edges you called out. Substantial content still fails closed exactly as before.
3. **Unbounded warn memo:** capped at 500 entries with FIFO eviction, mirroring the existing AFTER_TURN_RECONCILE_KEY_CAP pattern in this codebase. An evicted generation just re-warns once more, which is harmless noise, so plain size-capped FIFO suffices. Dedicated test proves the cap and eviction order; changeset extended since eviction is a rare but observable behavior change.

Full suite 98 files / 1722 tests green on the new head.

jalehman:

Issue #928, opened 2026-06-21
Gracefully degrade for claude-cli / CLI-backed OpenClaw runtimes instead of hard-failing agent-run


Adam-Researchh:
## Summary

Lossless Claw `0.12.0+` currently hard-fails OpenClaw agent turns when the selected model is routed through the `claude-cli` CLI backend. This blocks the whole user turn instead of gracefully degrading to legacy/no-LCM behavior for that backend.

The practical result is that a mixed OpenClaw setup cannot safely upgrade Lossless Claw past `0.11.2` if it uses:

- OpenAI / Pi / embedded runtime models with Lossless Claw enabled
- Anthropic canonical model refs such as `anthropic/claude-opus-4-6`
- Anthropic execution routed through OpenClaw's hidden `claude-cli` runtime/auth path

## Observed behavior

With Lossless Claw `0.11.3+` / `0.12.0+`, Claude CLI backed turns fail before the model runs with an error shaped like:

```text
cannot run operation "agent-run" on CLI backend "claude-cli"
```

From inspecting the current `0.13.1` package, Lossless declares full `agent-run` host requirements:

```text
bootstrap
assemble-before-prompt
after-turn
maintain
compact
runtime-llm-complete
```

and the unsupported message says:

```text
lossless-claw requires a native OpenClaw runtime with the full context-engine agent-run lifecycle.
Use the native Codex or Pi embedded runtime, or switch plugins.slots.contextEngine to legacy for CLI harness runs.
```

That message is understandable, but the current behavior bricks the turn for `claude-cli` instead of letting the session continue.

## Expected behavior

When the active backend is known to be CLI-backed, especially `claude-cli`, Lossless Claw should fail gracefully or intelligently degrade.

Possible acceptable behaviors:

- Detect `claude-cli` / CLI-backed runtimes at registration or per-turn capability negotiation and opt out of full `agent-run` assembly for that run.
- Let the host continue the turn using raw/legacy context while logging that Lossless full assembly is unavailable for this backend.
- Continue post-turn transcript ingestion / afterTurn maintenance if available, even if before-prompt assembly and runtime LLM compaction are unavailable.
- Return an unsupported/degraded mode result that OpenClaw treats as nonfatal instead of blocking the model call.
- Provide a documented config flag such as `degradeOnUnsupportedCliBackend: true` or `cliBackendMode: "legacy" | "ingest-only" | "fail-closed"`.

[...]

PR #995, opened 2026-07-12
feat: add capture-only mode for CLI hosts

### LINKED ISSUE #849: Feature: opt-in degraded mode for claude-cli backend (capture-only) instead of hard-disable
## Summary

Since #740 (lossless 0.11.3 + OpenClaw 2026.5.22), configuring `contextEngine: lossless-claw` with a `claude-cli` agent backend throws a hard `agent-run` capability error (`Missing host capabilities: assemble-before-prompt, compact, runtime-llm-complete`) and deactivates the engine. I understand and agree with the rationale in #740 / #690: `claude-cli` can't project assembled context into the model prompt, so silently running with degraded context was the wrong default.

This is a request for a **third option** between "full engine" and "hard disable": an explicit, opt-in **capture-only / degraded mode** for users who run mixed fleets where some agents are on `claude-cli`.

## Motivation

Real setup (OpenClaw 2026.6.1, lossless-claw): a fleet with a primary agent on `anthropic/claude-sonnet-4-6` via the `claude-cli` backend (OAuth `claude setup-token`), plus other agents on native Codex/Pi runtimes.

After the 0.11+ change, the only ways forward are:
1. Drop `claude-cli` and move every agent to a native runtime — not always possible (OAuth/subscription-bound access).
2. Set `contextEngine: legacy` globally — loses lossless for the native-runtime agents too.

There's currently no per-backend or opt-in middle ground in `configSchema` (verified: no `allowUnsupportedHosts` / force flag exists in 0.12.0).

## Empirical note (current behavior)

On a downgraded 0.10.0 (no `agent-run` host requirement) the plugin loads and the agent replies, **but claude-cli turns are not persisted to `lcm.db`** — confirmed by absence of the test session in `conversations` after a live `claude-cli` turn. So the pre-#740 path is exactly the silent-loss situation #690 described; the hard error is strictly better than that. A documented degraded mode would let users *choose* capture-only behavior knowingly.

## Proposed behavior

- New config flag, e.g. `allowDegradedHosts: false` (default) or `hostFallbackMode: "error" | "capture-only" | "off"`.
- When enabled for an unsupported host (claude-cli):
  - Skip `assemble-before-prompt` (host owns the prompt) — no false promise of context injection.
  - Still run capture/persistence hooks where the host contract allows, so turns aren't a total blind spot.
  - Emit a loud, once-per-session warning describing exactly what is and isn't active.
- Keep the current hard error as the default, so nobody gets degraded context by surprise.

## Alternatives considered

- Per-agent context engine selection (lossless for native-runtime agents, legacy for claude-cli ones) would also solve it, if that's more aligned with the architecture.

## Environment

- OpenClaw: 2026.6.1 (2e08f0f)
- lossless-claw: 0.12.0 (and tested 0.10.0 / 0.11.3 behavior)
- claude-cli backend via `anthropic:claude-cli` OAuth profile
- Refs: #740, #690, #472

Adam-Researchh:
## Summary

Adds an explicit, opt-in `unsupportedHostMode: "capture-only"` compatibility mode for generic CLI-backed agent runtimes such as `claude-cli`.

Strict compatibility remains the default. When capture-only is enabled:

- `agent-run` requires the generic CLI lifecycle OpenClaw actually provides: `bootstrap`, `after-turn`, and `maintain`
- CLI turns can continue and be persisted/maintained instead of failing before inference
- LCM does not claim pre-prompt assembly, LCM-owned compaction, or runtime LLM completion for those CLI runs
- CLI-backed child context remains host-managed instead of requiring LCM thread bootstrap projection
- native Codex and Pi embedded runs still receive the full context-engine lifecycle in mixed-runtime installations

The mode is available through plugin config or `LCM_UNSUPPORTED_HOST_MODE=capture-only`. Enabling it emits a once-per-process warning describing the degraded guarantees.

## Why

Lossless Claw currently declares the full native `agent-run` capability set unconditionally. OpenClaw's generic CLI runner advertises `bootstrap`, `after-turn`, and `maintain`, so selecting a `claude-cli`-backed model fails before inference even though the host can safely capture and maintain the completed turn.

Fail-closed remains the right default, but mixed fleets need an explicit middle ground between full LCM projection and globally switching the context engine to `legacy`.

Closes #849.
Addresses #928.

## Validation

- `npm run typecheck`
- `npm run build`
- `npm test` (96 files, 1,726 tests)
- `npm pack --dry-run`
- `git diff --check`

## Safety

- default behavior is unchanged (`unsupportedHostMode: "error"`)
- no persistence, reconciliation, assembly, or compaction implementation was modified
- the relaxed capability list exactly matches OpenClaw's current generic CLI host capability set
- config schema, UI hints, exhaustive docs, bundled skill reference, and patch changeset are included

### DIFF
--- a/.changeset/soft-cli-capture.md
+++ b/.changeset/soft-cli-capture.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add an opt-in `unsupportedHostMode: "capture-only"` compatibility mode so generic CLI backends and CLI-backed children can continue agent turns while Lossless bootstraps, ingests, and maintains their transcripts without claiming context assembly, projected child context, or compaction support.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,8 +8,14 @@ a native host that provides the full context-engine lifecycle: session bootstrap
 pre-prompt assembly, after-turn ingestion, maintenance, compaction, and runtime
 LLM completion. Native Codex and Pi embedded runs provide those capabilities;
 generic CLI harnesses such as `claude-cli` and `codex-cli` do not. If you must
-use a generic CLI harness, set `plugins.slots.contextEngine` to `legacy` for
-that run instead of `lossless-claw`.
+use a generic CLI harness, either set `plugins.slots.contextEngine` to `legacy`
+or explicitly set `unsupportedHostMode` to `capture-only`. Capture-only mode
+allows hosts that provide bootstrap, after-turn ingestion, and maintenance to
+continue the agent turn and persist its transcript, but those runs do not
+receive LCM-assembled context or LCM-owned compaction. Native Codex and Pi
+embedded runs retain the full lifecycle in a mixed-runtime installation. CLI
+subagent context remains host-managed rather than receiving an LCM-projected
+parent context.
 
 The optional programmatic `status` / `doctor` / `rotate` control surface requires
 a host that separately advertises context-engine capabilities/control dispatch.
@@ -38,6 +44,7 @@ Most installations only need to override a handful of keys. If you want a comple
   "ignoreSessionPatterns": [],
   "statelessSessionPatterns": [],
   "skipStatelessSessions": true,
+  "unsupportedHostMode": "error",
   "contextThreshold": 0.75,
   "contextThresholdOverrides": [
     {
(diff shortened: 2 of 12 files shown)


jalehman:
Thanks for putting this together. The underlying capture-only mode is a real Lossless-Claw fit, and this PR includes useful repo-sync pieces: manifest schema, docs, bundled skill reference, and a changeset.

I'm going to close this one in favor of #996 as the better implementation baseline. #996 keeps the `subagent-spawn` / `thread-bootstrap-projection` contract intact, uses the clearer `hostFallbackMode` naming from the issue discussion, and includes live OpenClaw proof that default mode still fails closed while capture-only allows CLI turns, persists to `lcm.db`, and keeps recall working.

The good packaging work from this PR should be carried over to #996 before merge, along with clarified mixed-runtime wording and an explicit regression around the global `ownsCompaction` contract. Thanks again for pushing this forward.

Issue #971, opened 2026-07-06
Context engine silently deselected (slot → legacy) on OpenClaw ≥2026.6.1 when agent runs use a generic CLI backend — ingest freezes with no plugin-side signal


Flakedict:
## Environment

- lossless-claw: 0.13.2 (`openclaw.compat.tested: ["2026.5.28"]`, peerDependency `openclaw >=2026.5.28`)
- OpenClaw core: 2026.6.11 (first seen on 2026.6.1)
- Agent setup: main agent's model routing includes a CLI backend (`claude-cli`) used as a deliberate failover/circuit-breaker lane; primary runs use the Codex app-server harness.
- macOS, launchd-managed gateway.

## What happened

OpenClaw ≥2026.6.1 enforces context-engine host capabilities per operation. lossless-claw declares for `agent-run`:

```
requiredCapabilities: [bootstrap, assemble-before-prompt, after-turn, maintain, compact, runtime-llm-complete]
```

The generic CLI backend host only offers `[bootstrap, after-turn, maintain]`, so the first agent-run on the CLI backend failed with:

```
Context engine "lossless-claw" cannot run operation "agent-run" on CLI backend "claude-cli".
Missing host capabilities: assemble-before-prompt, compact, runtime-llm-complete.
```

Following the doctor guidance, `plugins.slots.contextEngine` was switched to `"legacy"`. Result: **all** ingest (messages/summaries) stopped — including for sessions running on fully capable hosts (Codex app-server, which does own context-engine compaction). The freeze was only discovered ~4 weeks later because:

1. The legacy engine keeps touching conversation metadata rows in `lcm.db`, so file mtimes/WAL look alive.
2. lossless logs go to the independent log file (default `/tmp/openclaw/lossless-claw-YYYY-MM-DD.log`, 3-day retention), so the "no afterTurn ever fires" signal ages out quickly.
3. There is no plugin-side health warning like "engine loaded but not selected for N days".

## What we expected

One of:
- a **per-run graceful degradation** (CLI-backend runs fall back to legacy behavior, capable-host runs keep using lossless) instead of an all-or-nothing global slot,
- or a documented **CLI-harness-compatible mode** of lossless-claw (reduced capability set for CLI runs),
- or at minimum a **loud freshness warning** from the plugin when it is loaded but has not ingested for an extended period (`MAX(created_at)` stale while the gateway is active).

## Repro sketch

[...]

PR #740, opened 2026-05-21
fix: declare context engine host requirements

jalehman:
## What
Declares that lossless-claw requires OpenClaw agent runtimes that assemble context before prompt construction, and updates package metadata/docs to require the first stable OpenClaw release that contains that host-compatibility contract.

## Why
Generic CLI harnesses like claude-cli and legacy codex-cli do not project lossless-claw assembled context into the model prompt, so they should fail with a descriptive compatibility error instead of silently running with degraded context.

## Changes
- Added agent-run host requirement
- Required OpenClaw 2026.5.22
- Updated OpenClaw compatibility metadata
- Documented CLI harness fallback
- Added regression coverage
- Added patch changeset

## Release Verification
- OpenClaw PR `openclaw/openclaw#84994` merged as `cff5244a5b25a8e9fc76fb51dd5915a5a8f3eda6`.
- `git merge-base --is-ancestor cff5244a5b25a8e9fc76fb51dd5915a5a8f3eda6 v2026.5.22` passed.
- `git merge-base --is-ancestor cff5244a5b25a8e9fc76fb51dd5915a5a8f3eda6 v2026.5.20` failed, confirming 2026.5.20 does not contain it.
- GitHub Releases lists `v2026.5.22` as stable and published.
- npm `openclaw` latest is `2026.5.22`.

## Testing
- `npx vitest run test/engine.test.ts test/package-metadata.test.ts` - passed, 2 files / 234 tests.
- `npm run build` - passed.
- `npm install --package-lock-only` - passed.
- `npm ci --package-lock-only --ignore-scripts` - passed.
- `git diff --check` - passed.

## Real Behavior Proof
- Scenario: lossless-claw declares `agent-run.requiredCapabilities = ["assemble-before-prompt"]`.
- Command/path exercised: package metadata regression coverage and OpenClaw live gateway startup on the companion host-compatibility PR branch.
- Observed result: OpenClaw 2026.5.22 includes the host contract that accepts native Codex and Pi embedded runtimes, rejects unsupported CLI-style hosts with a descriptive missing-capability error, and lets lossless-claw require that behavior in published package metadata.

PR #996, opened 2026-07-12
feat: opt-in `hostFallbackMode: "capture-only"` for CLI-backed hosts

Kaspnov:
Implements #849 (opt-in capture-only mode); refs #928 and #971.

## Problem

Since 0.11.3 (#740), the engine declares `agent-run` host requirements of
`bootstrap, assemble-before-prompt, after-turn, maintain, compact, runtime-llm-complete`.
OpenClaw's generic CLI backends (e.g. `claude-cli`) advertise only
`bootstrap, after-turn, maintain`, and OpenClaw core enforces the requirement fail-closed, so
every CLI-backed turn dies before reaching the model:

```
Context engine "lossless-claw" cannot run operation "agent-run" on CLI backend "claude-cli".
Missing host capabilities: assemble-before-prompt, compact, runtime-llm-complete.
```

To the operator this surfaces as a silent no-reply on chat channels (openclaw/openclaw#93337
describes the signature). Setting the slot back to `legacy` un-breaks replies but freezes ingest
fleet-wide (#971) — all-or-nothing. OpenClaw core rejected automatic per-session demotion
(openclaw/openclaw#93569) as a reversal of the fail-closed contract, and pointed at a product-level
opt-in instead. This PR adds that opt-in where it belongs: in the engine's own declaration.

## Change

New plugin config key `hostFallbackMode` (env override `LCM_HOST_FALLBACK_MODE`):

- `"error"` (default) — exactly today's behavior: full lifecycle required, fail closed on CLI
  hosts. The unsupported-message now additionally mentions the opt-in.
- `"capture-only"` — the engine declares `agent-run` requirements of
  `bootstrap, after-turn, maintain` (the generic-CLI host set; anything weaker still fails
  closed). CLI-backed turns then run with full transcript capture into lcm.db and all recall
  tools, but WITHOUT lossless prompt assembly — the CLI harness owns the prompt, so there is no
  `assemble-before-prompt` seam to project into. This is an explicit, documented trade-off chosen
  by the operator, not a silent degradation.

Details:

- Native runtimes (Codex / Pi embedded) advertise the full capability set, so capture-only never
  changes behavior there — hosts invoke every seam they support regardless; `hostRequirements`
  is a fail-closed minimum, not a negotiation.
- The `subagent-spawn` requirement (`thread-bootstrap-projection`) is untouched.
- A one-time warning is logged at engine construction when capture-only is active, spelling out
  that assembly is not projected on CLI hosts and that summary compaction there needs
  `fallbackProviders` (no `runtime-llm-complete`).
- Unrecognized values resolve to `"error"`.
- Manifest `configSchema` extended (it is `additionalProperties: false`).

## Testing

- `npm run typecheck` clean.
- Full suite: 1730 passed, 0 failed.
- New tests: config resolution (default / plugin-config / env precedence / bogus value) and
  engine metadata in capture-only mode (reduced agent-run set, unchanged subagent-spawn).
- End-to-end on OpenClaw 2026.7.1-beta.5 with a claude-cli agent behind a live gateway:
  - default mode: reproduces the pre-model fail-closed kill verbatim;
  - capture-only: replies flow, engine active, one-time warning logged, all turns captured to
    lcm.db (including startup backfill of a pre-activation session), recall tools functional.

### DIFF
--- a/.changeset/soft-cli-capture.md
+++ b/.changeset/soft-cli-capture.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add an opt-in `hostFallbackMode: "capture-only"` setting so generic CLI backends can persist turns and use recall tools without Lossless prompt assembly or compaction. Strict full-lifecycle validation remains the default, and subagent projection requirements remain unchanged.
--- a/.changeset/soft-cli-capture.md
+++ b/.changeset/soft-cli-capture.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Add an opt-in `hostFallbackMode: "capture-only"` setting so generic CLI backends can persist turns and use recall tools without Lossless prompt assembly or compaction. Strict full-lifecycle validation remains the default, and subagent projection requirements remain unchanged.
+Add an opt-in `hostFallbackMode: "capture-only"` setting so generic CLI backends can persist turns and use recall tools without Lossless prompt assembly or host-triggered Lossless compaction. Strict full-lifecycle validation remains the default, backend-native compaction remains host-owned, and subagent projection requirements remain unchanged.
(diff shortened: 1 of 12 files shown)


jalehman:
Thanks for the contribution, @Kaspnov. The capture-only host fallback is landed now, including the config/schema/docs coverage, one-time warning, tests, and release changeset.

PR #1000, opened 2026-07-14
fix: safely deduplicate missing-transcript replays

ralf003:
## What
Handle `afterTurn` when Lossless cannot read a tracked transcript but retains its checkpoint. The reconciler emits `transcriptUnavailableWithCheckpoint`, and the dedup path suppresses a full runtime replay only when every transcript entry ID is already persisted.

## Why
`transcriptCovered` means the transcript was read through its frontier, so the missing-file path cannot assert it. Treating second-truncated timestamps as replay proof can discard a genuinely new repeated row. Exact transcript entry IDs provide deterministic proof while ambiguous batches retain the established degraded dedup behavior.

## Changes

- Add missing-transcript checkpoint signal
- Require persisted IDs for replay drops
- Preserve degraded ambiguous-batch dedup
- Cover full and partial overlaps
- Add patch release changeset

## Testing

- `npm test`: 105 files passed
- `npm run typecheck`: passed
- `npm run build`: passed
- Exact merge tree: 1,895 tests passed
- Final autoreview: no actionable findings

### DIFF
--- a/.changeset/missing-transcript-checkpoint-dedup.md
+++ b/.changeset/missing-transcript-checkpoint-dedup.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Avoid duplicate runtime rows when a tracked transcript is temporarily unavailable while preserving genuinely new rows from partially overlapping batches.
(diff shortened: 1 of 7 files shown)


jalehman:
Thanks for the contribution, @ralf003. This is merged now.

PR #1002, opened 2026-07-14
fix(engine): skip backoff in assemble emergency drain with force=true

mpz4life:
## What

When `assemble()` sees pending deferred compaction debt and prompt pressure above `tokenBudget`, it can bypass the debt row's retry backoff and run emergency compaction. The bypass applies only while `retryAttempts < 3`. The same `force` value reaches `executeCompactionCore`, which clears an active summary-spend backoff and runs the threshold recovery sweep.

Normal background and `maintain()` drains do not pass `force`. They continue to honor `nextAttemptAfter` and use the persisted or configured compaction sweep sizing.

## Why

A failed or incomplete compaction can leave deferred debt pending with an active retry backoff. If the prompt grows beyond its token budget before that timer expires, the previous assemble path returns `"deferred compaction backoff active"` and cannot recover the session.

The reported session compressed from about 165K tokens to 40K, then grew to 213K while the backoff blocked another emergency attempt. A bounded bypass lets assemble retry under overflow pressure without removing the spend guard from routine compaction.

## Changes

- Add an assemble-only backoff bypass
- Cap forced retries at three attempts
- Forward `force` through compaction core
- Preserve normal drain backoff behavior
- Preserve configured compaction sweep sizing
- Add focused boundary and sizing tests
- Add a patch Changeset

## Guardrails

- Assemble enters this path only when prompt pressure exceeds `tokenBudget`.
- `retryAttempts` values 0 through 2 allow the bypass.
- `retryAttempts >= 3` restores normal backoff behavior and degraded prompt fallback.
- Routine background and `maintain()` drains never enable the bypass.
- The change does not add dynamic `leafChunkTokens` or `freshTailMaxTokens` overrides.

## Testing

- `npm test`: 80 files and 1,463 tests passed
- `npm run typecheck`: passed
- `npm run build`: passed
- Focused tests cover:
  - assemble bypassing active debt backoff
  - `force` reaching `executeCompactionCore`
  - retry attempts 2 and 3 around the cap
  - normal drains retaining configured sweep sizing
  - provider fallback and message-preservation bounds

Hosted CI passed:

- `test`
- `plugin-inspector`
- `smoke-latest-openclaw`

## Security impact

This change adds no permissions, secrets handling, network calls, command execution, or data-access scope. The retry cap bounds additional summarizer spend during repeated failures.


mpz4life:
@jalehman  Could you please review this PR when you get a chance? It addresses a compaction deadlock scenario where the assemble emergency drain (pressureTokenCount > tokenBudget) is blocked by the deferred compaction retry backoff, preventing context recovery from overflow. The fix is minimal — a force flag on consumeDeferredCompactionDebt that skips the nextAttemptAfter check only in the assemble emergency path. Normal deferred drains (drainDeferredCompactionDebtIfIdle, maintain) continue to respect the backoff as before.

mpz4life:
**P1 addressed in de3e51d.**

Added a `retryAttempts < 10` gate in `maybeConsumeDeferredCompactionDebtForAssemble`: when `retryAttempts >= 10`, `force` defaults to `false` (i.e. `undefined`), causing the emergency drain to fall back to respecting the backoff and degrading the live prompt instead of retrying compaction.

10 was chosen because each force attempt compresses up to `leafChunkTokens` (20K) of raw messages outside the fresh tail. A worst-case backlog of ~200K outside-tail tokens needs at most 10 rounds. If compaction still fails after 10 attempts, there is likely nothing left to compact and the degraded fallback is the correct path.

jalehman:

PR #1003, opened 2026-07-15
fix: add transcriptCovered:true to ENOENT early returns in fullReadAfterTurnReconcile

ralf003:
## Summary

Fixes two ENOENT early-return paths in `fullReadAfterTurnReconcile` that were missing `transcriptCovered: true`, causing the engine to route through the wrong ingest path and produce duplicate messages.

## Changes

1. **ENOENT early return (before checkpoint)**: Refactored to use the same return path as the checkpoint-present case, which correctly includes `transcriptCovered: true`.
2. **ENOENT checkpoint-present return**: Added `transcriptCovered: true` to the return object so the engine routes through `alignRuntimeBatchAgainstCoveredFrontier` instead of the aggressive ingest path.
3. **Test update**: Updated "ingests oversized no-overlap batches without transcript coverage" test to expect correct deduplication behavior (no duplicate messages) instead of the previous buggy behavior.

## Testing

- Build: passes (`npm run build`)
- Test suite: 1808/1818 tests pass (10 pre-existing Windows path failures unrelated to this fix)
- The ENOENT-related scenarios now correctly deduplicate instead of ingesting duplicates

### DIFF
--- a/src/transcript-reconciler.ts
+++ b/src/transcript-reconciler.ts
@@ -2127,11 +2127,15 @@ export class TranscriptReconciler {
         this.host.deps.log.warn(
           `[lcm] afterTurn: session file missing; skipping transcript reconcile full reread; could not stat/read transcript; allowing live afterTurn persistence and seeding placeholder bootstrap_state at offset=0 to unblock next-turn recovery conversation=${conversation.conversationId} reason=${reason} sessionFile=${params.sessionFile}`,
         );
-      } else {
-        this.host.deps.log.warn(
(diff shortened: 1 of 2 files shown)


jalehman:
Thanks for working on this. I’m going to close this PR rather than hold it open because this same head is already open in #1000, and the current approach is not safe for Lossless as written.

The core issue is that the ENOENT-with-checkpoint path cannot return `transcriptCovered: true`: the transcript file was not read, so the covered-frontier contract is not satisfied. Routing that missing-file path through the covered dedup path can fail closed on a partially-overlapping live batch and drop a genuinely new message. The existing test only covers the all-already-persisted case.

A replacement should keep the missing-file path as not covered, preserve new mixed-overlap runtime rows, and add a regression test for the partial-overlap case.

PR #1007, opened 2026-07-16
fix: use context_items instead of inflated liveMessages for debt pressure

mpz4life:
## Summary

`resolveDeferredAssemblyPressure()` takes `Math.max(liveContextTokens, recordedContextTokens)` to conservatively estimate assembly pressure when deferred compaction debt is pending. After a compact successfully reduces stored context, the recorded token count from debt creation time becomes stale, forcing the system into permanent emergency drain with an impossible compaction target.

## Root cause

Two issues:

1. **`liveContextTokens` source**: `estimateSessionTokenCountForAfterTurn(liveMessages)` estimates ALL raw session messages (~210K), not the compressed context_items (~47K). Even after a successful compact, this still returns the inflated count.

2. **`resolveDeferredAssemblyPressure` max**: `Math.max(liveContextTokens, recordedContextTokens)` picks up the stale debt-time API input count (168K) even when current context is 47K. This makes the emergency drain compute `runtimeAdjustedSweepTargetTokens = max(1, target - overhead)` where overhead is the stale gap, resulting in a target of ~1 token. Compaction cannot meet this, so the debt stays pending forever.

## Fix

- **`engine.ts`**: `liveContextTokens` uses `summaryStore.getContextTokenCount()` (47K) instead of `estimateSessionTokenCountForAfterTurn(liveMessages)` (210K)
- **`assemble-fallback.ts`**: `resolveDeferredAssemblyPressure` uses `liveContextTokens` directly without inflating with stale recorded values

## Verification

Tested on a production session with 1099 turns and pending debt:

**Before fix (stuck in deadlock):**
```
mode: pending=1, current=167934, projected=182408
failure: "compacted but still over target"
behavior: each assemble() triggers emergency drain → compact target ~1 → fail → degraded live fallback → "Context is too large" error
```

**After fix (recovered normally):**
```
mode: pending=0, current=74690, projected=92280
failure: (cleared)
behavior: assemble() uses context_items (47K) within budget → normal assembly → session resumed with no errors
```

**Test suite:** 1815/1818 tests pass (3 pre-existing log redaction failures unrelated to this change)

### DIFF
--- a/.changeset/calm-context-pressure.md
+++ b/.changeset/calm-context-pressure.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Prevent stale deferred-compaction token counts from forcing repeated emergency drains after the stored context has already been compacted.
--- a/src/assemble-debug.ts
+++ b/src/assemble-debug.ts
@@ -152,13 +152,13 @@ export function describeAssembledPrefixChange(
 export function shouldLogOverflowDiagnostics(params: {
   diagnostics: AssemblyOverflowDiagnostics;
   assembledTokens: number;
-  liveContextTokens: number;
+  storedContextTokens: number;
 }): boolean {
   const budget = Math.max(1, params.diagnostics.tokenBudget);
   return (
     params.diagnostics.totalContextTokens > budget ||
     params.assembledTokens >= Math.floor(budget * 0.9) ||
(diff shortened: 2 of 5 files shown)


mpz4life:
@100yenadmin @jalehman  could you please review this change? Is it possible to directly use the context-items in the database as the context to estimate the token usage of the current context?

jalehman:

PR #1011, opened 2026-07-18
fix(compaction): guard leafPass against empty messageContents

mpz4life:
## Summary

When `selectOldestLeafChunk` returns items whose `messageId` fields are all null, or whose referenced messages no longer exist, `messageContents` in `leafPass` is empty. Previously the method would still call the summarizer, create a summary with `removedTokens=0`, and grow the context (`tokensAfter = tokensBefore + tokenCount`).

## Fix

Added an early return guard after building `messageContents` — if empty, `leafPass` logs a warning and returns `null`, letting callers bail without wasted LLM calls or context inflation.


jalehman:

PR #1014, opened 2026-07-20
fix(afterTurn): use context_items token count as fallback instead of full session messages

mpz4life:
## Problem

When the runtime does not provide a live prompt token count in `afterTurn`, estimating all session messages overstates the current compacted context and can schedule unnecessary compaction.

## Fix

- Leave the live observation unset when OpenClaw supplies no prompt token count
- Let `CompactionEngine.evaluate()` use its canonical stored `context_items` count
- Preserve runtime or supplied token counts as the preferred live signal
- Add a regression covering raw messages outside the protected fresh tail

## Testing

- `npx vitest run test/engine-after-turn.test.ts`
- `npm run typecheck`
- `npm test`
- `npm run build`
- `npm pack --dry-run`

### DIFF
--- a/.changeset/honest-afterturn-pressure.md
+++ b/.changeset/honest-afterturn-pressure.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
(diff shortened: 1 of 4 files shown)


mpz4life:
@jalehman Could you please review my code when you have a moment? Thank you!

jalehman:

PR #1021, opened 2026-07-21
feat(plugin): register lcm-control session action

ArthurNie:
## Problem

`LcmContextEngine` already implements `control()` supporting `status` / `doctor` / `rotate`, backed by `runLcmProgrammaticControl()` and a structured failure classifier (`classifyProgrammaticRotateUnavailableReason`). All of that machinery is unreachable from outside the plugin.

The reason is subtle: `ContextEngineControlRequest` / `ContextEngineControlResult` / the optional `control?()` member exist **only in this plugin's own `src/openclaw-bridge.ts`**. The OpenClaw host has no such contract — grepping the host's shipped `dist/` for `ContextEngineControl` returns nothing. So `control()` is implemented against an interface the host never calls.

The host's actual extension point for this is `api.session.controls.registerSessionAction`, which the plugin does not use.

## Change

Wiring only — **no rotation logic is added**:

- `src/openclaw-bridge.ts` — declare the host's `registerSessionAction` surface and its context/result types
- `src/plugin/index.ts` — register `lcm-control`, delegating to the existing `engine.control()`
- `openclaw.plugin.json` — declare `contracts.sessionActions`

The manifest entry is required: registrations not declared there are silently dropped by the host, with no error.

57 insertions, 0 deletions.

## Verification

Built and installed against a live OpenClaw gateway (2026.7.1-2), plugin v0.14.0:

```
$ openclaw gateway call plugins.sessionAction --json \
    --params '{"pluginId":"lossless-claw","actionId":"lcm-control","payload":{"operation":"status"}}'
{"ok": true, "result": {"operation": "status", "active": false, "messageCount": 0}}
```

Real `rotate` against a canary session:

```
[ws] ⇄ res ✓ plugins.sessionAction 90651ms
transcript: 2344 bytes -> 2088 bytes
```

Before the gateway restart the same call returned `unknown plugin session action: lossless-claw/lcm-control`, confirming the registration is what makes it reachable.

## Note for reviewers: a timeout footgun

The gateway client transport times out at **10000ms**, while `rotate` legitimately took **~90s** on a 2.3KB transcript. Our first call surfaced:

```
{"ok": false, "error": {"kind": "timeout", "message": "gateway timeout after 10000ms"}}
```

…even though the operation succeeded server-side. Callers must treat a client timeout as **inconclusive**, not failed, and re-read state before retrying — a naive retry would re-run a rotation that already happened.

Happy to split the bridge type declarations into a separate commit if you'd prefer, or to adjust the action id / add `requiredScopes` to match your intended authorization model — we left `requiredScopes` unset since we did not want to guess at your scope taxonomy.

### DIFF
--- a/.changeset/session-action-lcm-control.md
+++ b/.changeset/session-action-lcm-control.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add the `lcm-control` OpenClaw session action so authorized operators can invoke Lossless Claw control operations through the host session-action API.
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -24,6 +24,9 @@
       "lcm_describe",
       "lcm_expand",
       "lcm_expand_query"
+    ],
+    "sessionActions": [
+      "lcm-control"
     ]
   },
   "uiHints": {
--- a/src/openclaw-bridge.ts
+++ b/src/openclaw-bridge.ts
@@ -184,8 +184,33 @@ export type OpenClawPluginApi = {
     eventName: string,
     handler: (event: PluginLifecycleEvent, ctx: PluginLifecycleContext) => unknown | Promise<unknown>,
   ) => void;
+  session?: {
+    controls?: {
+      registerSessionAction?: (action: PluginSessionActionRegistration) => void;
+    };
+  };
   [key: string]: any;
 };
+
+export type PluginSessionActionRegistration = {
+  id: string;
+  description?: string;
+  schema?: unknown;
+  requiredScopes?: string[];
+  handler: (ctx: PluginSessionActionContext) => Promise<PluginSessionActionResult>;
+};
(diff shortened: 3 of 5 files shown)


jalehman:
Thanks, this looks like the right host integration point and the current CI is green. Before merge, please add a patch changeset for the new user-visible `lcm-control` session action.

I do not think the OpenClaw compatibility floor needs to change, since `2026.5.28` already has the session-action API and default `operator.write` gating, but the release-note gate still applies. Optional follow-ups: add invalid/missing-operation tests and document or explicitly declare the intended scope behavior for rotate.

ArthurNie:
Added the requested patch changeset in `22a05fe`; all current-head CI checks are green and the PR is cleanly mergeable. Ready for re-review/merge.

jalehman:
Thanks for the contribution, @ArthurNie. This is landed now.

PR #1030, opened 2026-07-23
fix(manifest): remove duplicate largeFilesDir declarations

cxbAsDev:
`openclaw.plugin.json` contained duplicate `largeFilesDir` keys in both `uiHints` and `configSchema.properties`. JSON parsers keep the last value, so the first declaration was silently ignored. The surviving entries now consistently describe the default as `<OPENCLAW_STATE_DIR>/lcm-files`, matching the runtime resolver and `docs/configuration.md`.

Changes:
- Removed the duplicate `uiHints.largeFilesDir` entry that described a hard-coded `~/.openclaw/lcm-files` default.
- Removed the bare duplicate `configSchema.properties.largeFilesDir` entry so the described entry survives.
- Added a regression test in `test/manifest.test.ts` asserting the surviving descriptions reference `OPENCLAW_STATE_DIR`.

Verification: `npm test -- test/manifest.test.ts` passes.

### DIFF
--- a/.changeset/plugin-json-duplicate-largefilesdir.md
+++ b/.changeset/plugin-json-duplicate-largefilesdir.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Remove duplicate `largeFilesDir` declarations from `openclaw.plugin.json`. The surviving entries describe the default as relative to `OPENCLAW_STATE_DIR`, consistent with the runtime resolver and configuration reference. A regression test now guards against duplicate keys and stale default descriptions.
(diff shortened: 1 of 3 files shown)


jalehman:
Thanks, this is the right scope and the manifest duplicate-key cleanup is a good fit. Before maintainer review, please tighten two small things:

- Keep the surviving `largeFilesDir` UI help aligned with runtime/docs by mentioning inline-image payloads as well as large-file text payloads.
- Make the regression test inspect the raw manifest text or use a duplicate-key-aware parser. The current test imports parsed JSON, so duplicate keys would already be collapsed before the assertion and the duplicate-key regression would not actually be caught.

Once those are fixed, this should be ready for a full review pass.

jalehman:

PR #1032, opened 2026-07-24
fix(engine): recover safe no-anchor bootstrap checkpoints

ralf003:
## What

Recover ingest-before-bootstrap conversations whose persisted frontier contains only proven non-anchoring OpenClaw metadata. Bootstrap imports the readable transcript through the existing guarded no-anchor recovery path before marking the conversation bootstrapped and persisting its checkpoint.

## Why

Without a checkpoint, subsequent `afterTurn` reconciliation can remain stuck at `checkpoint-missing` forever. Persisting an EOF checkpoint without lineage proof is also unsafe because it can splice an unrelated transcript into real stored history. This change restores progress only where the existing bounded frontier proof shows there is no real history to contaminate.

## Changes

- Reuse non-anchoring frontier proof
- Import transcript before checkpointing
- Preserve real-history fail-closed behavior
- Preserve unreadable-transcript retry behavior
- Prove `afterTurn` and re-bootstrap recovery
- Add patch release changeset

## Testing

- `npx vitest run test/engine-bootstrap.test.ts` — 73 passed
- `npm test` — 1,896 passed
- `npm run typecheck` — passed
- `npm run build` — passed
- `npm pack --dry-run` — passed
- Autoreview against `origin/main` — clean

### DIFF
--- a/.changeset/bootstrap-persist-checkpoint-no-overlap.md
+++ b/.changeset/bootstrap-persist-checkpoint-no-overlap.md
@@ -1,0 +1,18 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix a race in the bootstrap existing-conversation path where a conversation
+with a non-anchoring DB frontier (no overlapping content between the stored
+messages and the JSONL transcript) would have `bootstrapped_at` set but no
+`conversation_bootstrap_state` row persisted.
+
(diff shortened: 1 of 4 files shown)


jalehman:
Thanks for the contribution, @ralf003. This is landed now.

PR #1012, opened 2026-07-20
fix: respect OPENCLAW_STATE_DIR in all code paths

mpz4life:
## Summary

Three runtime code paths hardcoded `~/.openclaw` instead of respecting the `OPENCLAW_STATE_DIR` environment variable. All now follow the established pattern (`env var` → `~/.openclaw` fallback).

### Fixes

| File | Issue |
|---|---|
| `scripts/stub-tier-live-watcher.mjs` | Default log/DB paths ignored `OPENCLAW_STATE_DIR` |
| `scripts/stub-tier-assemble-bench.mjs` | `largeFilesDir` default ignored `OPENCLAW_STATE_DIR` |
| `tui/data.go` | `resolveDataPaths()` hardcoded `~/.openclaw` |

### Tests

- `tui/data_test.go` — new Go tests for `resolveOpenclawStateDir` and `resolveDataPaths` with/without env var
- `test/migrate-sessions.test.ts` — new TS tests for `defaultStateDir` covering unset, set, empty, whitespace, and trimmed values

### All 1888 existing tests pass
```
npm test
  Test Files  105 passed (105)
  Tests  1888 passed (1888)
```

### Script execution evidence

**stub-tier-live-watcher with env var:**
```
$ OPENCLAW_STATE_DIR=/custom/state node -e "const d = process.env.OPENCLAW_STATE_DIR?.trim() || require('os').homedir() + '/.openclaw'; console.log('log:', d + '/logs/gateway.log'); console.log('db:', d + '/lcm.db');"
log: /custom/state/logs/gateway.log
db: /custom/state/lcm.db
```

**stub-tier-live-watcher without env var (fallback):**
```
$ unset OPENCLAW_STATE_DIR; node -e "const d = process.env.OPENCLAW_STATE_DIR?.trim() || require('os').homedir() + '/.openclaw'; console.log('log:', d + '/logs/gateway.log'); console.log('db:', d + '/lcm.db');"
log: /root/.openclaw/logs/gateway.log
db: /root/.openclaw/lcm.db
```

**stub-tier-assemble-bench with env var:**
```
$ OPENCLAW_STATE_DIR=/custom/state node -e "const d = process.env.OPENCLAW_STATE_DIR?.trim() || require('os').homedir() + '/.openclaw'; console.log('largeFilesDir:', d + '/lcm-files');"
largeFilesDir: /custom/state/lcm-files
```

PR #1033, opened 2026-07-25
fix(config): remove duplicate largeFilesDir key in configHints

### LINKED ISSUE #1022: openclaw.plugin.json: duplicate `largeFilesDir` key in configHints, stale variant wins
## Summary

`openclaw.plugin.json` defines `largeFilesDir` **twice within the same object**, with conflicting help text. Per JSON semantics the **last** one wins, and in this case the surviving text is the stale one — it contradicts the schema and the behaviour shipped in #1012.

## Detail

v0.14.0, `openclaw.plugin.json` lines 133–140:

```json
    "largeFilesDir": {
      "label": "Large Files Directory",
      "help": "Directory for persisting large-file text payloads (default: <stateDir>/lcm-files). Uses OPENCLAW_STATE_DIR when set."
    },
    "largeFilesDir": {
      "label": "Large Files Directory",
      "help": "Directory for externalized large files and inline-image payloads (default: ~/.openclaw/lcm-files)"
    },
```

The second entry documents the default as `~/.openclaw/lcm-files` and makes no mention of `OPENCLAW_STATE_DIR`.

Meanwhile the JSON schema block in the same file (line ~693) says:

```json
      "largeFilesDir": {
        "description": "Directory for persisting large-file text payloads (default: <OPENCLAW_STATE_DIR>/lcm-files)",
        "type": "string"
      },
```

So the first variant is the one consistent with the schema and with `fix: respect OPENCLAW_STATE_DIR in all code paths` (#1012) — but it is the one that gets discarded.

## Impact

Config-hint consumers surface the pre-#1012 description, telling operators the path is `~/.openclaw/lcm-files` when `OPENCLAW_STATE_DIR` is honoured. Cosmetic rather than functional, but it points users at the wrong directory when diagnosing large-file storage.

Note the other two `largeFilesDir` occurrences (lines ~479 and ~693) are in different objects and are not duplicates.

## Suggested fix

Delete the second entry (lines 137–140), keeping the `OPENCLAW_STATE_DIR`-aware text.

Found while preparing #1021; deliberately left out of that PR to keep its diff scoped to the session-action wiring.

syltharion:
Closes #1022

The second `largeFilesDir` entry in `configHints` had stale text contradicting the schema and `OPENCLAW_STATE_DIR` support (#1012). Remove it, keeping the correct entry.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -131,10 +131,6 @@
       "label": "Large Files Directory",
       "help": "Directory for persisting large-file text payloads (default: <stateDir>/lcm-files). Uses OPENCLAW_STATE_DIR when set."
     },
-    "largeFilesDir": {
-      "label": "Large Files Directory",
-      "help": "Directory for externalized large files and inline-image payloads (default: ~/.openclaw/lcm-files)"
-    },
     "ignoreSessionPatterns": {
       "label": "Ignored Sessions",
       "help": "Glob patterns for session keys to exclude from LCM storage"
@@ -471,9 +467,6 @@
         "minimum": 2
       },
       "dbPath": {
-        "type": "string"
-      },
-      "largeFilesDir": {
         "type": "string"
       },
       "ignoreSessionPatterns": {


jalehman:
Thanks, this fixes a real manifest issue and the deletion itself looks directionally right.

I am going to close this in favor of #1030, which already covers the same duplicate `largeFilesDir` cleanup with a changeset and test scaffold. The remaining work should happen there: make the regression test inspect raw manifest text or use a duplicate-key-aware parser, and keep the surviving `largeFilesDir` help aligned with runtime/docs by mentioning inline-image payloads as well as large-file text payloads.

PR #1034, opened 2026-07-25
fix(doctor-apply): include message role in buildLeafSourceText

### LINKED ISSUE #1027: buildLeafSourceText drops message role on the doctor-apply path
## Summary

`buildLeafSourceText` in `src/plugin/lcm-doctor-apply.ts` builds the summarizer input
from `created_at` and `content` only:

```sql
SELECT m.created_at, COALESCE(m.content, '') AS content
  FROM summary_messages sm
  JOIN messages m ON m.message_id = sm.message_id
 WHERE sm.summary_id = ?
 ORDER BY sm.ordinal ASC
```

```ts
rows.map((row) => `[${formatSqliteTimestamp(row.created_at, params.timezone)}]\n${row.content}`)
```

`messages.role` is available but not selected, so a regenerated leaf summary cannot
distinguish an operator instruction from assistant reasoning or from material a tool
fetched out of another conversation.

## Relationship to #1026

#1026 fixes the equivalent defect on the **production compaction path**
(`src/compaction.ts`), which is where live summaries are produced. This issue covers
the **doctor-apply path**, which regenerates summaries during repair.

They are filed separately on purpose: different code paths, different blast radius,
and keeping the production fix reviewable on its own.

## Why it matters here too

Doctor-apply regenerates summaries over history that may already contain
tool-fetched cross-conversation material. Without provenance, a repair pass can
reintroduce the same intent-promotion the production fix is meant to prevent —
and because it runs over stored history, the rewritten summary persists.

## Suggested fix

Mirror #1026: add `m.role` to the SELECT and emit it on the header line
(`[<timestamp> | <role>]`), leaving bodies unchanged. No schema migration needed.

Not submitting a patch for this one — flagging it so the two paths do not drift.

syltharion:
Closes #1027

Mirrors the fix from #1026 (production compaction path) to the doctor-apply repair path. Adds `m.role` to the SQL SELECT and emits it on the header line as `[<timestamp> | <role>]`.

Without this, doctor-apply cannot distinguish operator instructions from assistant reasoning or tool-fetched cross-conversation material.

### DIFF
--- a/.changeset/doctor-repair-role-provenance.md
+++ b/.changeset/doctor-repair-role-provenance.md
@@ -1,0 +1,6 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve each message's role in leaf-summary source text rebuilt by `doctor apply`,
+so repaired summaries can distinguish operator input from assistant and tool content.
--- a/src/plugin/lcm-doctor-apply.ts
+++ b/src/plugin/lcm-doctor-apply.ts
@@ -335,19 +335,19 @@ function buildLeafSourceText(params: {
 }): string {
   const rows = params.db
     .prepare(
-      `SELECT m.created_at, COALESCE(m.content, '') AS content
+      `SELECT m.created_at, m.role, COALESCE(m.content, '') AS content
        FROM summary_messages sm
        JOIN messages m ON m.message_id = sm.message_id
        WHERE sm.summary_id = ?
        ORDER BY sm.ordinal ASC`,
     )
-    .all(params.target.summaryId) as Array<{ created_at: string; content: string }>;
+    .all(params.target.summaryId) as Array<{ created_at: string; role: string | null; content: string }>;
   if (rows.length === 0) {
     throw new Error("no messages linked to summary");
   }
 
   return rows
-    .map((row) => `[${formatSqliteTimestamp(row.created_at, params.timezone)}]\n${row.content}`)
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for picking this up. The runtime fix is the right minimal mirror of #1026, but I would like the regression test to drive the actual doctor-apply path before we do full maintainer review.

Right now `test/leaf-source-text-role.test.ts` copies the private SQL and timestamp formatting into the test, so it would still pass if `buildLeafSourceText` regressed back to omitting `m.role`. Please replace it with a focused test that seeds a broken leaf summary with linked messages, calls `applyScopedDoctorRepair` or the `/lossless doctor apply` command path with a fake summarizer that captures `sourceText`, and asserts the captured source contains `[<timestamp> | <role>]` headers for the linked roles while leaving message bodies unchanged.

Maintainers can handle the changeset decision before merge if needed.

jalehman:
Maintainer follow-up is pushed in 7af0cd8. The copied-private-SQL test is replaced with a focused `applyScopedDoctorRepair` test over a migrated database and real stores. Its capturing fake summarizer asserts the complete `[timestamp | role]` source text for user, tool, and assistant messages, including unchanged multiline bodies. Focused test, full 1,906-test suite, typecheck, build, thermonuclear review, and structured autoreview all pass. I also added the required patch changeset.

jalehman:
Thanks for the contribution, @syltharion. This is landed now.

PR #1035, opened 2026-07-25
fix(assembler): add __openclaw.kind=compaction to injected summary

### LINKED ISSUE #930: Compaction summary missing __openclaw.kind=compaction marker — Dashboard expander not rendered
## Problem

After LCM compaction runs, the compaction summary is injected into the conversation but the Dashboard (Control UI) does not render the "Compacted history" divider with expand/collapse functionality. The compaction summary appears as a regular assistant message without any UI affordance to access the original messages.

## Root Cause

The Dashboard frontend renders a "Compacted history" divider with an expander when it detects `__openclaw.kind=compaction` on a message. LCM does not set this marker on its compaction summaries, so the frontend treats them as normal messages.

## Environment

- lossless-claw version: 0.12.0
- OpenClaw version: 2026.6.10-beta.2
- Channel: Dashboard (Control UI), browser

## Steps to Reproduce

1. Run a conversation until LCM compaction triggers (contextThreshold: 0.12)
2. Observe the compaction summary injected into the conversation
3. In the Dashboard, the summary appears as a regular assistant message — no "Compacted history" divider, no expander, no way to view original messages

## Expected Behavior

The compaction summary should be rendered with the Dashboard's "Compacted history" divider and expander, allowing users to click and view the original compressed messages.

## Additional Context

- All original messages are preserved in lcm.db (messages table) — the data is intact, only the UI rendering is affected
- This may be a regression from 0.13.0 where the marker behavior was introduced or changed. We briefly ran 0.13.0 (before rolling back due to #902) and the expander may have worked at that point
- Rolling back to 0.12.0 (due to #902) coincided with losing the Dashboard expander
- The allowModelOverride bug (openclaw/openclaw#94289) is a separate issue — this is purely about the compaction message marker

## Related Issues

- #902 — summary spend backoff (reason for rollback to 0.12.0)
- openclaw/openclaw#94289 — allowModelOverride not propagated (separate bug)

syltharion:
Closes #930

When LCM injects a compaction summary into the conversation context, the message lacked the `__openclaw` metadata that the Dashboard uses to render the 'Compacted history' divider with expand/collapse UI.

Adds `__openclaw: { kind: 'compaction' }` to the injected summary message so the Dashboard can detect and render it properly.

### DIFF
--- a/.changeset/mark-compacted-history.md
+++ b/.changeset/mark-compacted-history.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Mark assembled summary messages as OpenClaw compaction boundaries so the Dashboard renders its "Compacted history" UI instead of showing summaries as ordinary messages.
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ The command defaults to `${OPENCLAW_STATE_DIR:-~/.openclaw}` and `${OPENCLAW_STA
 - Node.js 22+
 - An LLM provider configured in OpenClaw (used for summarization)
 
-> **Compatibility:** `lossless-claw@0.10.0` and newer require OpenClaw `2026.5.12` or newer for OpenClaw's `api.runtime.llm.complete` summarization capability. Releases that include context-engine memory supplement support require OpenClaw `2026.5.28` or newer for `buildMemorySystemPromptAddition`. If you cannot upgrade OpenClaw yet, stay on `lossless-claw@0.9.4` and remove `0.10.x`-only config such as `sweepMaxDepth`.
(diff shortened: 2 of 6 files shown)


jalehman:
Thanks for the focused fix. The direction looks right for #930, but I do not think this is merge-ready yet.

Please tighten the contract before merge:

- declare the `__openclaw?: { kind: "compaction" }` shape on the relevant `AgentMessage` type or use a named subtype/helper instead of relying on `as AgentMessage`;
- add a focused engine-level regression test proving the marker survives `LcmContextEngine.assemble()` through the production assembly/sanitization path, not only the raw `ContextAssembler` output;
- add a patch changeset for the user-visible Dashboard rendering behavior;
- document or update OpenClaw compatibility metadata for the first stable host version that consumes `__openclaw.kind=compaction`; I found the UI marker path in OpenClaw stable `v2026.6.10`, while this package still advertises `>=2026.5.28`.

I would keep the implementation narrow, but make the host contract and release trail explicit.

PR #1036, opened 2026-07-25
fix: see through injected plugin context on decorated current turns

gorkem2020:
## Problem

Plugin context injected via `before_prompt_build` (blocks such as `<relevant-memories>`, `<derived-focus>`, `<inherited-rules>`) silently disappears from the outbound prompt on channel turns that carry the inbound metadata decoration (Slack and Telegram shapes). Webchat turns are unaffected.

On the model-facing runtime face, the injected blocks land between the inbound metadata prelude and the user body. That interposition defeats two mechanisms at once:

1. `extractBodyAfterOpenClawInboundMetadataBlock` reduces the runtime face to `injected blocks + body`, never just `body`, so `openClawInboundBodiesMatch` returns false and the decorated runtime copy fails to collapse onto its bare persisted row. The store double-writes, and the metadata-decorated face lands as the last assembled user row.
2. `liveContentContainsBareBody` then cannot see that decorated row as a line-aligned trailing segment of the live face (the injected blocks split the prelude from the body), so `assembledRowIsStructuralBareCurrentTurn` rejects it on both sub-gates and the live memory-bearing copy is never re-appended after assembly. The memory-stripped assembled face is what reaches the model.

A session reset does not help, because the drop is self-contained to the current turn: the live face loses to its own persisted bare row. The webchat shape survives because it has no metadata prelude for the blocks to interpose into, so the existing marker path recognizes it.

## Fix (two fail-closed layers)

- `openclaw-inbound-metadata.ts`: the inbound-body reduction now strips validated, complete leading injected-context tag blocks (exact known tag names only, exported as `OPENCLAW_INJECTED_CONTEXT_TAG_NAMES`) after the metadata blocks and again after the recap handling. An unclosed or unknown tag stays in the body and blocks the match. Because `openClawInboundBodiesMatch` rides this same reduction, the change also restores the persist-side same-turn collapse, preventing the double-write that produces the decorated-last row in the first place.
- `live-coverage.ts`: `assembledRowIsStructuralBareCurrentTurn` gains a reduced-face equality branch: both faces go through the metadata extraction and must reduce to the same non-empty body, still gated on the last assembled user row plus a recognized injected marker. This rescues conversations whose store already carries the decorated face as the current-turn row. `INJECTED_CONTEXT_MARKERS` now derives from the exported tag list, so the reduction and the recognition gate cannot disagree on which tags count as injected context.

## Third commit: the line-form recap matcher is now linear

Deploying the first two commits surfaced a latent performance bug that the reduction above newly exposes, so the fix ships with it.

`OPENCLAW_INBOUND_HISTORY_RECAP_LINE_BLOCK_RE` had the shape `(?:ENTRY)(?:\r?\n(?:ENTRY))*(?=terminator)`, where ENTRY itself carries three optional-prefix alternates and a `.+: .+` split that matches at any `": "`. Whenever a long entry run fails the trailing terminator check, the engine explores the cross-product of every per-line ambiguity: about fifteen colon-heavy entries already exceed ninety seconds of blocking in one synchronous call. Before this PR the injected blocks made the reduction bail out early, so live decorated traffic rarely reached the matcher; after it, ordinary group-channel turns do. On our deployment that produced Socket Mode disconnects within two minutes of load, with no event-loop warning to show for it (the liveness sampler could not run either).

The replacement walks the lines once and keeps the semantics:

- All-or-nothing terminator behavior is preserved. A shorter run can never satisfy the terminator that the maximal run failed, because the character after an intermediate entry is always its successor line, so greedy matching loses nothing.
- Entry recognition mirrors the old alternation exactly, with one deliberately conservative divergence: an entry whose content is a lone carriage return is now rejected rather than accepted (the old `.` matched `\r`). That direction only ever strips less.
- Worst case measured at 0.23ms on a 200-entry colon-heavy face, against a matcher that previously did not return within ninety seconds on fifteen.

This matcher also feeds identity canonicalization, so a semantic drift would change persisted `identity_hash` values with no repair path for rows already at the current version. Two independent reviewers diff-fuzzed the old regex against the new walker over 400,000 generated inputs each (both headers, all three entry anchor shapes, LF and CRLF, thirteen terminator shapes): identical match length on every realistic input, and every divergence was the documented trailing-lone-CR-at-EOF shape, always in the strip-nothing direction.

## Fail-closed properties

- Only a complete `<tag>...</tag>` block for the known names is stripped; unclosed or unknown tags stay in the body and block the match.
- The new recognizer branch uses full-body equality, not containment, and keeps the existing last-assembled-user-row and marker gates, so distinct turns are never collapsed.
- A face carrying injected blocks ahead of the metadata prelude stays unreduced, pinned by a boundary test.
- The recap walker strips nothing on a header with no entries, an entry run that peters out into unanchored prose, or an entry line with no sender/content split.

## Tests

19 new tests. `test/same-turn-body-collapse.test.ts` and `test/live-turn-structural-supersede.test.ts` cover the reduction and the append path plus fail-closed pins (unclosed tag, unknown tag, concealed different body, mid-body quoted tags, before-prelude boundary, distinct-body guard), written red-first against the pristine tip. `test/inbound-recap-line-matcher.test.ts` adds a bounded-time regression pin for the backtracking shape plus walker semantics pins (mixed anchor shapes, CRLF, end-of-content termination, unanchored petering out, internal blank line, missing sender split). Full suite 1924/1924, typecheck clean. Changeset included.

Verified live on a deployment running this branch: a decorated channel turn that previously reached the model with no injected blocks now carries all three in the final request's last user message, and the webchat shape is unchanged.

## Validation

- `npm run build` clean
- `npx tsc --noEmit` clean
- Focused Vitest: `test/batch-dedup.test.ts`, `test/same-turn-body-collapse.test.ts`, `test/live-turn-structural-supersede.test.ts`, and `test/inbound-recap-line-matcher.test.ts` all green; the new cells fail on the previous code
- Full suite green
- GitHub checks green on the current head
- Running in production since 2026-07-27 with the collapse behavior verified on live traffic

### DIFF
--- a/.changeset/decorated-turn-injected-context-collapse.md
+++ b/.changeset/decorated-turn-injected-context-collapse.md
@@ -1,0 +1,17 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+See through plugin-injected context blocks on decorated channel turns.
+
+Memory/context plugins prepend blocks like `<relevant-memories>` to the
+model-facing body via `before_prompt_build`. On decorated channels those blocks
+sit between the inbound metadata prelude and the user body, which defeated both
+the same-turn body collapse and the current-turn live-face recognition: the
+memory-bearing live copy was neither collapsed onto its bare persisted row nor
+re-appended after assembly, so the injected context silently vanished from the
+outbound prompt. The inbound-body reduction now strips validated, complete
+leading injected-context tag blocks (known tag names only, fail-closed), and
+the structural current-turn recognizer accepts a metadata-decorated assembled
+face whose extracted body equals the live copy's extracted body (the last
+assembled user row and recognized-marker gates are unchanged).
--- a/.changeset/decorated-turn-injected-context-collapse.md
+++ b/.changeset/decorated-turn-injected-context-collapse.md
@@ -12,11 +12,12 @@ prelude and the user body, which defeated both the same-turn body collapse and
 the current-turn live-face recognition: the
 memory-bearing live copy was neither collapsed onto its bare persisted row nor
 re-appended after assembly, so the injected context silently vanished from the
-outbound prompt. The inbound-body reduction now strips validated, complete
-leading injected-context tag blocks (known tag names only, fail-closed), and
-the structural current-turn recognizer accepts a metadata-decorated assembled
-face whose extracted body equals the live copy's extracted body (the last
-assembled user row and recognized-marker gates are unchanged).
+outbound prompt. Anchored covered-frontier alignment now strips validated,
+complete leading injected-context tag blocks (known tag names only), while
+unanchored transcript adoption keeps user-authored tags verbatim. The
+structural current-turn recognizer also accepts a metadata-decorated assembled
+face whose extracted body exactly equals the live copy's extracted body (the
+last assembled user row and recognized-marker gates are unchanged).
 
 The line-form history recap matcher is now a linear line walker instead of a
 composite backtracking regex. The old pattern went catastrophic (minutes of
--- a/.changeset/decorated-turn-injected-context-collapse.md
+++ b/.changeset/decorated-turn-injected-context-collapse.md
@@ -17,3 +17,11 @@ leading injected-context tag blocks (known tag names only, fail-closed), and
 the structural current-turn recognizer accepts a metadata-decorated assembled
 face whose extracted body equals the live copy's extracted body (the last
 assembled user row and recognized-marker gates are unchanged).
+
+The line-form history recap matcher is now a linear line walker instead of a
+composite backtracking regex. The old pattern went catastrophic (minutes of
+event-loop blocking per call) on entry runs that fail the trailing terminator
+check while containing per-line ambiguity, a shape real group-chat recaps
+produce and which the reduction above newly exposes to routine traffic.
+Semantics are unchanged and pinned by tests, including the all-or-nothing
+fail-closed rejection of an unterminated run.
--- a/.changeset/decorated-turn-injected-context-collapse.md
+++ b/.changeset/decorated-turn-injected-context-collapse.md
@@ -5,9 +5,11 @@
 See through plugin-injected context blocks on decorated channel turns.
 
 Memory/context plugins prepend blocks like `<relevant-memories>` to the
-model-facing body via `before_prompt_build`. On decorated channels those blocks
-sit between the inbound metadata prelude and the user body, which defeated both
-the same-turn body collapse and the current-turn live-face recognition: the
+model-facing body via `before_prompt_build`, at prompt-build time — strictly
+after the bare transcript row is persisted, so persisted rows never carry
+them. On decorated channels those blocks sit between the inbound metadata
+prelude and the user body, which defeated both the same-turn body collapse and
+the current-turn live-face recognition: the
 memory-bearing live copy was neither collapsed onto its bare persisted row nor
 re-appended after assembly, so the injected context silently vanished from the
 outbound prompt. The inbound-body reduction now strips validated, complete
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -24,7 +24,7 @@ import {
 } from "./message-content.js";
 import { messageIdentity } from "./message-signatures.js";
 import type { AgentMessage } from "./openclaw-bridge.js";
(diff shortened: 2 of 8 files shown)


gorkem2020:
Converting to draft for now: under sustained multi-agent load, our deployment shows an event-loop pressure regression with this change (Socket Mode disconnects and starved plugin timers within minutes), which is absent on the pristine tip at equal load in an A/B on the same host. Root-causing and optimizing; will push a revised version and mark ready again once it holds up under the same load.

**Update: root cause found, revised version pushed.** It was not the reduction itself but what the reduction newly exposes. `OPENCLAW_INBOUND_HISTORY_RECAP_LINE_BLOCK_RE` has the shape `(?:ENTRY)(?:\r?\n(?:ENTRY))*(?=terminator)`, where ENTRY carries three optional-prefix alternates and a `.+: .+` split that can match at any `": "`. When a long entry run fails the trailing terminator check, the engine explores the cross-product of every per-line ambiguity: about fifteen colon-heavy entries already exceed ninety seconds of blocking in one synchronous call. Multi-line chat messages produce exactly that terminator-failing shape in a real recap. Before this PR, injected context blocks made the body reduction bail out before the matcher ever ran, so live traffic rarely reached it; after it, ordinary decorated turns do.

The third commit replaces it with a single-pass line walker. Semantics preserved (all-or-nothing terminator behavior; one deliberately conservative divergence where entry content is a lone carriage return), worst case 0.23ms against a matcher that previously did not return within ninety seconds. Since this matcher also feeds identity canonicalization, two independent reviewers diff-fuzzed old against new over 400,000 generated inputs each: identical match lengths on every realistic input, all divergences in the strip-nothing direction, so persisted hashes are unaffected.

On our deployment the revised build has been clean since it went out: no Socket Mode errors, no starved timers, and no event-loop warnings at all, across multi-agent chat rounds and live human-driven group traffic, at several times the turn volume that broke the previous build within two minutes. Injected blocks arrive intact on decorated turns carrying recaps up to thirty lines. We also captured a real decorated face from a live request and replayed it through both builds: the previous matcher fails to collapse it (the original bug reproducing on real data), the revised one collapses it in about a millisecond, and on the same face with its recap run broken by a continuation line, which is what any multi-line chat message produces, the previous matcher does not return in bounded time while the revised one finishes in 0.12ms. Marking ready for review.

gorkem2020:
Thanks, all three landed usefully. Addressed in [e33a8f1](https://github.com/Martian-Engineering/lossless-claw/pull/1036/commits/e33a8f1), tests only.

**P2, the identity vs body-match asymmetry.** Agreed that it is an implicit invariant rather than a bug, and it is now pinned in `test/same-turn-body-collapse.test.ts`: body-match sees through a leading injected block, canonicalized identity content still carries that block verbatim, a decorated turn and its bare row hash differently by design, and two different injected payloads on the same body stay distinct in identity. The pin is deliberately two-sided, because either half moving alone is a defect: if identity began stripping, two live faces carrying different injected content would collide on one hash; if body-match stopped, the decorated turn would no longer collapse onto its persisted row at all. Rows already written at the current identity version have no repair path, which is why we would rather pin the asymmetry than have a later change quietly "align" the two reducers.

**P3, the loose linearity ceiling.** Fair point that 2s only catches the original catastrophic case. The ceiling is now 250ms, and a second case grows the entry run 64x (2560 entries) under the same bound, so the test can actually separate linear from quadratic instead of relying on one absolute number. Measured on this head: 40 entries 0.10ms, 160 entries 0.17ms, 640 entries 0.46ms, 2560 entries 0.98ms, so the walker visits each line once. A quadratic reintroduction from that 0.1ms base would land in the hundreds of milliseconds at 2560 entries and fail; a catastrophic one would not return at all. The remaining headroom is for loaded CI runners rather than for the implementation.

**P3, the lone trailing CR.** Acknowledged and deliberately unchanged. Adding `rest === "\r"` to the terminator set would be a semantic change to this matcher, and the matcher also feeds identity canonicalization, so widening what it strips silently rewrites `identity_hash` for content already persisted at the current version, with no repair path for those rows. Weighed against a line ending we have never observed a host emit, and a failure mode that is degraded rather than lossy (the recap text stays in the body, nothing is dropped), the conservative rejection is the better trade. If a CR-only recap ever turns out to be a real emission shape, the right change is a versioned canonicalization step with a migration, not an inline widening.

jalehman:

PR #1039, opened 2026-07-27
fix: guard sqlite: sessionFile paths against filesystem stat operations

ralf003:
## Problem

OpenClaw 7.2+ stores sessions in SQLite and passes sessionFile values with a `sqlite:` prefix (e.g. `sqlite:linus:c0e057d2:path`). Before this fix, the lossless-claw plugin passed these values through to `fs.stat`, `fs.createReadStream`, and `fs.open` in three paths:

1. **auto-rotate** - stat fails with ENOENT, every turn logs `session-file-stat-failed`
2. **transcript reconcile** - `session file missing; skipping transcript reconcile`, checkpoint offset stays 0 forever
3. **bootstrap checkpoint refresh** - ENOENT on every turn, leading to full re-reads instead of incremental updates

On Windows, the stat fails with the path resolved against `C:\Windows\system32\sqlite:...`, causing 532 warn lines and 196 error lines in a single day of logs.

## Solution

This patch introduces `isSqliteSessionFile()` in value-utils and applies it in every call site that previously performed filesystem I/O on session files.

| File | Change |
|------|--------|
| `value-utils.ts` | + `isSqliteSessionFile()` + `normalizeSessionFilePathForComparison` no longer resolve()s sqlite: paths |
| `engine.ts` | bootstrap uses 0-size placeholder stat for sqlite sessions |
| `transcript.ts` | 5 I/O functions return early for sqlite paths |
| `session-rotation.ts` | auto-rotate skips sqlite sessions entirely |
| `session-rollover.ts` | rotation/ambiguity checks skip sqlite-tracked transcripts |
| `transcript-reconciler.ts` | stat wrapped in `!isSqliteSessionFile` guard; refreshBootstrapState uses placeholder |

**6 files, +78 -16**

### Before (production logs, OpenClaw 2026.7.2-beta.4 + LCM 0.14.0)

Each agent turn produces 3 warn lines:
```
[lcm] afterTurn: session file missing; skipping transcript reconcile
  reason=append-only-ineligible sessionFile=sqlite:linus:c0e057d2:...

[lcm] afterTurn: bootstrap checkpoint refresh failed
  ENOENT: stat 'C:\Windows\system32\sqlite:linus:c0e057d2:...'

[lcm] auto-rotate: session-file-stat-failed
  error=ENOENT stat 'C:\Windows\system32\sqlite:linus:c0e057d2:...'
```

### After (ECS smoke test)

Patched dist loaded on OpenClaw 2026.7.2-beta.4, manual verification:
- `isSqliteSessionFile` present in minified dist (4 occurrences)
- Module loads without errors
- No ENOENT / stat-fail / checkpoint-refresh-fail in plugin initialization

### Build & lint

- `npm run build` - passes
- `npx oxlint` - 0 errors (127 pre-existing warnings unchanged)

## Changes

- `src/value-utils.ts`: add `isSqliteSessionFile`, adapt `normalizeSessionFilePathForComparison`
- `src/engine.ts`: replace `stat()` with placeholder for sqlite sessions
- `src/transcript.ts`: add guards to `readLeafPathMessages`, `readLeafPathRawEntries`, `readSessionParentSessionReference`, `readFileSegment`, `readLastJsonlEntryBeforeOffset`
- `src/session-rotation.ts`: skip rotation for sqlite sessions in `maybeAutoRotateManagedSessionFile`, `prepareStartupAutoRotateCandidate`, `rewriteTranscriptForRotate`
- `src/session-rollover.ts`: skip stat checks for sqlite-tracked transcripts
- `src/transcript-reconciler.ts`: wrap stat in guard; placeholder stat for sqlite in `refreshBootstrapState`

### DIFF
--- a/.changeset/sqlite-sessionfile-guard.md
+++ b/.changeset/sqlite-sessionfile-guard.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Guard SQLite-backed session file paths (`sqlite:`) against filesystem stat/read/write operations in auto-rotate, rollover detection, transcript reconciliation, and bootstrap checkpoint refresh. OpenClaw 7.2+ sessions stored in SQLite use `sqlite:` prefixed paths that were passed through to `fs.stat`, `fs.createReadStream`, and `fs.open`, causing ENOENT errors on every agent turn. SQLite sessions now skip file-based rotation, rollover stat checks, and transcript I/O.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -90,7 +90,7 @@ import { canonicalizeOpenClawInboundMetadataIdentityContent } from "./openclaw-i
 import { PROMPT_RECALL_MAX_MESSAGES, PROMPT_RECALL_SEARCH_CANDIDATE_LIMIT, buildPromptRecallProjectionFingerprint, extractPromptRecallIdentifiers, extractPromptRecallSnippet, findPromptRecallIdentifierIndex, isPromptRecallEligibleRole, normalizePromptRecallCoverageText, normalizePromptRecallText, renderPromptRecallMessage } from "./prompt-recall.js";
 import { listTranscriptToolResultEntryIdsByCallId } from "./replay-metadata.js";
 import { extractRuntimePromptTokenCount } from "./token-accounting.js";
-import { asRecord, formatDurationMs, resolvePositiveInteger } from "./value-utils.js";
+import { asRecord, formatDurationMs, isSqliteSessionFile, resolvePositiveInteger } from "./value-utils.js";
 
 type AgentMessage = Parameters<ContextEngine["ingest"]>[0]["message"];
 const LOSSLESS_AGENT_RUN_REQUIRED_HOST_CAPABILITIES: ContextEngineHostCapability[] = [
@@ -1859,7 +1859,8 @@ export class LcmContextEngine implements ContextEngine {
     this.ensureMigrated();
     const startedAt = Date.now();
     const sessionLabel = formatSessionLabel(params.sessionId, params.sessionKey);
-    const sessionFileStats = await stat(params.sessionFile);
+    const sqliteSession = isSqliteSessionFile(params.sessionFile);
+    const sessionFileStats = sqliteSession ? { size: 0, mtimeMs: 0 } : await stat(params.sessionFile);
     const sessionFileSize = sessionFileStats.size;
     const sessionFileMtimeMs = Math.trunc(sessionFileStats.mtimeMs);
     const parentSessionReference = await readSessionParentSessionReference(params.sessionFile);
--- a/src/session-rollover.ts
+++ b/src/session-rollover.ts
@@ -16,7 +16,7 @@ import type { AgentMessage } from "./openclaw-bridge.js";
 import { isIsolatedCronSessionKey } from "./session-patterns.js";
 import type { ArchiveCause, ConversationStore } from "./store/conversation-store.js";
 import { getTranscriptEntryMeta } from "./transcript.js";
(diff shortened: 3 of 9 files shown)


ralf003:
Thanks for the review @evaos-code-review-bot. All 4 findings addressed in 7f428847.

jalehman:
Thanks for this. The direction is useful as a narrow hotfix for current `main`, but I think it needs one more pass before review/merge.

Two concrete asks:

1. Guard `importInitialAfterTurnTranscript` too. It still calls `stat(params.sessionFile)` on first-contact afterTurn before `readLeafPathMessages(...)`; the error is swallowed, but it is still a filesystem stat against `sqlite:`.

2. Add focused regression coverage for the new boundary: `isSqliteSessionFile`, `normalizeSessionFilePathForComparison("sqlite:...")`, and at least the transcript-reader/initial-afterTurn no-op path for `sqlite:` markers.

Please also either canonicalize the predicate consistently with comparison normalization, or point to the OpenClaw contract that guarantees lowercase trimmed `sqlite:` markers.

SweetSophia:
Confirmed that this needs to cover more than the `sqlite:` prefix.

On OpenClaw `2026.7.2-beta.6` (`4d6bdbd`) with Lossless Claw `0.15.0`, the context-engine runtime now supplies a canonical session-key locator in `sessionFile`, for example:

```text
agent:main:telegram:direct:<redacted-chat-id>
```

Lossless Claw treats that value as a filesystem path and calls `stat()` on it. OpenClaw then quarantines the context engine for the process:

```text
[context-engine] Context engine "lossless-claw" owner=plugin:lossless-claw failed during bootstrap:
ENOENT: no such file or directory, stat 'agent:main:telegram:direct:<redacted-chat-id>';
quarantining it for this process and falling back to default engine "legacy".
```

The earlier beta.4 form was `sqlite:<agent>:<session-id>:<path>`, so the `isSqliteSessionFile()` work in this PR covers that shape but not beta.6's `agent:` locator. The same positive-filesystem assumption exists across the bootstrap checkpoint, transcript reconciliation, and rotation paths covered by this PR.

Could the guard be generalized to a session-locator predicate rather than a `sqlite:`-only predicate? At minimum it should treat both `sqlite:` and `agent:` values as non-filesystem locators. A positive path classification would be safer long term, but it must preserve Windows drive-letter paths such as `C:\\...`.

Suggested regression cases:

```text
sqlite:main:<session-id>:/home/user/.openclaw/agents/main/sessions/sessions.json
agent:main:telegram:direct:<chat-id>
agent:main:<session-id>
```

Our local compatibility guard for both locator families prevents bootstrap quarantine on beta.6. This appears to be the same contract transition addressed by this PR, with one additional input shape.

PollyBot13:
Thanks @ralf003 — I checked this against the current release-line split before trying to revive the patch.

`next/1.0` already contains the SQLite transcript-projection architecture from #952, and its bootstrap/reconcile path reads OpenClaw's projection instead of treating `sessionFile` as a filesystem path. Meanwhile `main` is intentionally the 0.x/file-backed line and now excludes SQLite-backed OpenClaw through its peer range. That makes this nine-file guard valuable historical diagnosis, but not the patch we should carry into the upcoming 1.0 beta.

The remaining reproducible `next/1.0` compatibility gap is gateway lifecycle handling: the fixes from #1073 and #1083 are present on `main` but absent from `next/1.0`.

If you have a current `next/1.0` reproduction showing a locator still reaches filesystem I/O, please share it and I will incorporate that evidence. Given the Tuesday release window, I did not block the lifecycle port on a response.

Update: the focused fallback is now #1089. It credits this investigation, leaves #1039 unchanged for maintainers to close or repurpose, and includes an exact OpenClaw `2026.8.1-beta.3` same-PID restart canary. No response is required for #1089 to proceed; a current locator reproduction remains welcome as separate evidence.

Issue #942, opened 2026-06-28
Deferred-compaction drain honors a persisted context_threshold over live config; a stale lowered threshold becomes immortal and wedges active conversations


gorkem2020:
## Summary

The deferred-compaction **drain** path resolves its `contextThreshold` from the persisted `conversation_compaction_maintenance` row (`persistedContextThresholdOverride(maintenance) ?? resolver.resolve(...)`) instead of the live config. If a debt row was recorded while `contextThreshold` was temporarily low (for example an operator lowered it for testing, then reverted), that stale low value is rehydrated on every drain and honoured over the current config.

It is effectively immortal: it survives a gateway restart (it lives in `lcm.db`), and reverting the config does not clear it (the drain bypasses config). On an **active** conversation whose live runtime context exceeds the stale low target, every compaction then ends `ok=false reason=stored_compaction_exhausted_but_live_context_still_exceeds_target` (the "transcript wedge"), because the target is unreachable.

## Observation

- Live config `plugins.entries.lossless-claw.config.contextThreshold = 0.75`; startup banner logs `threshold=0.75`.
- Earlier the threshold had briefly been `0.02` (and `0.42`) for testing. Deferred after-turn checks recorded `conversation_compaction_maintenance` rows with `context_threshold=0.02, context_threshold_source='global', reason='threshold'`.
- After reverting config to `0.75` and restarting, an active conversation still compacts toward `0.02 * tokenBudget` (about 20k): `compact: decision ... targetTokens = 0.02 * tokenBudget`, `shouldCompact=true` firing far below the `0.75` trigger, then `compact: done ok=false reason=stored_compaction_exhausted...`. The stored side compacts fine (down to ~1k); the live runtime overhead alone exceeds the stale target, so compaction can never succeed.

## Root cause

- The drain resolves the threshold as `persistedContextThresholdOverride(maintenance) ?? this.contextThresholdResolver.resolve(...)`. The `??` short-circuits on the persisted value, so the live config is never consulted on a debt drain.
- `persistedContextThresholdOverride` returns the row's `context_threshold` whenever it is a finite number, with no comparison to the current global config. (`'global'` here is just the resolver's no-override-matched source label, not a separate knob.)
[...]

PR #1041, opened 2026-07-27
fix(maintenance): supersede stale persisted context thresholds against live config

gorkem2020:
## Problem

[#942](https://github.com/Martian-Engineering/lossless-claw/issues/942): the deferred-compaction drain resolves its context threshold from the value persisted on the `conversation_compaction_maintenance` row and prefers it unconditionally over live config. A threshold recorded during a since-reverted config experiment therefore outlives the revert: the row survives restarts, config edits, and `/reset`, and no doctor cleaner targets it.

Observed twice on a production fleet: a `0.02` threshold recorded during a short-lived low-threshold experiment survived the config revert to `0.75`. Every subsequent drain targeted roughly 20k tokens, below the session's roughly 24k irreducible runtime overhead, so compaction could never reach the target and aborted with `stored_compaction_exhausted_but_live_context_still_exceeds_target` on every attempt until the conversation was manually reset.

## Fix

The persistence exists for a good reason (per the existing code comment, a background drain may lack the runtime model metadata that originally selected an override, and re-resolving could silently flip the compaction decision), so this change does not simply prefer live config. `reconcilePersistedContextThreshold` draws the trust boundary instead:

- A persisted `override` threshold stays authoritative while any configured rule could still match this context with absent runtime metadata treated as satisfiable (`couldAnyRuleMatch`, a wildcard variant of the rule matcher; a known session key remains decisive). When no configured rule could possibly produce it, the persisted value is provably stale and the live resolution wins.
- A persisted `global` threshold needs no runtime metadata, so live config is authoritative: it is superseded when the configured global diverges or a live rule now matches this context. An equal persisted global is kept unchanged.
- Every supersession logs an `[lcm] maintain:` info line carrying both the persisted and live values.

Existing behavior is otherwise preserved, including the fresh-tail and leaf-chunk fallback fills for older rows written before those columns existed.

## Tests

- Two engine integration regressions, red-proofed against the previous behavior (both fail without the fix): a stale persisted override with no remaining configured rules, and a stale persisted global after a config revert, both now drain at the live `0.75`.
- Unit coverage for the `couldAnyRuleMatch` wildcard semantics (absent window or model metadata counts as satisfiable, a known session key is decisive) and the full reconcile matrix (override kept while a rule is still plausible, equal global kept, divergent global and live-override supersessions).
- Full suite green: 107 files, 1917 tests; typecheck and build clean.

## Notes

- No changeset included; flagging that for a maintainer call. The change affects drain behavior only for rows whose persisted threshold provably cannot originate from current config.

### DIFF
--- a/.changeset/supersede-stale-context-threshold-debt.md
+++ b/.changeset/supersede-stale-context-threshold-debt.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Deferred maintenance drains no longer honour a persisted context-threshold override that current config can no longer produce: the persisted value is kept only while a plausibly-matching override rule still carries the same threshold and recorded sizing, and a persisted global threshold that diverges from the configured global is superseded, so stale rows from removed or reverted threshold experiments cannot wedge compaction.
--- a/src/context-threshold.ts
+++ b/src/context-threshold.ts
@@ -186,6 +186,34 @@ function rulePayloadProduces(
   return true;
 }
 
+function resolvedPayloadMatchesPersisted(
+  resolved: Pick<
+    ResolvedContextThreshold,
+    "contextThreshold" | "freshTailCount" | "leafChunkTokens"
+  >,
+  persisted: Pick<
+    ResolvedContextThreshold,
+    "contextThreshold" | "freshTailCount" | "leafChunkTokens"
+  >,
+): boolean {
+  if (resolved.contextThreshold !== persisted.contextThreshold) {
+    return false;
+  }
+  if (
+    persisted.freshTailCount !== undefined &&
+    resolved.freshTailCount !== persisted.freshTailCount
+  ) {
+    return false;
+  }
+  if (
+    persisted.leafChunkTokens !== undefined &&
+    resolved.leafChunkTokens !== persisted.leafChunkTokens
+  ) {
+    return false;
+  }
+  return true;
+}
+
 // Summarize which matchers selected the winning rule for log lines.
 function describeRuleMatch(
   rule: ContextThresholdOverride,
@@ -266,7 +294,10 @@ export function reconcilePersistedContextThreshold(params: {
     return { resolved: live, supersededStalePersisted: false };
   }
   if (persisted.source === "override") {
-    if (params.anyRuleCouldProducePersisted) {
+    if (
+      params.anyRuleCouldProducePersisted &&
+      (live.source !== "override" || resolvedPayloadMatchesPersisted(live, persisted))
+    ) {
       return { resolved: persisted, supersededStalePersisted: false };
     }
     return { resolved: live, supersededStalePersisted: true };
--- a/src/context-threshold.ts
+++ b/src/context-threshold.ts
@@ -107,6 +107,54 @@ function ruleMatches(params: {
   return true;
 }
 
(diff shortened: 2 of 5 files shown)


jalehman:
Thanks for the detailed writeup and tests. This is the right area and #942 is a real problem, but I think the stale-override trust boundary is still too broad.

`reconcilePersistedContextThreshold` keeps any persisted override when `couldAnyRuleMatch` is true. That only proves some current override rule might match with missing metadata; it does not prove current config could produce the persisted threshold/fresh-tail/leaf settings. If the old low-threshold override was removed but an unrelated override rule remains plausible, the stale persisted value can still be honored.

Could you tighten this so persisted override debt is kept only when a current plausible rule can produce the persisted threshold/config payload, and add a regression for an unrelated remaining override rule with a different threshold?

Also flagging the release-note gate for later: this changes user-visible maintenance behavior, so we should decide on a patch changeset before merge.

gorkem2020:
Agreed, rule presence was too coarse a proxy for that trust boundary. Pushed [d161f86](https://github.com/gorkem2020/lossless-claw/commit/d161f86afbce6888a699944c047b9e6f44f2300e) tightening it:

- `couldAnyRuleMatch` is now `couldAnyRuleProduce`: a persisted override survives reconciliation only if some rule both plausibly matches (absent runtime metadata still counts as satisfiable) and carries the persisted payload, meaning an exact threshold match plus equality on every sizing field the row recorded. Matching and producing must be satisfied by the same rule, not one rule each.
- Sizing fields the row never recorded (rows written before those columns existed) do not count against a rule, since an absent value there means "not recorded" rather than "recorded as absent"; the existing runtime fill-in covers those rows on the kept path.
- Added the regression you asked for: with an unrelated surviving override rule at a different threshold, the removed rule's persisted 0.02 is superseded and the drain runs at the configured global. Verified failing against the previous head before the fix, plus unit cells for the payload boundary (threshold mismatch, recorded-sizing mismatch, legacy rows, same-rule requirement).
- Added a patch changeset (`supersede-stale-context-threshold-debt`) for the release-note gate.

Typecheck, build, and the full suite are green on the new head.

gorkem2020:
Review round handled in [a8c78a8](https://github.com/gorkem2020/lossless-claw/commit/a8c78a83db5964e4602f38468544c05348e3e598), all three as proof/documentation work (no runtime change needed):

- Equal-threshold sub-case pinned: a reconcile unit cell asserts a live override at the same numeric threshold as the persisted global wins and its payload (freshTailCount/leafChunkTokens) is adopted from the override.
- Engine-level fill-in coverage: new cell where a legacy debt row (threshold only, no recorded sizing) is kept because the producing rule survives, and the unrecorded sizing fills from the live resolution; the persisted reason string proves the kept object is the persisted one. One correction to the finding's premise: the positive keep-path with fully recorded sizing was already engine-covered by the pre-existing "consumes deferred threshold debt when the host opts in" cell, so the gap was specifically the fill-in merge, which the new cell reaches.
- The unrelated-rule regression now also asserts the supersession info line including the persistedThreshold and liveThreshold values.
- Added a comment documenting that the global staleness check is intentionally scalar-only and must grow a payload comparison if global-level sizing ever becomes configurable.

Typecheck and the full suite are green on the new head.

jalehman:

PR #1042, opened 2026-07-28
fix(dedup): anchor metadata-face collapse on transcript provenance

gorkem2020:
## What

Restores the same-turn bare+decorated collapse on metadata-decorated lanes, with a stronger and non-forgeable anchoring rule: a metadata-body match may collapse the decorated runtime copy onto its bare persisted row **if and only if that persisted row carries a `transcript_entry_id`**, meaning it was written by the host's own transcript flush. A user can forge a metadata block; they cannot forge a row the transcript reconcile imported. Provenance-less matches stay weak and the pair deliberately duplicates (duplicates over deletion).

Also unglues the history-recap strip from the `history_count` metadata field: live telegram inbound carries a recap block while its metadata record has no `history_count` at all, so announcement-gating left the recap fused to the body and the reduction could never match. Structural validation (the existing header, fence, and line-grammar validators) is the real gate; the record field is at most a hint.

## Why now

On a multi-agent production install, every inbound message on metadata-decorated lanes currently persists twice (bare row with `transcript_entry_id` + decorated row without), and both faces replay into every later prompt. Store evidence from one conversation: 24 bare user rows and 23 decorated twins. The three contributing changes each made sense in isolation:

- [d163ed6] put the body-equality match behind `allowUntimestampedInboundBodyMatch`, wired at only the covered-frontier site, so the oversized/degraded/suffix routes silently lost it.
- The alignment-strength rework later classified the surviving covered-route match as `"unanchored-inbound"`, which `isStrongReplayAnchor` excludes, so a lone decorated copy never trimmed even where the match fired.
- [421e7cb]'s reduction gated the recap strip on `history_count`, which the host omits on some channels.

The forgery concern behind the original tightening is real, and this PR keeps it: the new gate is strictly stronger than "no collapse ever" for provenance-less rows, and strictly more correct than "no collapse ever" for transcript-proven rows, which is the trusted marker the earlier comment said was missing.

## Implementation

- `MessageRecord` now carries `transcriptEntryId` (projected by the message SELECTs; the column and unique index already existed).
- New match kind `"provenanced-inbound"`; `isStrongReplayAnchor` accepts it. The covered-frontier site labels a body match by the persisted row's provenance.
- The oversized-boundary, degraded, and suffix-coverage sites now pass the space-run and body-match options gated on the persisted row's provenance, so heuristic routes can collapse a transcript-proven pair and still refuse a provenance-less one.
- The recap strip in `extractBodyAfterOpenClawInboundMetadataBlock` runs on structural validation alone.

## Tests

Red-proofed: with the source changes stashed, 5 cells fail (the three route cells plus the two reduction cells); green with them.

- New `test/metadata-face-double-write.test.ts`: covered, degraded, and oversized-suffix routes each collapse a transcript-proven pair to exactly one row (the bare face always survives); a decorated frame concealing a different body never collapses; a provenance-less pair still duplicates.
- `test/same-turn-body-collapse.test.ts`: the "no history announced" cell is re-derived to the live host shape (recap present, `history_count` absent, strip succeeds); a new cell pins the line-grammar recap face; the concealment, quoted-header, and malformed-recap neighbours stay fail-closed and unchanged.
- `test/transcript-entry-id.test.ts`: the "keep a lone conv-info copy" cell is re-derived: with a transcript-proven bare row the copy now collapses without a sibling anchor; the provenance-less descendant of the old premise lives in the new test file.

Full suite green (108 files, 1911 tests), typecheck clean.

### DIFF
--- a/.changeset/transcript-provenance-dedup.md
+++ b/.changeset/transcript-provenance-dedup.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Collapse metadata-decorated inbound copies onto transcript-proven bare rows across covered, degraded, and oversized replay paths, including host history recaps whose metadata omits `history_count`.
--- a/.changeset/transcript-provenance-dedup.md
+++ b/.changeset/transcript-provenance-dedup.md
@@ -2,4 +2,4 @@
 "@martian-engineering/lossless-claw": patch
 ---
 
-Collapse metadata-decorated inbound copies onto transcript-proven bare rows across covered, degraded, and oversized replay paths, including host history recaps whose metadata omits `history_count`.
+Require transcript provenance before untimestamped metadata-body matches can support covered-frontier replay alignment, while keeping unannounced recaps and heuristic matches fail-closed.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -43,17 +43,10 @@ type StoredIncomingMatch =
   | "unproven-externalized"
   | "redacted";
 type CoveredStoredIncomingMatch = StoredIncomingMatch | "decorated";
-type AlignmentMatch =
-  | CoveredStoredIncomingMatch
-  | "unanchored-inbound"
-  | "provenanced-inbound";
+type AlignmentMatch = CoveredStoredIncomingMatch | "unanchored-inbound";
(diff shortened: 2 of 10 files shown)


jalehman:

PR #1043, opened 2026-07-28
fix(reconcile): adopt oldest match for repeated same-body recovery turns

gorkem2020:
## Problem

`adoptDecorationInvariantEntryId` fails closed when more than one recently persisted, still unstamped decorated user row reduces to the same bare body:

```ts
if (matches.length !== 1) {
  return false;
}
```

When a user sends the same short reply twice (two "ok" turns) and a restart-class recovery (placeholder or checkpoint-missing) later replays the transcript, adoption refuses both candidates, and each bare transcript copy is imported as an additional persisted user row. The store then carries a duplicate of a message it already holds, replayed into context on every subsequent turn. Two persisted rows are outside the runtime-vs-persisted metadata-face collapse, so nothing downstream removes them.

The multiplicity refusal was introduced in [d163ed6's sibling hardening round, commit d897771](https://github.com/Martian-Engineering/lossless-claw/commit/d897771) on [#935](https://github.com/Martian-Engineering/lossless-claw/pull/935).

## Why adopting the oldest match is safe

The refusal protects against stamping the wrong row, but the candidates here are byte-identical after reduction: `openClawInboundBodiesMatch` already guarantees every match reduces to the exact bare body. Interchangeable bodies are interchangeable provenance carriers, so there is no wrong choice, only a deterministic one. Rows whose reduced body differs are untouched (the filter is unchanged).

Adopting the oldest match also converges under repetition: each restamp removes that row from the next call's unstamped candidate set, so N identical turns recovered from a transcript holding N bare entries end with N stamped rows and zero imports.

The previous behavior traded a non-existent mis-stamp risk for a guaranteed persisted duplicate, which is the worse failure for a store whose contract is no duplicates.

## Tests

`test/placeholder-recovery-double-write.test.ts`:

- Re-derived the multiplicity cell: two same-body decorated rows plus one bare transcript entry now end as two rows with the oldest stamped, instead of the previously pinned three-row outcome (the bare duplicate import asserted as expected).
- New cell: two identical turns recovered from a transcript carrying both bare entries end with both rows stamped, zero bare imports, and both assistant replies imported (no over-suppression).

Both cells fail on the previous code with the predicted duplicate imports. Full suite: 1906 passing.

No changeset included (external contribution); flagging for a maintainer to add one if this qualifies as a user-facing patch.

## Validation

- `npm run build` clean
- `npx tsc --noEmit` clean
- Focused Vitest: `test/placeholder-recovery-double-write.test.ts` passes 7/7; the two new cells fail on the previous code with the predicted duplicate imports and pass with the fix
- Full suite green (1906 tests, 107 files)
- GitHub checks green on the current head

### DIFF
--- a/.changeset/repeated-same-body-recovery.md
+++ b/.changeset/repeated-same-body-recovery.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Prevent repeated same-body recovery turns from replaying duplicate bare transcript rows by adopting the oldest matching decorated runtime row during transcript reconciliation.
--- a/src/transcript-reconciler.ts
+++ b/src/transcript-reconciler.ts
@@ -1219,10 +1219,15 @@ export class TranscriptReconciler {
    * Stamp a brand-new transcript entry id onto a recently-persisted,
    * still-unstamped user runtime row that is the DECORATED face of the same
    * turn (its recognized metadata wrapper reduces to the exact bare
-   * model-facing body). When that body matches exactly one candidate, keeps the
-   * decorated row and anchors it instead of importing the bare transcript copy
-   * as a duplicate. Bounded to the flush-lag tail window and fails closed on
-   * ambiguous repeated bodies so legacy or unrelated rows are never mis-adopted.
+   * model-facing body). Keeps the decorated row and anchors it instead of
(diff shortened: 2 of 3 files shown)


jalehman:

PR #1044, opened 2026-07-28
fix(stable-event-key): remove timestamp-only fallback key per PR #1044 review

bowenluo718:
## Summary

Addresses PR #1044 review comments by **removing the timestamp-only fallback key entirely**.

When OpenClaw is configured with `logging.redactPatterns`, the transcript file contains **redacted** message content while the live `afterTurn` batch carries **original** content. Because `identity_hash = SHA256(role + "\0" + content)`, the two representations produce different hashes, causing the same semantic event to be ingested twice.

This PR adds a **stable event key** to every `messages` row and short-circuits ingest when the key already exists for the conversation.

**Per PR review feedback**: The original implementation included a timestamp-based fallback key for messages without responseId or toolCallId. This fallback was removed because it could incorrectly deduplicate distinct messages that happen to share the same role and millisecond timestamp.

## Changes

| Area | What |
|------|------|
| `src/stable-event-key.ts` | New pure helper extracting stable identity from any `AgentMessage`. **Timestamp fallback removed — returns null for messages without responseId/toolCallId.** |
| `src/db/migration.ts` | `messages.stable_event_key` column + partial unique index |
| `src/store/conversation-store.ts` | `hasMessageByStableEventKey`, `filterExistingStableEventKeys` |
| `src/engine.ts` | Short-circuit in `ingestSingle` before any side effects |
| `src/batch-dedup.ts` | Final-pass filter on runtime batch |
| `test/` | Unit tests for the helper, e2e tests for responseId/toolCallId scenarios, **removed timestamp fallback tests** |

## Design

Stable event key priority:
1. `assistant` → `responseId` (or `response_id`)
2. `tool` / `toolResult` → `toolCallId` (via pairing helper)
3. **No fallback — returns `null` (falls back to existing identity_hash-based dedup and messagesDifferOnlyByHostRedaction)**

The partial unique index on `(conversation_id, stable_event_key) WHERE stable_event_key IS NOT NULL` is the final defense; existing identity mechanisms (`transcript_entry_id`, `identity_hash`, `messagesDifferOnlyByHostRedaction`) are preserved.

## Verification

- `npm run typecheck` — clean
- `npm test` — 1915/1915 tests pass (108 files)
- `npm run build` — all three sub-builds green
- Design spec and implementation plan in `docs/superpowers/`

## Review Addressed

From PR #1044 review:
- ✅ P1 (Data loss): Fixed by removing timestamp-only fallback that could collapse distinct messages
- ✅ P2 (Runtime correctness): Fixed by removing redundant normalizedRole duplication  
- ✅ P3 (CI/build): Added missing trailing newline
- ✅ Added regression test coverage for equal-timestamp distinct-content scenario (now returns null instead of incorrect key)


jalehman:
Thanks for the PR. The duplicate-ingestion bug is real, and the `responseId` / `toolCallId` direction is plausible, but I do not think this branch is safe to merge as written.

The `stable_event_key` timestamp fallback is too broad: it derives identity from same conversation + role + millisecond timestamp, then both the ingest short-circuit and the partial unique index treat that as definitive duplicate proof without comparing content. That can silently drop two distinct same-role messages that share the same source timestamp, which violates Lossless’ data-preservation invariant.

The current tests do not cover that failure mode. The “distinct timestamp” test only proves a 1ms difference survives, and an existing same-second regression was changed so the second turn no longer shares the timestamp.

Please narrow this before review:

- remove the timestamp-only stable-key fallback, or add a content discriminator that preserves the redaction case without collapsing distinct content;
- add equal-timestamp / distinct-content regression coverage;
- keep the fix as close as possible to the existing `batch-dedup` / transcript-reconcile ownership unless a schema-level identity column is still necessary after that narrowing.

jalehman:
Thanks for the contribution, @bowenluo718. This is landed now.

PR #1046, opened 2026-07-29
fix(engine): declare accepted OpenClaw host parameters

steipete:
## Why

I recently removed OpenClaw's LosslessClaw-specific validator-error retry shim in [openclaw/openclaw#115872](https://github.com/openclaw/openclaw/pull/115872). The replacement contract lets context engines declare the host-added lifecycle parameters they accept through `info.acceptedHostParams`.

Engines without that declaration receive the legacy pre-`sessionKey` parameter set through 2026-08-12. After that compatibility window, undeclared engines receive the full host parameter set. Declaring LosslessClaw's contract now keeps its behavior explicit and independent of that cutoff.

## What changed

- Declare `sessionKey`, `prompt`, and `runtimeContext` in `info.acceptedHostParams`. I verified this list against the engine's lifecycle parameter shapes; LosslessClaw does not consume `runtimeSettings` or `sessionTarget`.
- Report the actual package version in `info.version` instead of the stale `0.1.0` literal. This field is engine-version metadata, not a protocol version.
- Add focused regression coverage and a patch changeset.

The lifecycle boundary has no strict runtime object validator and already ignores additional host-side properties, so no validation relaxation was necessary.

## Proof

- `npm test` — 109 files, 1,964 tests passed
- `npm run typecheck`
- `npm run build`

## Related

OpenClaw-side removal of the legacy host-param default, gated on this PR: https://github.com/openclaw/openclaw/pull/115948

### DIFF
--- a/.changeset/accepted-host-params.md
+++ b/.changeset/accepted-host-params.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Declare the OpenClaw host parameters accepted by the context engine and report the package version in engine metadata.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -2,6 +2,7 @@ import { createHash } from "node:crypto";
 import { open, stat } from "node:fs/promises";
 import { join } from "node:path";
 import type { DatabaseSync } from "node:sqlite";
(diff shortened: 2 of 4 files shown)


jalehman:
thanks pete

Issue #992, opened 2026-07-11
Regression: compacted projection triggers false missing-tool-result repair despite matching canonical results


100yenadmin:
## Summary

On Lossless Claw `0.13.2` with OpenClaw `2026.7.1-beta.5`, compacted/replayed history can omit or misorder matching real tool-result rows even though the canonical raw session transcript already contains them. OpenClaw's generic transcript repair then inserts false `missing tool result in session history` errors for those call ids.

This is narrower than a tool execution or persistence failure: the real results exist in canonical storage, but the Lossless compacted projection presented the calls as missing and polluted later recall with synthetic errors.

## Environment

- OpenClaw: `2026.7.1-beta.5 (b6387af)`
- Active context engine: Lossless Claw `0.13.2`
- Surface: webchat / Control UI
- Session shape: long-running main-agent conversation with many sequential GitHub, shell, and LCM tool calls
- Observed: July 11, 2026

## Current reproduction

Later compacted conversation history contained repeated entries in this form:

```text
[lossless-claw] missing tool result in session history; inserted synthetic error result for transcript repair.
```

A call-id pairing audit against the canonical raw session JSONL found the corresponding real tool-result rows. The audited July 11 trajectories had no unmatched live tool-call/tool-result pairs for the affected calls.

Therefore the supported claim is:

- canonical persistence had the real results;
- compacted Lossless projection/replay omitted, reordered, or failed to associate them;
- OpenClaw's generic transcript repair then synthesized false missing-result errors.

This report does **not** claim that every adjacent tool command succeeded, nor that the OpenClaw session guard never has a separate flush race. It isolates the case where a real matching result is already present in canonical storage before the compacted history is reconstructed.

## Historical lineage

- **2026-03-28 to 2026-03-30 — Lossless PR #205:** merged assembler protections intended to avoid synthetic orphan tool repairs by dropping historical orphan calls and protecting matching tail results: https://github.com/Martian-Engineering/lossless-claw/pull/205
- **2026-04-02 — Lossless #234:** ACP-focused tracker; an April 7 comment also records a long-session/compaction synthetic-result symptom, but the owning reproduction is different: https://github.com/Martian-Engineering/lossless-claw/issues/234
[...]

Issue #1040, opened 2026-07-27
Bug: Duplicate message ingestion in afterTurn when transcript redaction differs from live message content


bowenluo718:
## Bug Description

The lossless-claw context engine can ingest the same semantic message twice into the LCM database when OpenClaw is configured with `logging.redactPatterns`. This occurs because the transcript file contains **redacted/sanitized** message content while the live `afterTurn` parameters contain **original/redacted** content, causing content-based deduplication to fail.

## Reproduction Steps

1. Configure OpenClaw with `logging.redactPatterns` that match sensitive fields (e.g., API keys, tokens)
2. Run a conversation turn that produces assistant/tool messages containing sensitive data
3. Observe duplicate message pairs in the LCM database with different `identityHash` values but identical `responseId` or `toolCallId`

### Current afterTurn Flow

```
reconcileTranscriptTailForAfterTurn (first) → deduplicateAfterTurnBatch → ingestBatch (last)
```

### The Problem

| Step | Source | Content State | Deduplication Key |
|------|--------|---------------|-------------------|
| **Step 1**: `reconcileTranscriptTailForAfterTurn` | transcript JSONL file | **Redacted** by `logging.redactSensitive()` at write time | N/A - imports directly |
| **Step 2**: `deduplicateAfterTurnBatch` | runtime snapshot | Original content | `SHA256(role + "\0" + content)` |
| **Step 3**: `ingestBatch` | live afterTurn params | **Original/unredacted** content | Same hash as Step 2 |

The core contradiction:
- Step 1 reads from the transcript file where content has already been sanitized (e.g., `"***": "sk-688…5e6d"`)
- Steps 2-3 use live parameters with original content (e.g., `"apiKey": "sk-6886f7dcd92a41d687ce92cb7d9f5e6d"`)
- Since `identityHash = SHA256(role + "\0" + content)`, different `content` values produce different hashes
- Result: the same message ingests twice

### Concrete Example from Production Logs

From `conversation-store-2026-6-11-17-01-09.log`:

```
messageId: 219, identityHash: fbae8572...  (content: "***": "sk-688…5e6d" ← from transcript reconcile)
messageId: 221, identityHash: 8fff6bea...  (content: "apiKey": "sk-6886f..." ← from live afterTurn)

messageId: 220, identityHash: f0d987b7...  (content: "***": "e27aed8..." ← from transcript reconcile)  
[...]

PR #1054, opened 2026-07-31
fix(ingest): preserve tool pairing for empty-content tool results

jetd1:
## Problem

Tools that return `{content: [], details: {...}}` (e.g. OpenClaw's `update_plan`, formerly `structured_output` and `agents-wait`, and any third-party tool) trigger a 5-step failure chain:

1. **Ingest** `buildMessageParts()` (`src/message-content.ts`): toolResult message with `content=[]` → ordinal loop runs zero times → **0 parts persisted**. Top-level `toolCallId`/`toolName`/`isError` only ever live inside part metadata → all lost. (messages row: `role=tool, content='', 0 parts`.)
2. **Assembly** `resolveMessageItem` (`src/assembler.ts` ~1810-1835): `role=tool` + 0 parts → `pickToolCallId()` returns `undefined` → demoted to `role:"assistant"` with `content: []`.
3. `isEmptyMessageContent` filter drops the demoted message entirely.
4. `sanitizeToolUseResultPairing` (`src/transcript-repair.ts`) sees the assistant's `toolCall` with no matching result → inserts `[lossless-claw] missing tool result…` synthetic error **every assemble** while the call is in the context window.
5. Model sees plan calls as "failed" → retry loops, corrupted context.

## Production evidence

~100% of `update_plan` results since April orphaned:

| Month | Orphaned | Total |
|-------|----------|-------|
| Apr   | 573      | 720   |
| May   | 605      | 619   |
| Jun   | 444      | 453   |
| Jul   | 69       | 70    |

Live repro on v0.15.1: conv 468, msg 472198 (call) / 472199 (role=tool, clen=0, 0 parts) → synthetic error.

## Why #993 didn't close the hole

Issue #992 was closed by PR #993 (`27cda30`) which dedups real-vs-synthetic results at **repair time** — but the **ingest hole** (zero parts persisted for empty-content results) was untouched at HEAD. The dedup only helps when a real result eventually arrives; for tools that genuinely return empty content, no real result ever exists.

## Relation to OpenClaw #116916

OpenClaw fixed `update_plan` itself (commit `87d5cb9f674`, PR #116916 — returns `textResult("Plan updated")`). They also fixed `structured_output` + `agents-wait`, confirming this is a class of tools. However, LCM must stay robust: any current/future/third-party tool returning empty content hits the same chain, and hosts running older OpenClaw keep emitting empty-content results.

## Fix design

**`buildMessageParts()` (`src/message-content.ts`):** when a `tool`/`toolResult` message's content is an empty array and no parts were produced, emit ONE fallback part:
- `partType: "tool"`, `ordinal: 0`
- `toolCallId`/`toolName` from top-level message fields
- `textContent: " "` (single space — keeps the fallback out of blank-content filter paths; effective tool output is still the empty string)
- `metadata`: `{originalRole, toolCallId, toolName, isError, emptyContentFallback: true}`

**Role-gated, not id-gated:** `blockFromPart` treats a metadata-less `"tool"` part as a `toolCall`, so only `tool`/`toolResult` roles may land here — an id-bearing `assistant`/`user` message would otherwise be misassembled as a phantom call.

**`message-signatures.ts`:** `createCanonicalEmptyToolResultCoverageSignature` canonicalizes the coverage identity for empty-content tool results to `(role, toolCallId)` — `toolName` is intentionally omitted because it may be present on the DB round-trip representation and absent on the live transcript block, while `toolCallId` is the unique pairing identity. This prevents live-coverage dedup from double-emitting the same logical result.

## Behavior matrix

| Message shape | Before | After |
|--------------|--------|-------|
| `toolResult` + `content: []` + `toolCallId` | 0 parts → demote → drop → synthetic error | 1 fallback part → `role=toolResult` preserved, no synthetic error |
| `toolResult` + `content: []` + no `toolCallId` | 0 parts → demote → drop | 1 fallback part (no id) → demote to assistant (same as today) |
| `toolResult` + `content: "text"` | 1 text part (unchanged) | 1 text part (unchanged) |
| `toolResult` + `content: [{type:"tool_result",...}]` | 1+ tool parts (unchanged) | 1+ tool parts (unchanged) |
| `assistant` + `content: []` | 0 parts → empty-assistant skip (intended) | 0 parts → empty-assistant skip (unchanged) |
| `assistant` + `content: []` + `reasoning_content` | 1 reasoning part (unchanged) | 1 reasoning part (unchanged) |
| `user` + `content: []` | 0 parts → dropped (intended) | 0 parts → dropped (unchanged) |

## Audit results

- **a. Signatures:** `createLosslessMessageSignature` now includes the fallback part → new signatures for empty-content tool messages. `createCanonicalEmptyToolResultCoverageSignature` ensures live-coverage signatures match across live/DB/assembly representations. Transcript-reconciler replay-prefix detection uses `createLosslessMessageSignature` on bootstrap candidates; since empty-content tool messages are `isBootstrapReplayCandidateMessage` (role=tool), old persisted rows (0 parts) and new live messages (1 fallback part) would produce different signatures — but this only affects bootstrap replay detection across an upgrade boundary, which is safe: a mismatch means the guard fails to detect a replay (false negative), not a false positive drop. The canonical coverage signature handles the within-process live-vs-assembly case.
- **b. `live-coverage.ts`:** `normalizeLiveMessageForAssemblyReconciliation` calls `buildMessageParts` → `contentFromParts`. With the fallback part, this now produces a `tool_result` block instead of empty content, consistent with the assembled shape.
- **c. `transcript-repair.ts`:** With the fallback part, `pickToolCallId` returns the correct id → `resolveMessageItem` sets `role=toolResult` → `sanitizeToolUseResultPairing` finds the matching result → **no synthetic error inserted**. The #993 real-result-preferred dedup (`shouldUseCandidateToolResult`) still works because the fallback part produces a non-synthetic `toolResult` message.
- **d. `engine.ts` ingest (~2911):** `buildMessageParts` is called after all interceptors. The empty-content check is inside `buildMessageParts` itself — no count/emptiness assumptions at the call site.
- **e. Externalization:** The `!rawPayloadExternalized` guard prevents the fallback from firing on raw-payload-externalized messages. The fallback part's `textContent: " "` is well under any large-file threshold. `interceptLargeToolResults` only processes array content (non-empty check via `!Array.isArray` returns null for `[]`), so no externalization triggered.

## Test coverage

- `test/empty-content-tool-result-parts.test.ts` (3 tests): ingest→assemble round-trip (paired toolResult, no synthetic error), repeated assemble stability, legacy zero-part rows no worse than today.
- `test/message-content-empty-tool-result.test.ts` (16 tests): fallback part emission (toolResult empty, tool role, callid-less, isError preserved), role-gating (assistant/user NOT affected, id-bearing assistant NOT latched), block routing (toolResultBlockFromPart, serializable content), live-coverage signature stability (live vs assembled match, id-less no collision), normalizeMessageContentForStorage shape.

Full suite: `npx vitest run --dir test` → 113 files, 2003 tests passed. `npx tsc --noEmit` → clean.

## Notes

- Trivial merge adjacency with PR #1053 (`fix/preserve-thinking-identity`, same file `message-content.ts` different concern) may exist — the hunks are in different functions.
- Legacy DB rows (already stored with 0 parts) are out of scope for repair — they keep today's demote-to-assistant behavior. A future migration could backfill fallback parts, but that is not included here.

Refs #992.

### DIFF
--- a/.changeset/fix-empty-content-tool-result-parts.md
+++ b/.changeset/fix-empty-content-tool-result-parts.md
@@ -1,0 +1,19 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix tool-result pairing loss when a tool returns empty content. Tools like
+OpenClaw's `update_plan` (and any third-party tool) that return
+`{content: [], details: {...}}` produced zero `message_parts` rows on ingest,
+losing the top-level `toolCallId`/`toolName`/`isError` pairing identity. On
+assembly, the zero-part `role=tool` row was demoted to `role="assistant"` with
+empty content and dropped by the empty-content filter, causing
+`sanitizeToolUseResultPairing` to insert a synthetic "missing tool result"
+error on every assemble while the call sat in the context window — the model
+read its own plan calls as "failed" and retried in a loop. The fix emits one
+identity-carrying fallback part in `buildMessageParts()` when a tool-result
+message's content is an empty array, so assembly can reconstruct the
+`toolResult` with the correct `toolCallId`. A canonical empty-tool-result
+coverage signature is added to `message-signatures.ts` so live-coverage dedup
+recognizes the live, DB-round-trip, and assembled representations of the same
+logical result as identical.
--- a/src/message-content.ts
+++ b/src/message-content.ts
@@ -627,6 +627,50 @@ export function buildMessageParts(params: {
     });
   }
 
+  // Empty-content tool-result fallback (issue #992): a tool result message
+  // whose content is an empty array (e.g. OpenClaw `update_plan` shape
+  // `{content: [], details: {...}}`) produces zero parts above, which loses
+  // the top-level pairing identity (toolCallId/toolName/isError live only in
+  // part metadata). Downstream, a zero-part role=tool row can no longer be
+  // rehydrated as a toolResult — it is demoted to assistant, filtered as
+  // empty, and `sanitizeToolUseResultPairing` then injects a synthetic
+  // "missing tool result" error on every assemble while the call is in the
+  // context window. Persist one fallback part that carries the pairing
+  // identity so assembly can reconstruct the toolResult. Role-gated (not
+  // id-gated): pre-existing routing in blockFromPart treats a metadata-less
+  // "tool" part as a toolCall, so only tool/toolResult roles may land here —
+  // an id-bearing assistant/user message would otherwise be misassembled as
+  // a phantom call. User and assistant empty content intentionally stays
+  // part-less (the empty-content filter and the empty-assistant skip are
+  // desired behavior). Callid-less tool rows keep today's demote-to-assistant
+  // assembly path.
+  if (
+    message.content.length === 0 &&
+    (role === "tool" || role === "toolResult") &&
+    !rawPayloadExternalized
+  ) {
+    parts.push({
+      sessionId,
+      partType: "tool",
+      ordinal: 0,
+      textContent: " ",
+      toolCallId: topLevelToolCallId ?? null,
+      toolName: topLevelToolName ?? null,
+      metadata: toJson({
+        // Always identify as "toolResult" so blockFromPart routes through
+        // toolResultBlockFromPart (not toolCallBlockFromPart which would
+        // emit a phantom toolCall block). The raw role may be "tool" but
+        // the logical shape is a tool RESULT.
+        originalRole: "toolResult",
+        rawType: "tool_result",
+        toolCallId: topLevelToolCallId,
+        toolName: topLevelToolName,
+        isError: topLevelIsError,
+        emptyContentFallback: true,
+      }),
+    });
+  }
+
   return parts;
 }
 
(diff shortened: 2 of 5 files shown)


jetd1:
Closing in favour of #1062.

While validating this branch in production we hit a much larger data-loss bug in the same area: `extractStableEventKey()` (#1040) treats `tool-result:<toolCallId>` as conversation-globally unique, but model-authored tool-call ids (Kimi K3 `exec:101`, `read:92`, …) recur across turns. Once one result for an id was stored, every later distinct result reusing that id was skipped — on both the runtime ingest path and the transcript-reconcile path — until the assembled transcript was mostly synthetic `missing tool result` errors.

#1062 carries this PR's `#992` empty-content fix unchanged, plus the root-cause fix (provider-unique stable-key scoping, never-lose-a-row conflict degradation) and occurrence-scoped pairing/coverage hardening. Full incident write-up and evidence in the #1062 description.

Thanks for the review here — both inline P3 findings (the legacy zero-part test that never deleted its parts, and the dropped `details` payload) are addressed in #1062.

PR #1058, opened 2026-08-01
fix(session-rollover): accept host-minted /new replacement transcripts (sibling-header instant correlation)

### LINKED ISSUE #1055: Identity-overlap freshness check after /new freezes ingest permanently and near-silently when the user re-sends the same text
## Summary

After a `/new`, the ambiguous-rollover freshness check refuses to rebind the conversation when the first message of the new transcript repeats substantial (more than 8 characters) content from the last 50 persisted rows. The refusal itself is the documented fail-closed defense against a foreign session reusing a stale sessionKey, and this report does not argue with that goal. The problem is the shape of the failure:

1. **It is permanent for the entire life of the new session.** The overlapping entry never leaves the new transcript, so the verdict re-derives identically on every subsequent turn. One same-text re-send wedges the lane's ingest until the user happens to run `/new` again. Every message sent in the frozen window is permanently absent from the database (it survives only in the transcript file on disk).
2. **It is nearly silent.** Exactly one WARN fires, naming neither the overlapping content nor the candidates, and the once-only memo suppresses all restatements. Every later turn logs the same INFO deferral line as a healthy transient rebind, so a frozen lane is indistinguishable from normal operation without a direct DB probe.

"Reset, then re-send my question" is a natural user pattern, so this is easy to hit accidentally and very hard to notice.

## Reproduction (live, main @ 8ac4720)

Reproduced 2026-08-01 on a build cut from current main ([`8ac4720`](https://github.com/Martian-Engineering/lossless-claw/commit/8ac47205d259), includes [#1042](https://github.com/Martian-Engineering/lossless-claw/pull/1042)), OpenClaw 2026.6.10, dashboard chat main lane. Log lines below have the agent name and home directory genericized; everything else is verbatim.

1. Send the agent any message longer than 8 characters, for example a short greeting. It ingests normally (in our run: conversation 145, rows 92/93).
2. Run `/new` and wait for rotation to complete (the `.reset.<ISO>` archive sibling appears next to the old transcript).
3. Re-send the byte-identical message as the first message of the new session.
4. Keep chatting normally.

Step 3 fires the one and only WARN:

```json
{"time":"2026-08-01T08:25:44.456Z","level":"warn","plugin":"lossless-claw","message":"[lcm] afterTurn: ambiguous rollover not provably fresh conversation=145 sessionKey=agent:example:main freshness=identity-overlap-with-persisted-history lastPersistedAt=2026-08-01T08:20:07.000Z firstCandidateAt=2026-08-01T08:25:42.092Z"}
```

Every subsequent turn (step 4) logs only this INFO line, which reads like a benign transient:

```json
{"time":"2026-08-01T08:37:31.614Z","level":"info","plugin":"lossless-claw","message":"[lcm] afterTurn: tracked transcript archived after /new (reset sibling present); deferring to ambiguous-rollover rebind conversation=145 sessionKey=agent:example:main oldSessionId=ef5418da-3586-400d-a5fd-f03f445f614d newSessionId=d0b38af6-43e6-4091-bc5a-14536ff66f51 oldFile=~/.openclaw/agents/example/sessions/ef5418da-3586-400d-a5fd-f03f445f614d.jsonl"}
```

Database state during the frozen window: the conversation row stays `active=1`, frozen at its pre-`/new` row count (93), no successor conversation is created for the key, and none of the five messages sent after the `/new` ingested. The lane was still wedged 15+ minutes later at the time of writing; the state does not self-resolve.

Healing requires running `/new` again and opening with text the agent has not seen recently. The rebind then resolves immediately (`ambiguous rollover resolved by fresh-transcript rebind ... candidateMessages=2`) and ingest resumes. The messages from the frozen window remain lost to the DB.

## Mechanism (references to main @ 8ac4720)

- [`hasResetArchivedTranscriptSibling`](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L176) detects the `/new` archive; the assemble/afterTurn paths then defer to the ambiguous-rollover rebind ([L268](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L268), the INFO line above).
- [`evaluateAmbiguousRolloverFreshness`](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L437) scans the last [`AMBIGUOUS_ROLLOVER_OVERLAP_WINDOW = 50`](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L31) persisted rows and returns [`identity-overlap-with-persisted-history`](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L591) when a candidate shares content longer than [`TRIVIAL_ROLLOVER_OVERLAP_CONTENT_MAX_LENGTH = 8`](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L61).
- The [module docstring](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L39-L60) is explicit that a substantial overlap fails closed even when the `/new` marker and the archive sibling are both present, "because that is exactly the shape a foreign session reusing a stale sessionKey would produce".
- Permanence: the offending entry is part of the new transcript, so every turn re-derives the same verdict and the comparison never advances.
- Near-silence: the [once-only WARN memo](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L128-L145) (applied at [L648-L653](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/session-rollover.ts#L648-L653)) suppresses every restatement, and the WARN itself carries no content detail to act on.

## Additional consequences while wedged (observed live on the repro lane)

Two further user-visible consequences, both verified on the wire (provider request payloads) on the repro lane above. They materially raise the impact beyond "ingest stops".

**1. The model-visible context face silently flips.** While the lane is wedged, `assemble` never completes, so the host-native transcript face reaches the provider instead of the assembled one. Concretely, on the wedged lane every replayed user message suddenly carries the host's envelope timestamp decoration (for example `[Sat 2026-08-01 11:25 GMT+3] ...`), no condensed context is injected, and assistant reasoning appears only in provider-native sibling fields rather than the assembled face. On a healthy lane none of that is visible. So a wedged lane does not just stop remembering; it changes what the model sees on every turn, silently.

**2. Every turn loses the agent's newest reply, desynchronizing the conversation.** The ambiguous-rollover branch returns `safeFallback()` ([src/engine.ts#L3627](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/engine.ts#L3627)), whose prefill-strip loop ([src/engine.ts#L3557-L3563](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/engine.ts#L3557-L3563)) pops trailing assistant messages. On the embedded OpenClaw host the trailing assistant is the completed previous reply (the current inbound arrives via the separate `prompt` argument), so each request's history is missing the newest reply, exactly one turn behind, deterministically. The model sees the previous user turn as unanswered and re-answers it alongside the current one; the user experiences persistent double-answering with shifting tones. We verified the reply generated at 08:26:17Z is absent from the 08:28:09Z request and present in the 08:29:25Z one, with the lag unchanged across an 8-minute idle gap.

The pop itself is a separable defect (it fires on every fallback path, not only this one) and is filed separately as [#1056](https://github.com/Martian-Engineering/lossless-claw/issues/1056); the wedge is what turns it from an occasional single-request loss into an every-turn desync. Worth noting: the assemble branch's comment says "bootstrap/afterTurn heal with the full transcript file on the next call" ([src/engine.ts#L3615-L3626](https://github.com/Martian-Engineering/lossless-claw/blob/8ac47205d259/src/engine.ts#L3615-L3626)), but under the identity-overlap verdict that next-call heal never arrives, which is exactly the interaction this report is about.

## Proposed fix (happy to submit a PR if the direction works for you)

The defense wants proof that the new transcript is the host-minted replacement for exactly this `/new`, rather than a foreign transcript that grabbed the key. Marker plus sibling presence alone cannot prove that, as the docstring correctly argues. But two independently host-minted instants can:

- the archive sibling's `.reset.<ISO>` timestamp, which `hasResetArchivedTranscriptSibling` already locates, and
- the new transcript's own `{"type":"session"}` header creation timestamp.

Both are minted by the same host-side `/new` operation; on a live rotation we measured them 1 ms apart. A foreign transcript reusing a stale sessionKey cannot satisfy the correlation, since its header instant is unrelated to the reset instant.

Proposal: when both artifacts are present and the delta is within a conservative tolerance (seconds, not minutes; fail closed outside it), accept the rebind with a new verdict reason such as `fresh-host-minted-reset-replacement-transcript`, bypassing the content-overlap scan for exactly this case. Missing either artifact keeps today's behavior bit-for-bit, so the foreign-reused-key defense stands. Existing fixtures emit no session header, which means the pinned tests (`test/soft-reset-new-sibling-rebind.test.ts` and the trivial-content boundary test in `rollover-identity-scope.test.ts`) pass unchanged; the new path would ship with its own red-proof tests.

Wiring sketch: `hasResetArchivedTranscriptSibling` returns the parsed instant instead of a boolean, the rollover context carries it, and the transcript header parse exposes its creation timestamp to the freshness evaluator.

Independent of the rebind change, two small observability improvements would make a frozen lane diagnosable at all: include the candidate count and the overlapping-content length in the not-provably-fresh WARN, and re-emit it at INFO periodically instead of never again after the memo.

One caveat we flagged ourselves: the 1 ms delta was measured on one rotation on one host, so the tolerance needs to absorb slow disks and busy hosts while staying conservative.

## Related

[#887](https://github.com/Martian-Engineering/lossless-claw/issues/887) also describes a permanent post-rollover freeze, but through a different mechanism (real unstamped rows, missing entry-id lineage). This one is the identity-overlap fail-closed verdict firing on a genuine host-minted `/new` replacement transcript.

[#1056](https://github.com/Martian-Engineering/lossless-claw/issues/1056) tracks the fallback prefill-strip defect that the wedge amplifies into the every-turn reply lag described above.

gorkem2020:
Closes [#1055](https://github.com/Martian-Engineering/lossless-claw/issues/1055).

## Problem

The ambiguous-rollover freshness gate fails closed on any substantial identity overlap with recent persisted history, and the verdict re-derives identically on every turn because the overlapping entry never leaves the new transcript. One same-text re-send across `/new` (the natural "reset, then re-send my question" pattern) therefore wedges the lane's ingest until the next `/new`, near-silently. Full trace, live reproduction, and downstream consequences are in [#1055](https://github.com/Martian-Engineering/lossless-claw/issues/1055).

## Fix

Correlate two independently host-minted instants: the archive sibling's `.reset.<ts>` tail and the new transcript's `{"type":"session"}` header `timestamp`. Both are minted by the same host-side `/new` operation (measured 0-1 ms apart on a live rotation), so their agreement within a conservative tolerance (30 s, `HOST_MINTED_RESET_REPLACEMENT_TOLERANCE_MS`) proves the new transcript is the host's replacement for exactly that reset, and overlapping content across the boundary is the same user's traffic. The freshness gate then accepts the rebind with a new verdict reason, `fresh-host-minted-reset-replacement-transcript`, bypassing only the content-overlap scan.

The defense the docstring describes is preserved and sharpened. A foreign transcript reusing a stale sessionKey carries a header instant unrelated to the reset instant and cannot satisfy the correlation. The bypass additionally requires all of: Lossless's own `/new` marker (`softResetPrunedAt`), the archive sibling, a parseable sibling instant, and a parseable header instant; missing any of them keeps today's behavior bit-for-bit. The per-entry timestamp gates (`candidate-missing-timestamp`, `candidate-entries-predate-last-persisted`) still run ahead of the bypass.

Wiring: `TranscriptHeader` gains `sessionHeaderCreatedAt` (parsed from the header's `timestamp` field); the detector computes `hostMintedResetReplacement` inside `findAmbiguousSessionKeyRuntimeRollover` (both tier-2 resolving callers already pass `sessionFile`, so no caller changes); `evaluateAmbiguousRolloverFreshness` takes the evidence as an opt-in flag that defaults to false, so the isolated-cron path is untouched.

Observability (the second half of the issue): the not-provably-fresh WARN now carries `candidateMessages` and `overlapContentLength`, and a still-frozen lane re-emits the suppressed restatement at INFO every 25th occurrence instead of going debug-silent forever after the once-only WARN.

## Validation

- Red-proof: the new test "rebinds despite substantial same-text overlap when the sibling and header instants correlate" fails on current master (the lane freezes) and passes with the fix; the tolerance test also red-proves the WARN enrichment (2 failed on master, both green after).
- New fail-closed pins, all keeping the freeze: header instant outside tolerance; headerless new transcript despite a parseable sibling instant; unparseable sibling suffix.
- Existing pinned tests unchanged and green: the foreign reused-key transcript with a sibling (`test/soft-reset-new-sibling-rebind.test.ts`) and the trivial-content boundary (`rollover-identity-scope.test.ts`). Existing fixtures emit no session header, so they exercise the unmodified fail-closed path by construction.
- Targeted file: `test/soft-reset-new-sibling-rebind.test.ts` 10/10.
- Full suite: 111 files, 1988/1988 passing. `npm run typecheck` clean, build clean.
- Live-deploy note: we run a fleet on the embedded OpenClaw host and will stage this build and re-run the identical-text `/new` probe that reproduced [#1055](https://github.com/Martian-Engineering/lossless-claw/issues/1055) on 2026-08-01, expecting acceptance via the new verdict reason instead of the wedge. Keeping this PR as a draft until that live proof is in hand.

## Notes for review

- Tolerance: 0-1 ms measured on one rotation on one host; 30 s is deliberately generous for slow disks and busy hosts while staying far below any plausible foreign-transcript coincidence. Happy to tune.
- The archive-suffix parser accepts the host's dashed-time shape (`2026-08-01T08-25-10.924Z`, verified live) plus plain ISO; anything else yields null and fails closed.

## Changeset

User-facing behavior change (a same-text re-send across `/new` no longer freezes the lane). Flagging for a maintainer-added changeset per contributor convention rather than running the Changesets workflow from a fork.

### DIFF
--- a/.changeset/host-minted-reset-replacement.md
+++ b/.changeset/host-minted-reset-replacement.md
@@ -1,0 +1,8 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Recover ingestion after `/new` when the replacement transcript repeats
+persisted content by correlating the host-minted reset archive and session
+header timestamps. Transcripts without matching reset evidence continue to
+fail closed.
--- a/src/session-rollover.ts
+++ b/src/session-rollover.ts
@@ -125,13 +125,14 @@ export type AmbiguousSessionKeyRuntimeRollover = {
    */
   hasDeliberateRolloverEvidence: boolean;
   /**
-   * True when the new transcript's session-header creation instant follows
-   * the newest `.reset.` sibling's archive instant within
+   * True when the new transcript's header id matches the current runtime
+   * session and its creation instant follows the newest `.reset.` sibling
+   * archive instant within
    * `HOST_MINTED_RESET_REPLACEMENT_TOLERANCE_MS` (on top of the
-   * deliberate-rollover evidence above): the new transcript is the
-   * host-minted replacement for exactly this /new, so the freshness gate may
-   * bypass the identity-overlap scan. Missing either artifact leaves this
-   * false and behavior unchanged (fail closed).
+   * deliberate-rollover evidence above): the new transcript is bound to the
+   * host's current session for this /new, so the freshness gate may bypass the
+   * identity-overlap scan. Missing or mismatched evidence leaves this false
+   * and behavior unchanged (fail closed).
    */
   hostMintedResetReplacement: boolean;
 };
@@ -299,22 +300,23 @@ export class SessionRolloverDetector {
   }
 
   /**
-   * True when the new transcript's session-header creation instant follows
-   * the newest `.reset.` sibling's archive instant within
-   * `HOST_MINTED_RESET_REPLACEMENT_TOLERANCE_MS`. Both artifacts are minted
-   * by the same host-side /new operation, so their ordered proximity proves
-   * the new transcript is the host's replacement for exactly that reset; a
-   * foreign transcript reusing a stale sessionKey carries a header instant
-   * unrelated to the reset instant and cannot satisfy the correlation. Requires the
-   * deliberate-rollover evidence pair and both instants; anything missing
-   * yields false (fail closed).
+   * True when the new transcript's header id matches the current runtime
+   * session and its creation instant follows the newest `.reset.` sibling's
+   * archive instant within `HOST_MINTED_RESET_REPLACEMENT_TOLERANCE_MS`.
+   * OpenClaw mints the header id from that runtime session id, so the identity
+   * match binds the candidate file to the current host session while the
+   * ordered timestamps bind it to the reset window. Requires the deliberate
+   * rollover evidence pair, matching identity, and both instants; anything
+   * missing yields false (fail closed).
    */
   private async isHostMintedResetReplacementTranscript(params: {
     hasDeliberateRolloverEvidence: boolean;
+    sessionId: string;
     trackedSessionFile: string;
     sessionFile?: string;
   }): Promise<boolean> {
-    if (!params.hasDeliberateRolloverEvidence || !params.sessionFile) {
+    const sessionId = params.sessionId.trim();
+    if (!params.hasDeliberateRolloverEvidence || !sessionId || !params.sessionFile) {
       return false;
     }
     const resetArchivedAt = await this.latestResetArchivedTranscriptSiblingInstant(
@@ -324,7 +326,7 @@ export class SessionRolloverDetector {
       return false;
     }
     const header = await readTranscriptHeader(params.sessionFile);
-    if (!header.sessionHeaderCreatedAt) {
+    if (header.sessionHeaderId !== sessionId || !header.sessionHeaderCreatedAt) {
       return false;
     }
     const headerCreatedAt = Date.parse(header.sessionHeaderCreatedAt);
@@ -525,6 +527,7 @@ export class SessionRolloverDetector {
 
     const hostMintedResetReplacement = await this.isHostMintedResetReplacementTranscript({
       hasDeliberateRolloverEvidence,
+      sessionId: params.sessionId,
       trackedSessionFile,
       sessionFile: params.sessionFile,
     });
--- a/src/session-rollover.ts
+++ b/src/session-rollover.ts
@@ -68,9 +68,14 @@ function isTrivialRolloverOverlapContent(content: string): boolean {
  * Max distance between an archive sibling's `.reset.<ts>` instant and the new
  * transcript's session-header `timestamp` for the pair to count as one
  * host-minted /new operation. Both artifacts are minted by the same host
- * handler (measured 0-1 ms apart on a live rotation); the allowance absorbs
- * slow disks and busy hosts while staying far below any plausible
- * foreign-transcript coincidence.
+ * handler within one /new (measured 0-1 ms apart on a live rotation); the
+ * allowance absorbs slow disks, busy hosts, and stalls between the archive
+ * rename and the header write while staying far below any plausible
+ * foreign-transcript coincidence. The window is also not the only defense: a
+ * transcript inside it must still carry Lossless's own /new marker
+ * (`softResetPrunedAt`) plus the `.reset.` sibling, and the per-entry
+ * timestamp gates (`candidate-missing-timestamp`,
+ * `candidate-entries-predate-last-persisted`) run ahead of the bypass.
  */
 const HOST_MINTED_RESET_REPLACEMENT_TOLERANCE_MS = 30_000;
 
@@ -84,12 +89,15 @@ const AMBIGUOUS_ROLLOVER_FROZEN_RESTATEMENT_INFO_EVERY = 25;
 /**
  * Parse the `<ts>` tail of a host archive name (`${file}.reset.<ts>`). The
  * host mints ISO-8601 with the time separators replaced by dashes
- * (`2026-08-01T08-25-10.924Z`); a plain ISO tail is accepted too. Returns
- * epoch ms, or null when the tail does not parse (callers fail closed).
+ * (`2026-08-01T08-25-10.924Z`); a plain ISO tail is accepted too. A tail
+ * without a time component (e.g. a bare date, which `Date.parse` would
+ * happily read as midnight) is rejected: it cannot pin the archive to an
+ * instant precisely enough to correlate. Returns epoch ms, or null when the
+ * tail does not parse (callers fail closed).
  */
 function parseResetArchiveSuffixInstant(suffix: string): number | null {
   const trimmed = suffix.trim();
-  if (!trimmed) {
+  if (!/T\d{2}/.test(trimmed)) {
     return null;
   }
   const candidates = [trimmed, trimmed.replace(/T(\d{2})-(\d{2})-(\d{2})/, "T$1:$2:$3")];
--- a/src/session-rollover.ts
+++ b/src/session-rollover.ts
@@ -15,7 +15,7 @@ import { createBootstrapEntryHash, messageIdentity } from "./message-signatures.
 import type { AgentMessage } from "./openclaw-bridge.js";
 import { isIsolatedCronSessionKey } from "./session-patterns.js";
 import type { ArchiveCause, ConversationStore } from "./store/conversation-store.js";
-import { getTranscriptEntryMeta } from "./transcript.js";
+import { getTranscriptEntryMeta, readTranscriptHeader } from "./transcript.js";
 import { isMissingFileError } from "./value-utils.js";
 import type { SummaryStore } from "./store/summary-store.js";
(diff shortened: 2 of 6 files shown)


gorkem2020:
Live proof executed on our fleet (OpenClaw 2026.6.10 embedded host, this branch deployed via our tip-tracking build):

1. A unique baseline message ingested normally (conversation at 91 rows).
2. `/new`: the host archived the old transcript and minted the new session 1 ms apart (sibling `.reset.2026-08-01T10-27-59.813Z` vs header `timestamp` `2026-08-01T10:27:59.814Z`), exactly the correlation this fix keys on.
3. Re-sent the byte-identical baseline as the first message of the new session. On the previous build this exact sequence permanently wedged a lane earlier today (the reproduction in [#1055](https://github.com/Martian-Engineering/lossless-claw/issues/1055)); on this branch the rebind resolved within 2 seconds:

```json
{"time":"2026-08-01T10:28:28.613Z","level":"info","plugin":"lossless-claw","message":"[lcm] afterTurn: ambiguous rollover resolved by fresh-transcript rebind conversation=169 sessionKey=agent:example:main oldSessionId=5b943c15-2e56-440d-972b-b9ccb360256c newSessionId=e04985a6-7cac-48a8-be1d-02cb41064dc1 candidateMessages=2 ..."}
```

4. The re-sent message and its reply ingested into the same conversation (93 rows), and a follow-up unique message kept ingesting (95 rows). No warnings fired; the transient pre-heal deferral lines now carry the new `candidateMessages`/`overlapContentLength` fields.

Marking ready for review. (Log line lightly genericized: agent name only.)

gorkem2020:
Review round addressed in 214d80b: the 30s tolerance stays as reviewed, with the constant's docstring expanded to spell out the rationale (same-handler mint measured 0-1 ms apart, margin for slow disks and host stalls, and the marker, sibling, and per-entry timestamp gates that still apply inside the window) and two boundary pins added (29s accepts, 31s fails closed). `parseResetArchiveSuffixInstant` now requires a time component, so a date-only tail (which `Date.parse` would read as midnight) is rejected; the new date-only test fails on the previous head and passes with the fix. The positive host-minted test now also asserts the retained summary carries across the rebind, mirroring the plain sibling-rebind case. Typecheck, build, and full suite (1991/1991) green.

jalehman:
Thanks for the thoughtful contribution, @gorkem2020. This is landed now.

PR #1061, opened 2026-08-01
fix: delegated expansion cannot collect replies on current OpenClaw beta

### LINKED ISSUE #1059: [Bug]: lcm_expand_query cannot collect replies after subagent.getSession removal
## Summary

`lcm_expand_query` can successfully run delegated expansion agents but then fails to collect their completed replies because LosslessClaw calls the removed `api.runtime.subagent.getSession()` compatibility alias instead of the supported `getSessionMessages()` method.

## Environment

- LosslessClaw installed build: `0.15.0-main.2dbf925.pr952.addf9cb.pr1025.9c49bc9`
- Latest released LosslessClaw checked: `0.15.1`
- OpenClaw: `2026.7.2` (`5c061507d3`)
- macOS arm64, Node.js `v26.3.0`

Current LosslessClaw `main` still contains the same call in `src/plugin/index.ts`:

```ts
case "sessions.get":
  return sub.getSession({
    sessionKey: String(params.params?.key ?? ""),
    limit: params.params?.limit as number | undefined,
  });
```

## Steps to reproduce

1. Run a supported OpenClaw version where `api.runtime.subagent` exposes `getSessionMessages()` but no longer exposes the deprecated `getSession()` alias.
2. Invoke `lcm_expand_query` with summaries that require delegated expansion.
3. Allow the delegated subagent to complete successfully.
4. Observe reply collection fail when LosslessClaw dispatches its internal `sessions.get` adapter.

## Expected behavior

LosslessClaw reads the completed delegated session through `sub.getSessionMessages()` and returns the synthesized answer with its cited summary IDs.

## Actual behavior

The delegated agents complete and produce valid JSON answers, but the parent tool cannot retrieve them and returns failures with:

```text
sub.getSession is not a function
```

The tool classifies the buckets as:

```text
errorCode: DELEGATED_EXPANSION_REPLY_MISSING
phase: read_reply
```

This classification is misleading in this case: the replies exist; the adapter crashes while reading them.

## Compatibility evidence

OpenClaw exposed `getSessionMessages()` alongside the deprecated `getSession()` alias in both `v2026.5.28` and `v2026.7.2-beta.2`. OpenClaw removed the expired alias in commit [`c7e7ac272891`](https://github.com/openclaw/openclaw/commit/c7e7ac2728917f085aff0ad50755fd39aa57f2b3).

Therefore changing LosslessClaw to `getSessionMessages()` is compatible with its currently declared supported OpenClaw range; it does not require dropping support for those older supported versions.

## Impact

- `lcm_expand_query` spends time and model usage completing delegated work but cannot return it.
- Multiple selected conversation buckets can all fail at the reply-reading phase.
- Operators must recover answers from gateway logs or the underlying sessions.
- Ordinary context management, summarization, `lcm_grep`, and `lcm_describe` remain functional.

## Suggested regression coverage

Exercise one delegated expansion through completion with a runtime exposing only the current subagent API (`run`, `waitForRun`, `getSessionMessages`, and `deleteSession`) and assert that the synthesized answer and citations are returned.

jjjhenriksen:
Closes #1059

## What Problem This Solves

Fixes an issue where `lcm_expand_query` delegated expansions could complete successfully but fail to return their answer and citations on current OpenClaw beta hosts.

## Why This Change Was Made

OpenClaw removed the deprecated `api.runtime.subagent.getSession()` alias while retaining `getSessionMessages()`. This changes the Lossless Claw gateway adapter to use the supported method and adds regression coverage through the real delegated expansion path. The change is limited to issue #1059; the separate fallback and rollover fixes in #1057 and #1058 remain out of scope.

## User Impact

Delegated expansion replies are collected again on current OpenClaw beta releases instead of being reported as `DELEGATED_EXPANSION_REPLY_MISSING`. The regression fixture exposes only `run`, `waitForRun`, `getSessionMessages`, and `deleteSession`, matching the current runtime contract.

## Evidence

- `npm test -- --run test/plugin-config-registration.test.ts` — 49 passed
- `npm run typecheck` — passed
- `npm run build` — passed
- `npm test` — 111 test files and 1,985 tests passed

### DIFF
--- a/.changeset/runtime-subagent-session-messages.md
+++ b/.changeset/runtime-subagent-session-messages.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Use OpenClaw's supported `subagent.getSessionMessages()` runtime method when collecting delegated expansion replies, restoring `lcm_expand_query` compatibility with current OpenClaw beta releases.
--- a/src/plugin/index.ts
+++ b/src/plugin/index.ts
@@ -1465,7 +1465,7 @@ function createLcmDependencies(
             timeoutMs: (params.params?.timeoutMs as number) ?? params.timeoutMs,
           });
         case "sessions.get":
-          return sub.getSession({
+          return sub.getSessionMessages({
             sessionKey: String(params.params?.key ?? ""),
             limit: params.params?.limit as number | undefined,
           });
--- a/test/plugin-config-registration.test.ts
+++ b/test/plugin-config-registration.test.ts
@@ -9,6 +9,7 @@ import { closeLcmConnection } from "../src/db/connection.js";
 import { LcmContextEngine } from "../src/engine.js";
 import { clearAllSharedInit } from "../src/plugin/shared-init.js";
 import { resetStartupBannerLogsForTests } from "../src/startup-banner-log.js";
+import { runDelegatedExpansionLoop } from "../src/tools/lcm-expand-tool.delegation.js";
 
 const buildMemorySystemPromptAdditionMock = vi.hoisted(() => vi.fn());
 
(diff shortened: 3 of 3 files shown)


jalehman:
Thanks for the contribution, @jjjhenriksen. This is landed now.

PR #1066, opened 2026-08-04
fix(dedup): anchor the metadata-face collapse on fresh transcript provenance

### LINKED ISSUE #1050: Metadata-face duplicates return on decorating channels after the #1042 rider commit: provenance no longer anchors the pairwise collapse
## Summary

After [#1042](https://github.com/Martian-Engineering/lossless-claw/pull/1042) merged, the rider commit [`1307b950a`](https://github.com/Martian-Engineering/lossless-claw/pull/1042/commits/1307b950aea9f8f11c44f48022e5e633d30126fc) ("fix(dedup): keep heuristic metadata matches weak") reopens the metadata-face double-write that the PR was built to close, on channels whose inbound is decorated (observed on Slack DMs).

## Observed on a live deployment (v0.15.1, three consecutive Slack DM turns)

Every inbound stores twice:

1. a bare row carrying `transcript_entry_id` (the transcript-anchored ingest), then
2. a decorated row (`Conversation info (untrusted metadata)` wrapper) with NO `transcript_entry_id`, written by the end-of-turn persist of the prompt-side current turn.

With the see-through rendering from [#1036](https://github.com/Martian-Engineering/lossless-claw/pull/1036), both faces replay into every later prompt, so each turn compounds one duplicate for the life of the conversation.

## A/B evidence isolating the rider commit

- Same probe shape, same lane, pre-rider head `ed617a2` (what #1042 looked like before the final push): exactly one user row per turn, zero decorated wrappers.
- Current main (`8ac4720`, which also contains [#1044](https://github.com/Martian-Engineering/lossless-claw/pull/1044)): 3 turns, 3 bare+decorated pairs.
- Reverting only `1307b950a` on top of current main (clean revert, typecheck green, suite 1985/1985): back to one row per turn, zero wrappers, verified live on the same lane.

So the tightening commit is the mechanism; nothing else in the merge wave is implicated.

## Why (from the diff)

The decorated end-of-turn face is recognized only via the untimestamped metadata-body match. Under `1307b950a` that match no longer anchors a collapse even when the persisted counterpart is transcript-proven; it now requires an exact or timestamp anchor, which this pair structurally never has (the decorated face carries no transcript envelope).

## The tension, and two possible directions

The rider's rationale (a metadata block can be forged, so heuristic matches should stay weak) reads soundly for covered-frontier replay alignment. At the single-message collapse site, though, the trust decision can rest on the PERSISTED row's provenance rather than on the incoming face: a stored `transcript_entry_id` row was written by the host's own transcript flush and cannot be forged by message content, which was the original anchoring argument in #1042.

Happy to PR whichever direction you prefer:

1. restore provenance anchoring at the pairwise collapse site only (keep replay-alignment support weak as the rider intends), or
2. gate the decorated end-of-turn persist itself: skip persisting the prompt-side face when a transcript-proven body twin already exists in the same turn.

## Repro sketch

Any channel that decorates inbound (Slack): send one message, let the turn complete, then read the messages store for that conversation: two user rows for the turn, one with an entry id, one wrapper-faced without.

gorkem2020:
## Problem

[#1050](https://github.com/Martian-Engineering/lossless-claw/issues/1050): since [`1307b950a`](https://github.com/Martian-Engineering/lossless-claw/pull/1042/commits/1307b950aea9f8f11c44f48022e5e633d30126fc), every inbound turn on decorated channels (Slack DMs and similar) stores twice: the bare transcript-anchored row, then the decorated end-of-turn face as a second row. With the see-through rendering from [#1036](https://github.com/Martian-Engineering/lossless-claw/pull/1036), both faces replay into every later prompt and each turn compounds one duplicate for the life of the conversation. The commit is unreleased so far (0.15.1 predates it); the next release would ship it.

## The design tension, and how this resolves it

[`1307b950a`](https://github.com/Martian-Engineering/lossless-claw/pull/1042/commits/1307b950aea9f8f11c44f48022e5e633d30126fc)'s rationale stands: a metadata-body match alone must never anchor a collapse, and stored transcript provenance authenticates only the persisted row, not the claim that the incoming face is a replay of it. The problem is that the end-of-turn pair structurally never has the exact or timestamp anchor the commit now requires, so the double-write is guaranteed.

This PR adds the missing link as ground truth rather than a heuristic. Each transcript import path now reports the exact message ids it inserted, and the reconcile result carries them to afterTurn. Collapse strength then exists only on the transcript-covered route (the one route that proves this turn's inbound was actually ingested, because the reconcile read the transcript to its frontier) and requires the persisted row to satisfy all three of:

1. transcript provenance (host-written row),
2. its `message_id` in the set THIS turn's reconcile reported as inserted (collected at the insert sites themselves, so no insertion-order or seq property needs to be inferred),
3. being the conversation's newest user row after the reconcile.

Condition 2 ties the anchor to this very reconcile's writes (per review, replacing the earlier watermark inference entirely); condition 3 ties it to the turn's own inbound: a historical row a catch-up reconcile backfilled in the same pass is in the inserted set too, but on the covered route the turn's own inbound is always the newest user row, so the backfilled row never anchors while a newer user row exists (per review). The strength is also local to the matched row: a `provenanced-inbound` match authenticates only itself, never a neighboring redacted or otherwise unanchored match (per review). Concretely:

* The double-write pair always qualifies: the bare row is ingested by this very turn's reconcile as the newest user row, and the decorated face arrives in the same afterTurn.
* Any older or backfilled row never qualifies. A repeated body seconds later, an echo, a replay, or a catch-up import faces a row that fails condition 2 or 3, stays weak, and duplicates instead of trimming (the loss-safe direction is preserved).
* Routes without the covered proof (degraded checkpoint, heuristic boundary/full-cover, oversized, suffix) never strong-collapse and keep the pair.
* An aligned suffix trims only under the pre-existing independent anchor contract (exact/decorated, with redacted rows requiring an independent neighbor), or when every row in it proves itself. A suffix containing an unanchored row alongside only `provenanced-inbound` strength is left untrimmed.
* Provenance-less rows keep the existing behavior: never collapse on a metadata-body match.

No wall-clock heuristics are involved; the correlation is the row-insertion order under the per-session queue plus the newest-user-row identity, which is the "correlate the rows to the same transcript entry/turn" shape suggested in review.

## What changed

* `src/transcript-reconciler.ts` and `src/openclaw-bridge.ts`: `ingestSingle` reports the persisted row's message id; every import path (entry-id tail, batch/slow path, append-only fast path, afterTurn bootstrap, fresh-rollover import, no-anchor epoch import) collects the ids it inserted into `insertedMessageIds` on the reconcile result.
* `src/engine.ts`: afterTurn passes that set to the covered-frontier alignment only (absent or empty keeps the strong collapse off). Import-capped reconciles need no special handling: `blockedByImportCap` already makes the turn unsafe-to-advance and afterTurn skips persistence entirely.
* `src/batch-dedup.ts`: `persistedRowAnchorsSameTurnCollapse(record, floor, newestUserMessageId)` implements the three-condition rule; `alignedSuffixIsAnchored` implements the trim decision (independent anchor contract, or every row self-proven); `redactedMatchesHaveAdjacentAnchor` accepts only exact/decorated neighbors. The heuristic, oversized, checkpoint, and suffix routes are reverted to their previous behavior (keep the pair).
* `src/store/conversation-store.ts`: `getLastMessageWithRole` (newest user row lookup), provenance docstring updated to the three-condition rule.
* `.changeset/provenance-anchored-collapse.md`: patch changeset (review ask).
* Deliberately untouched: the recap-strip gate in `openclaw-inbound-metadata.ts` from the same commit (unrelated mechanism, interacts with identity canonicalization).

## Tests

* Red against current main, green here: `collapses a conv-info-only runtime copy onto the same turn's own transcript ingest without a sibling anchor` (`test/transcript-entry-id.test.ts`), a real afterTurn whose reconcile ingests the bare row, then collapses the decorated face; this is the live pair's exact route and reproduces the #1050 double-write on main. Plus the arm-level positive `collapses a lone decorated face onto exactly its own same-turn ingest`, also red against main.
* Red against the previous head of this branch, green here: `never anchors on a backfilled above-floor row` pins the review finding (an older user row imported by the same reconcile, above the floor, must stay weak; the prior head wrongly trimmed it).
* Pins kept or added for the loss-safe direction: the degraded and oversized routes keep the pair even for a transcript-proven row created moments ago (rapid repeated-body case), the missing-floor case keeps the pair, the fail-closed different-body case, and the no-provenance case.

## Validation

* `npm run typecheck` clean, `npm run build` clean.
* Red-proof matrix run both ways: with `src/` from current main, exactly the two positive collapse tests fail (the double-write reproducing); with `src/` from the previous head, exactly the backfill test fails (the false anchor reproducing); with this branch all pass. Full suite: 113 files, 2035/2035 passing (two pre-existing exact-shape pins on the ingest result updated for the new `messageId` field).
* Live deploy: running on our OpenClaw fleet (deployed 2026-08-04). Six ordinary decorated Slack turns each persisted exactly one user row on this fix lineage (the [`1307b950a`](https://github.com/Martian-Engineering/lossless-claw/pull/1042/commits/1307b950aea9f8f11c44f48022e5e633d30126fc) build stores bare plus decorated pairs for the same shape), and a byte-identical message re-sent 12 minutes later was correctly kept as a genuine new turn. The current form is under continued live observation.

Closes [#1050](https://github.com/Martian-Engineering/lossless-claw/issues/1050).

### DIFF
--- a/.changeset/provenance-anchored-collapse.md
+++ b/.changeset/provenance-anchored-collapse.md
@@ -1,0 +1,8 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Collapse the decorated end-of-turn face of an inbound message onto its bare
+transcript row when that row was ingested by the same turn's own reconcile,
+removing the per-turn double-write on decorated channels. Rows from earlier
+turns, however recent, and provenance-less rows never anchor a collapse.
--- a/src/batch-dedup.ts
+++ b/src/batch-dedup.ts
@@ -43,10 +43,30 @@ type StoredIncomingMatch =
   | "unproven-externalized"
   | "redacted";
 type CoveredStoredIncomingMatch = StoredIncomingMatch | "decorated";
-type AlignmentMatch = CoveredStoredIncomingMatch | "unanchored-inbound";
+type AlignmentMatch = CoveredStoredIncomingMatch | "unanchored-inbound" | "provenanced-inbound";
+
+/**
+ * True when a persisted row is host-proven (transcript provenance) AND was
+ * inserted during the current afterTurn's own transcript reconcile: its
+ * message_id lies above the conversation's newest message id captured just
+ * before that reconcile ran. Session writes are queue-serialized and
+ * message_id is insertion-ordered, so rows above the floor are exactly this
+ * turn's transcript ingests. Provenance authenticates the persisted row; the
+ * floor ties the incoming decorated face to that very ingest rather than to
+ * any older row, however recent. Either condition missing (including callers
+ * that pass no floor) keeps the match weak: it may support alignment but
+ * never anchors a collapse.
+ */
+function persistedRowIsCurrentTurnIngest(
+  record: Pick<MessageRecord, "transcriptEntryId" | "messageId">,
+  sameTurnIngestFloor: number | undefined,
+): boolean {
+  if (record.transcriptEntryId == null) return false;
+  return sameTurnIngestFloor !== undefined && record.messageId > sameTurnIngestFloor;
+}
 
 function isStrongReplayAnchor(match: AlignmentMatch | undefined): boolean {
-  return match === "exact" || match === "decorated";
+  return match === "exact" || match === "decorated" || match === "provenanced-inbound";
 }
 
 function redactedMatchesHaveAdjacentAnchor(matches: readonly AlignmentMatch[]): boolean {
@@ -244,6 +264,7 @@ export class BatchDeduplicator {
     sessionId: string,
     sessionKey: string | undefined,
     batch: AgentMessage[],
+    sameTurnIngestFloor?: number,
   ): Promise<AgentMessage[]> {
     if (batch.length === 0) return batch;
 
@@ -310,10 +331,19 @@ export class BatchDeduplicator {
               { allowUntimestampedInboundBodyMatch: true },
             )
           ) {
-            // Stored provenance qualifies this as alignment support, but it
-            // authenticates only the persisted row. Another exact/timestamp
-            // anchor must still prove the runtime batch is a replay.
-            matches.push("unanchored-inbound");
+            // Stored provenance authenticates only the persisted row. When
+            // that row was ingested by THIS turn's own reconcile (above the
+            // pre-reconcile floor), the incoming decorated face is the
+            // end-of-turn copy of that very ingest (the store double-write)
+            // and anchors the collapse. Any older row, however recent, stays
+            // weak and needs an independent exact/timestamp anchor, so an
+            // echo or replay of an earlier turn duplicates instead of
+            // trimming.
+            matches.push(
+              persistedRowIsCurrentTurnIngest(tailMessages[i]!, sameTurnIngestFloor)
+                ? "provenanced-inbound"
+                : "unanchored-inbound",
+            );
             continue;
           }
           aligned = false;
@@ -369,6 +399,7 @@ export class BatchDeduplicator {
     sessionId: string,
     sessionKey: string | undefined,
     batch: AgentMessage[],
+    sameTurnIngestFloor?: number,
   ): Promise<AgentMessage[]> {
     if (batch.length === 0) return batch;
 
@@ -391,6 +422,7 @@ export class BatchDeduplicator {
     }
     return this.deduplicateAfterTurnBatch(sessionId, sessionKey, batch, {
       oversizedNoOverlap: "ingest",
+      sameTurnIngestFloor,
     });
   }
 
@@ -418,7 +450,7 @@ export class BatchDeduplicator {
     sessionId: string,
     sessionKey: string | undefined,
     batch: AgentMessage[],
-    options?: { oversizedNoOverlap?: "ingest" | "skip" },
+    options?: { oversizedNoOverlap?: "ingest" | "skip"; sameTurnIngestFloor?: number },
   ): Promise<AgentMessage[]> {
     if (batch.length === 0) return batch;
 
@@ -472,6 +504,18 @@ export class BatchDeduplicator {
           lastDbMessage.content,
           batchAtBoundary.role,
           batchAtBoundary.content,
+          {
+            // Heuristic routes may collapse only against a row this turn's
+            // own reconcile ingested (see persistedRowIsCurrentTurnIngest).
+            allowCollapsedSpaceRunMatch: persistedRowIsCurrentTurnIngest(
+              lastDbMessage,
+              options?.sameTurnIngestFloor,
+            ),
+            allowUntimestampedInboundBodyMatch: persistedRowIsCurrentTurnIngest(
+              lastDbMessage,
+              options?.sameTurnIngestFloor,
+            ),
+          },
         ) &&
           !(await this.messagesDifferOnlyByHostRedaction(
             lastDbMessage,
@@ -488,7 +532,7 @@ export class BatchDeduplicator {
           rawPayloadContents,
           storedMessageCount,
           "prefix-mismatch",
-          { onNoOverlap: "ingest" },
+          { onNoOverlap: "ingest", sameTurnIngestFloor: options?.sameTurnIngestFloor },
         );
       }
     }
@@ -533,6 +577,16 @@ export class BatchDeduplicator {
             storedMessages[i]!.content,
             storedBatch[i]!.role,
             storedBatch[i]!.content,
+            {
+              allowCollapsedSpaceRunMatch: persistedRowIsCurrentTurnIngest(
+                storedMessages[i]!,
+                options?.sameTurnIngestFloor,
+              ),
+              allowUntimestampedInboundBodyMatch: persistedRowIsCurrentTurnIngest(
+                storedMessages[i]!,
+                options?.sameTurnIngestFloor,
+              ),
+            },
           )
         ) {
           return batch;
@@ -563,7 +617,7 @@ export class BatchDeduplicator {
     rawPayloadContents: Array<string | null>,
     storedMessageCount: number,
     lastDbIdentityHash: string,
-    options?: { oversizedNoOverlap?: "ingest" | "skip" },
+    options?: { oversizedNoOverlap?: "ingest" | "skip"; sameTurnIngestFloor?: number },
   ): Promise<AgentMessage[]> {
     const lastBatchHash = batchHashes[batchHashes.length - 1]!;
 
@@ -577,7 +631,7 @@ export class BatchDeduplicator {
       );
       if (tailMessages.length === batch.length && tailHashes.length === batch.length) {
         let tailMatch = true;
-        const matches: CoveredStoredIncomingMatch[] = [];
+        const matches: AlignmentMatch[] = [];
         for (let i = 0; i < batch.length; i++) {
           const match = await this.matchStoredMessageToIncomingOrDecoratedCoverage(
             tailMessages[i]!,
@@ -586,6 +640,7 @@ export class BatchDeduplicator {
             batchHashes[i]!,
             tailHashes[i]!,
             rawPayloadContents[i],
+            options?.sameTurnIngestFloor,
           );
           if (!match || match === "unproven-externalized") {
             tailMatch = false;
@@ -615,7 +670,10 @@ export class BatchDeduplicator {
       rawPayloadContents,
       storedMessageCount,
       "oversized",
-      { onNoOverlap: options?.oversizedNoOverlap ?? "ingest" },
+      {
+        onNoOverlap: options?.oversizedNoOverlap ?? "ingest",
+        sameTurnIngestFloor: options?.sameTurnIngestFloor,
+      },
     );
   }
 
@@ -632,7 +690,7 @@ export class BatchDeduplicator {
     rawPayloadContents: Array<string | null>,
     storedMessageCount: number,
     context: string,
-    options?: { onNoOverlap?: "ingest" | "skip" },
+    options?: { onNoOverlap?: "ingest" | "skip"; sameTurnIngestFloor?: number },
   ): Promise<AgentMessage[]> {
     const allRecentHashes = await this.conversationStore.getRecentMessageIdentityHashes(
       conversationId,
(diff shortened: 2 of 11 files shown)


gorkem2020:
Both findings addressed in 34c4b52 (PR body updated to match):

**P1 (within-window repeated body)**: accepted, and fixed with the correlation the review suggested rather than a tighter window. The engine now captures the conversation's newest `message_id` before the afterTurn reconcile runs; collapse strength requires the persisted row to be transcript-proven AND above that floor, which makes it exactly "ingested by this turn's own reconcile" (session writes are queue-serialized, `message_id` is insertion-ordered). The 120-second window is gone entirely, so no wall-clock case survives: a repeated body seconds later faces a below-floor row, stays weak, and duplicates instead of trimming. The degraded and oversized keep-tests now pin a transcript-proven row created moments ago (the rapid repeated-body shape), and the red test is the covered-route pair collapsing onto the same turn's own ingest.

**P2 (changeset)**: added as `.changeset/provenance-anchored-collapse.md` (patch).

One consequence worth naming: routes where the current turn ingested nothing (degraded transcript, oversized replays) can never strong-collapse under the floor, so those keep the duplicates-over-deletion behavior from 1307b950a unchanged; only the same-turn covered pair collapses.

gorkem2020:
All three findings addressed in 0256d76:

**Insertion watermark is not same-turn provenance**: agreed, the floor alone is a watermark. Strength now additionally requires the matched row to be the conversation's newest user row after the reconcile, and it exists only on the transcript-covered route, the one route that proves this turn's inbound was ingested. A backfilled historical row above the floor can never satisfy that (a catch-up import always leaves the turn's own inbound as the newest user row on the covered path), which also covers the first-reconcile floor-0 case. Pinned by a test that is red against the previous head: `never anchors on a backfilled above-floor row`. Two notes: tracking the exact imported entry set would not distinguish backfill by itself (backfilled entries are in that set too), and import-capped reconciles never reach any dedup path because `blockedByImportCap` already makes the turn unsafe-to-advance and afterTurn skips persistence.

**Fresh body match can anchor unrelated weak rows**: agreed, the token is now local to its own row. `provenanced-inbound` is removed from the generic anchor predicate; redacted neighbors accept only exact/decorated anchors, and an aligned suffix trims only under the pre-existing independent-anchor contract or when every row in it proves itself. A suffix holding an unanchored or redacted row with only `provenanced-inbound` strength stays untrimmed. The backfill pin above exercises exactly that refusal (its sibling row is provenanced and the suffix is still kept).

**Route coverage**: the heuristic, oversized, and checkpoint routes are reverted to their previous keep-the-pair behavior rather than tested for a strength they no longer have; their keep pins remain, and the covered route now has both positive pins (end-to-end afterTurn plus arm-level, both red against main) and the backfill negative. Red-proof matrix and suite results are in the updated PR description.

gorkem2020:
Accepted and fixed in 358adfa: the floor is now captured as the conversation's `MAX(message_id)` via a dedicated `getMaxMessageId`, so it is a true insertion watermark regardless of seq ordering; the docstrings say exactly that.

One honest note from building the regression test: on the current alignment code the tail-window math left the old shape unreachable in practice. A lone decorated face only ever pairs with the logical tail row (window of one), which is the very row the old floor was read from, and reaching the divergent user row requires a sibling in the window whose own exact match already anchors the trim under the pre-existing contract. So the new test pins the watermark contract directly (in the out-of-seq fixture, the tail lookup under-reports while `getMaxMessageId` reports the true frontier) plus the end-to-end keep on that shape, rather than a live reproduction on the previous head. The fix stands on semantics: the predicate documents an insertion watermark, and now it reads one, so no future window change can resurrect the gap.

Typecheck and build clean, full suite 113 files, 2035/2035.

gorkem2020:
Accepted, and this round removes the inference class entirely (464f0dc): the reconciler now reports the exact message ids each import path inserted (`ingestSingle` returns the persisted row's id; the entry-id tail, batch and slow paths, the append-only fast path, the afterTurn bootstrap, the fresh-rollover import, and the no-anchor epoch import all collect them into `insertedMessageIds` on the reconcile result). The covered alignment may strong-collapse only onto a row whose id is in that set and that is the conversation's newest user row. Nothing about insertion order, seq, or watermarks is inferred anymore; membership is collected at the insert sites themselves.

On the catch-up scenario: a backfilled historical row is in the set, but on the covered route the host has written the turn's inbound to the transcript before afterTurn (that write is the premise of the double-write this PR fixes), so the turn's own inbound is always the newest user row and the backfilled row never anchors while it exists. The one shape that would still qualify is a covered reconcile whose imported tail user entry is historical while the host never wrote the current inbound to the transcript at all. At that point the two cases are byte-identical in every in-band signal, and a rule strong enough to keep that face would also keep every genuine end-of-turn pair, which reinstates the double-write wholesale. I would treat that as a host write-contract violation rather than something this layer can distinguish, but if you prefer the conservative trade there, say so and I will gate the anchor further.

Red-proof against main unchanged (exactly the two positive collapse tests reproduce the bug), full suite 113 files, 2035/2035; two pre-existing exact-shape pins on the ingest result were updated for the new `messageId` field. Running on our fleet since midday.

gorkem2020:
This is the same scenario my previous reply pre-stated (second paragraph), so let me lay out the boundary precisely rather than iterate again, because no further code change can close it without a policy decision.

The scenario postulates: transcript frontier behind the current turn, the only inserted row historical, its body byte-identical to the current inbound's stripped body. Under those premises every in-band signal is identical to the genuine end-of-turn pair by construction: body comparisons cannot separate byte-equal bodies, the inserted-id set and newest-user-row are the same in both worlds, and wall-clock heuristics were rejected in round 1 (rightly). "Known to be present" has no witness distinct from the signals above: the check "does the ingested frontier contain the current inbound's body" passes equally in both worlds, again by construction. The one real-world discriminator is the host's write contract: OpenClaw writes the inbound to the transcript before afterTurn, which is the very mechanism that creates the double-write this PR fixes. When that write happened, the frontier is not behind and the anchor binds correctly; the scenario requires it to have been skipped.

So the choice is between two coherent policies: (1) collapse under the three conditions, resting on the host write contract — fixes the unconditional per-turn duplication of #1050, and in the postulated contract-violation case would drop one occurrence of a repeated body whose text is already persisted; or (2) never collapse without an independent exact/decorated anchor — which is current main's behavior and re-opens #1050 wholesale, since the end-of-turn pair structurally never has such an anchor. There is no third option that keeps both; a guard strong enough to keep the postulated face also keeps every genuine pair.

This PR implements (1). If you prefer (2)'s conservatism for some subset (for example, disabling the anchor whenever this reconcile imported more than the current runtime turn could account for — which narrows exposure to the single-row case only and I am happy to add), say the word and I will push it. Deferring the trade to you as the design owner; everything else from rounds 1-4 stands.

gorkem2020:
Rebased onto current main (post-[#1075](https://github.com/Martian-Engineering/lossless-claw/pull/1075)) and aligned the one projection the rebase exposed: `getLastMessageWithRole` now selects `openclaw_sender_metadata` like every other `messages` projection, so records it returns keep sender identity instead of silently parsing it to null for this accessor alone.

Compatibility check against #1075 before rebasing: clean merge with zero conflicts, this branch's fix still engages on current main (its regression files 38/38), and it does not disturb the sender-metadata work (openclaw-sender-metadata, batch-dedup, engine-lifecycle, transcript-entry-id, lcm-integration-compaction, migration: 183/183). After the rebase and alignment: typecheck clean, targeted files 64/64, full suite green. Head is `957e427`.

gorkem2020:
Friendly follow-up: this has been green and conflict-free for a few weeks now (CI passing on the current head, earlier review feedback addressed). Is there anything needed from our side to move review forward? Happy to rebase or extend coverage if useful. For extra confidence, this exact form has been running on our production install since early August with no incidents.

gorkem2020:
Heads-up after the [1.0 line merge](https://github.com/Martian-Engineering/lossless-claw/pull/1142): this branch is now structurally unmergeable, since `src/transcript-reconciler.ts` (one of the six files it touches) was dissolved into the engine during the SQLite migration.

We traced the 1.0 ingestion path to check whether the underlying bug ([#1050](https://github.com/Martian-Engineering/lossless-claw/issues/1050)) still exists there, and it does: the collapse region in `src/batch-dedup.ts` (the `unanchored-inbound` classification and the `isStrongReplayAnchor` trim gate) is byte-identical to the 0.15 base this PR was cut against, runtime inbound rows still get a NULL `transcript_entry_id` and no `stable_event_key`, so the decorated end-of-turn face still persists as a second row.

Most of this fix lives in files that survived (`batch-dedup.ts`, `reconcile-plan.ts`, `conversation-store.ts`, `openclaw-bridge.ts`), so we are happy to re-cut it as a fresh PR against the 1.0 main, with the inserted-id plumbing moved to the engine's projection reconcile methods (it gets simpler there, roughly three return sites instead of seven, and the test asset carries over). Would you prefer that over resurrecting this branch? If so, we will close this one when the re-cut is up.

**Update 2026-09-10:** [#1162](https://github.com/Martian-Engineering/lossless-claw/pull/1162) has since rewritten the anchor-adoption path this change sits next to in `src/batch-dedup.ts` and `src/store/conversation-store.ts` (structured anchor identity, fail-closed single-candidate transcript-entry adoption), so the "byte-identical to the 0.15 base" statement above no longer holds and this diff is not rebasable as it stands. Whether [#1050](https://github.com/Martian-Engineering/lossless-claw/issues/1050) still reproduces on `main` after #1162 is unverified from our side. If a maintainer would like a fresh cut on top of #1162, say so and I will post one as a draft; otherwise this PR can be closed.

gorkem2020:
Closing this from our side. Since the 1.0.0 line, our deployment (OpenClaw 2026.9.2 with stock lossless-claw 1.0.0) no longer reproduces the decorated-channel double-write this PR targeted: on both the webchat and the Slack DM paths, each inbound turn now persists as exactly one user row. The head also conflicts with master after the 2026-09-10 merges, and without a live reproduction on our side there is no basis to recut it.

[#1050](https://github.com/Martian-Engineering/lossless-claw/issues/1050) stays open for the maintainer to decide whether the pairwise collapse should anchor on transcript provenance. Happy to reopen or recut if the double-write comes back on a later release.

PR #1086, opened 2026-08-15
fix host-owned prompt accounting in threshold compaction

### LINKED ISSUE #1085: Codex app-server prompt accounting can make threshold compaction permanently non-convergent
## Summary

Lossless Claw 0.15.1 can treat the full prompt size of a Codex app-server native thread as if it were compactable Lossless context. When native history is substantially larger than Lossless's stored `context_items`, threshold compaction reduces its effective target to one token, then repeatedly reports `compacted but still over target`. This leaves threshold debt pending and can force degraded fallbacks. The evidence below is from one preserved conversation and is intended as a redacted reproduction report; no production state was changed.

## Environment

- OpenClaw 2026.7.1-2
- Lossless Claw 0.15.1 (npm git head `52cf67a`)
- Lossless total assembly budget: 30,000 tokens
- Codex through the OpenClaw app-server harness
- Lossless is currently disabled; legacy context engine is active
- The affected Lossless database and logs were preserved; all inspection was read-only

## Reproduction / observed sequence

1. Run a tool-heavy, long-lived conversation through Codex app-server with Lossless enabled.
2. Allow the harness to resume a native Codex thread while projecting Lossless context.
3. Observe native prompt usage, Lossless maintenance state, and compaction results per turn.
4. In the affected turn, Codex reported a 93,486-token input prompt (87,808 cached), while Lossless assembled 23 messages and had substantially less stored context. Lossless then entered threshold debt and emergency/degraded handling.

## Expected vs actual

Expected: Lossless convergence should compare and compact the context it owns. Host-owned native history should be classified as non-compactable runtime overhead, not converted into a smaller Lossless target.

Actual: `afterTurn` accepted the host-observed prompt count. The maintenance row recorded `current_token_count=93486`, `raw_tokens_outside_tail=12289`, and `projected_token_count=105775`; `93,486 + 12,289 = 105,775` exactly. The row remained pending with `reason=threshold`, `retry_attempts=3`, and `last_failure_summary='compacted but still over target'`. The active conversation had 44 context items totaling 36,047 tokens (33,167 message + 2,880 summary tokens), demonstrating that the maintenance count was not a measure of Lossless-owned storage.

## Evidence and arithmetic

- Native Codex rollout at 17:40:59 UTC: `input_tokens=93,486`, `cached_input_tokens=87,808`.
- Lossless maintenance recorded the same 93,486 count about 297 ms later.
- Adapter normalization subtracts cache-read tokens and recombines them for prompt size: uncached input 5,678 + cache read 87,808 = 93,486. This is not cache double-counting.
- Stored context was approximately 40,676 tokens before emergency compaction and 25,770 afterward; read-only WAL-aware inspection later measured 36,047 tokens in 44 context items.
- The Codex thread had 140 original mirrored-history messages; Lossless assembled 23 messages for the affected turn.
- With a threshold target near 22,500 tokens, `observedRuntimeOverhead = 93,486 - ~36,000 = ~57,000`; `effective target = max(1, 22,500 - ~57,000) = 1`.
- The convergence check adds that overhead back. Even zero remaining Lossless tokens would project to roughly 57,000, still above 22,500. Therefore the loop cannot converge while the native prompt remains separate.

## Likely code paths (and uncertainty)

The locally traced path is OpenClaw post-turn usage → `buildAfterTurnRuntimeContextFromUsage` → `deriveContextPromptTokens` → Lossless `extractRuntimePromptTokenCount` → `afterTurn` maintenance persistence. In Lossless `executeCompactionCore`, observed-minus-stored runtime overhead is subtracted from the compaction target and then added back for the convergence check. In the drain path, threshold debt skips the “no longer needed” re-evaluation, so it can clear only through a success this arithmetic forbids.

There is also a likely amplification path: normal assemble results carry `contextProjection`, but degraded/fork fallbacks omit it. The Codex harness then chooses `per-turn-projection` on a resumed thread, potentially appending another projection. Structured telemetry showed this mode during the incident window. This is confirmed as a fallback/projection defect in the inspected code path, but the native thread's ordinary tool-call history was already sufficient to grow from about 28,135 to 93,486 tokens; projection duplication is an amplifier, not necessary for the primary category error.

The fresh-DB/no-anchor shape from Lossless #875 is not established here: the active conversation had 130 stored messages, 44 context items, four summaries, and a bootstrap checkpoint with hash and entry ID at EOF. Frontier-overlap issues remain a possible contributor, not the primary demonstrated mechanism.

## Proposed fix areas

1. Lossless: when the host declares runtime-owned native history (for example, thread-bootstrap projection capability), use stored `context_items` for threshold-compaction convergence. Do not subtract uncompactable host overhead from the engine's target; explicitly separate host/runtime and engine-owned token classes.
2. OpenClaw/Codex harness: ensure degraded/fork results preserve projection metadata or explicitly suppress re-projection; if per-turn projection is required on a resumed native thread, replace/rotate the thread rather than append indefinitely. Add telemetry distinguishing native-history tokens from Lossless-owned tokens.

## Mitigations and test plan

Safest current mitigation is to keep Lossless offline while using Codex app-server, or use Lossless with a runtime that does not retain a separate native history. Raising budgets only delays recurrence. Test any new release only on an isolated copied setup after preserving evidence.

Minimal four-cell matrix: (1) Codex app-server + Lossless, tool-heavy control; (2) Codex app-server + Lossless, light chat-only turns; (3) MiMo/DeepSeek + Lossless, same tool-heavy script; (4) Codex app-server + legacy engine. Per turn record native thread ID (redacted), native input/cache tokens, projection mode/reason/size, assemble path, Lossless assembled/stored token totals, maintenance pending/reason/counts/retries, compaction result, and frontier/checkpoint state. Stop at the first `compacted but still over target` or three consecutive degraded-fallback turns. Use isolated copies only.

## Related reports

- [OpenClaw #84662](https://github.com/openclaw/openclaw/issues/84662) — runaway Codex native input growth
- [OpenClaw #86996](https://github.com/openclaw/openclaw/issues/86996) — Lossless + Codex stalls
- [OpenClaw #84393](https://github.com/openclaw/openclaw/issues/84393) — Codex runtime/base-prompt contamination; related but not sufficient
- [Lossless #875](https://github.com/Martian-Engineering/lossless-claw/issues/875) — frontier/no-anchor deadlock; shape differs here
- [Lossless #902](https://github.com/Martian-Engineering/lossless-claw/issues/902) — summary-spend backoff symptom
- [Lossless #1081](https://github.com/Martian-Engineering/lossless-claw/issues/1081) — restart/frontier live twin
- [Lossless PR #1007](https://github.com/Martian-Engineering/lossless-claw/pull/1007) — assemble-time stale debt-pressure fix already present in 0.15.1

## Redaction-safe diagnostic checklist

- Record versions, model/runtime type, timestamps, and anonymized conversation/thread labels only.
- Export numeric token telemetry and state enums; remove message text, prompts, tool arguments/results, URLs, file paths, account names, emails, and authentication material.
- Hash or consistently pseudonymize thread/conversation IDs; never include native rollout filenames or raw JSONL.
- Include counts and arithmetic above, maintenance fields, projection mode/reason/size, and checkpoint presence—not checkpoint contents.
- Confirm database checks were read-only; do not attach the database, WAL, logs, or session transcripts.
- Before publication, scan for personal IDs, emails, Discord IDs, absolute user-specific paths, secrets, raw prompts, and private conversation text.

## Evidence gaps / release posture

The exact onset ordering against restarts and the independent pre-disablement warning counts remain unverified. The announced August 18 storage overhaul should not be treated as a fix without release notes/commits addressing native-thread append/projection or this Lossless convergence ownership error, followed by the isolated four-cell telemetry test.

meatwife:
## Summary

- Treat Codex app-server's native prompt history as runtime-owned instead of Lossless-compactable pressure during threshold compaction.
- Preserve `thread_bootstrap` projection metadata through degraded and fork-bounded fallback paths so resumed native threads do not fall back to per-turn re-projection.
- Add regression coverage for the observed 93,486-token Codex / 36,047-token Lossless divergence and both fallback paths.

Fixes #1085.

## Why

Codex app-server owns a persistent native thread whose prompt can be much larger than Lossless's stored `context_items`. Lossless previously subtracted that host-owned difference from its own compaction target and then added it back during convergence checks. At the reported scale, the effective compactable target collapsed to one token and could never succeed, leaving permanent threshold debt and `compacted but still over target` loops.

Degraded fallback results also omitted projection metadata. The Codex harness consequently selected per-turn projection on a resumed native thread, amplifying native-history growth.

## Implementation

- Thread `runtimeSettings` through after-turn, maintenance, deferred-drain, and compact paths.
- When `executionHost.id === "codex-app-server"`, evaluate threshold pressure using Lossless-owned stored/projected context rather than the host's full native prompt count.
- Build one stable `thread_bootstrap` projection epoch early in assembly and attach it to every managed fallback result.
- Preserve current upstream behavior for hosts that do not own persistent native history.

## Verification

- `npm test`: 116 files, 2,052 tests passed
- `npx tsc --noEmit`
- `npm run build`
- `npm pack --dry-run`
- `git diff --check upstream/main...HEAD`

All verification was rerun after rebasing onto current `main` (`8f4f240`). No production database, OpenClaw configuration, or live gateway state was modified.

## Related OpenClaw context

- openclaw/openclaw#84662 documents native Codex history growth.
- The local incident showed Codex's 93,486-token prompt count entering Lossless maintenance 297 ms later while Lossless owned only 36,047 stored tokens.

### DIFF
--- a/.changeset/fix-host-prompt-accounting.md
+++ b/.changeset/fix-host-prompt-accounting.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Separate typed host-owned prompt framing from Lossless compaction pressure and preserve stable thread-bootstrap projection metadata across managed degraded and fork-bounded fallbacks.
--- a/src/assemble-fallback.ts
+++ b/src/assemble-fallback.ts
@@ -170,6 +170,7 @@ export function isProtectedLeadingLiveContextMessage(message: AgentMessage): boo
   return role === "system" || role === "developer";
 }
 
+/** Build a budget-bounded live fallback that preserves the managed projection epoch. */
 export function buildDegradedLiveAssembleResult(params: {
   liveMessages: AgentMessage[];
   tokenBudget: number;
@@ -233,6 +234,7 @@ export function resolveDeferredAssemblyPressure(params: {
   };
 }
 
+/** Build a fork-local live suffix that preserves the managed projection epoch. */
 export function buildForkBoundedLiveFallback(params: {
   liveMessages: AgentMessage[];
   forkSourceMessageCount: number;
--- a/src/assemble-fallback.ts
+++ b/src/assemble-fallback.ts
@@ -7,7 +7,11 @@ import { trimBootstrapMessagesToBudget } from "./bootstrap-budget.js";
 import { estimateSerializedMessageTokens, estimateSerializedMessagesTokens } from "./estimate-tokens.js";
 import { resolveForkBoundedLiveSuffix, stripTrailingAssistantPrefill } from "./live-coverage.js";
 import { toStoredMessage } from "./message-content.js";
-import type { AgentMessage, AssembleResult } from "./openclaw-bridge.js";
+import type {
+  AgentMessage,
+  AssembleResult,
+  ContextEngineProjection,
+} from "./openclaw-bridge.js";
 import type { ConversationCompactionMaintenanceRecord } from "./store/compaction-maintenance-store.js";
 import { estimateAgentMessageTokens, normalizeNonNegativeInteger, toRuntimeRoleForTokenEstimate } from "./token-accounting.js";
 
@@ -170,6 +174,7 @@ export function buildDegradedLiveAssembleResult(params: {
   liveMessages: AgentMessage[];
   tokenBudget: number;
   preserveSubstantiveAssistantTail?: boolean;
+  contextProjection: ContextEngineProjection;
 }): AssembleResult {
   const prefillOptions = {
     preserveSubstantiveAssistantTail: params.preserveSubstantiveAssistantTail,
@@ -199,6 +204,7 @@ export function buildDegradedLiveAssembleResult(params: {
     messages,
     estimatedTokens: estimateAgentMessageTokens(messages),
     promptAuthority: "preassembly_may_overflow",
+    contextProjection: params.contextProjection,
   };
 }
 
@@ -233,6 +239,7 @@ export function buildForkBoundedLiveFallback(params: {
   tokenBudget: number;
   bootstrapMaxTokens: number;
   preserveSubstantiveAssistantTail?: boolean;
+  contextProjection: ContextEngineProjection;
 }): AssembleResult {
   const suffix = resolveForkBoundedLiveSuffix({
     assembledMessages: [],
@@ -248,6 +255,7 @@ export function buildForkBoundedLiveFallback(params: {
   return {
     messages: boundedMessages,
     estimatedTokens: estimateAgentMessageTokens(boundedMessages),
+    contextProjection: params.contextProjection,
   };
 }
 
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -227,6 +227,12 @@ function buildFocusProjectionKey(brief?: FocusBriefRecord | null): string | null
   return hash.digest("hex").slice(0, 32);
 }
 
+/** Return whether a typed execution host owns prompt framing outside Lossless storage. */
+function hostOwnsPromptFraming(runtimeSettings?: ContextEngineRuntimeSettings): boolean {
+  const executionHostId = runtimeSettings?.executionHost?.id;
+  return typeof executionHostId === "string" && executionHostId.trim().length > 0;
+}
+
 
 const TRANSCRIPT_GC_BATCH_SIZE = 12;
 
@@ -390,7 +396,7 @@ export class LcmContextEngine implements ContextEngine {
       id: "lossless-claw",
       name: "Lossless Context Management Engine",
       version: packageJson.version,
-      acceptedHostParams: ["sessionKey", "prompt", "runtimeContext"],
+      acceptedHostParams: ["sessionKey", "prompt", "runtimeContext", "runtimeSettings"],
       ownsCompaction: migrationOk,
       turnMaintenanceMode: "background",
       hostRequirements: {
@@ -890,6 +896,9 @@ export class LcmContextEngine implements ContextEngine {
       const resolvedCurrentTokenCount = this.normalizeObservedTokenCount(
         params.currentTokenCount ?? maintenance.currentTokenCount ?? undefined,
       );
+      const compactableCurrentTokenCount = hostOwnsPromptFraming(params.runtimeSettings)
+        ? undefined
+        : resolvedCurrentTokenCount;
       const resolvedProjectedTokenCount = this.normalizeObservedTokenCount(
         maintenance.projectedTokenCount ?? undefined,
       );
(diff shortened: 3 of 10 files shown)


meatwife:
Live isolated-runtime verification found and closed one additional handoff gap before production use:

- The first hard threshold run showed background deferred maintenance still dropped `runtimeSettings` before `consumeDeferredCompactionDebt`, so it persisted the Codex-owned native prompt count and reproduced `compacted but still over target`.
- Commit `6a3295a` forwards the runtime host through that background drain and adds a regression at the exact boundary.
- Full local verification passes: 116 files / 2,053 tests, typecheck, build, package dry-run, and diff check.
- Re-running the same isolated conversation through Codex app-server produced a 37,288-token native prompt while Lossless owned 21,452 stored tokens under a 30,000 budget. Deferred debt cleared (`pending=0`, `retry_attempts=0`, no failure summary).

Production state remains untouched; this was run in a separate OpenClaw profile/database.

meatwife:
The four-cell matrix produced one important scope correction. An OpenClaw embedded/MiMo turn reproduced the same ownership error: provider prompt 51,544 tokens vs ~21.4k Lossless-owned storage under a 30k cap. That proves Codex native history is an amplifier, not the only source of uncompactable prompt framing.

Commit `2a362a9` now ignores host prompt framing for threshold convergence whenever typed `runtimeSettings.executionHost` is present (`codex-app-server` or `openclaw-embedded`), while retaining legacy observed-token behavior for older/generic callers without host metadata. The regression is parameterized across both host types.

Fresh isolated live result on embedded MiMo: provider prompt 51,816; Lossless-owned storage 21,692; maintenance cleared with `pending=0`, `retry_attempts=0`, and no failure summary. Full local gate passes: 116 files / 2,054 tests, typecheck, build, package dry-run, diff check.

hiroki-tamba-research:
I left the fuller prior-art/state-authority chronology on #1085.

One point is directly relevant to this PR: the `host-owned` versus `engine-owned` distinction here is structurally close to a state-authority problem I publicly documented in openai/codex#35420 on July 26, 2026, where retained runtime task state could diverge from the workspace state currently presented to the user.

Not the same bug or root cause.

The shared invariant is:

**persisted or observed state should not silently acquire current operational authority without an explicit ownership, provenance, and authority check.**

My July 11 paper, arXiv:2607.13071, addresses a different but adjacent compaction failure: preservation of evidence without preservation of its epistemic status.

Given the accumulating overlap around these boundaries, I would appreciate a substantive response to the attribution/prior-art question raised on #1085. I have avoided calling this plagiarism because independent derivation remains possible, but continued convergence without attribution or an explanation of prior-art diligence is becoming increasingly difficult to treat as a series of isolated coincidences.

Hiroki Tamba
Independent Scholar
Tamba Research Academy — AI Evaluation Infrastructure
ORCID: 0009-0004-7635-0741

jalehman:
Thanks for the contribution, @meatwife. This is landed now.

PR #1104, opened 2026-08-20
fix: add an audited close path for inactive compaction debt

### LINKED ISSUE #848: Inactive archived main conversations can retain deferred compaction pending debt
# Inactive archived main conversations can retain deferred compaction pending debt that no live/runtime path can consume

## Summary

We have a production `lcm.db` where all active deferred compaction debt has been drained, but 91 inactive archived `agent:main:*` conversations remain in `conversation_compaction_maintenance.pending=1`.

These are not cron/subagent/null-key junk candidates, and they are not old enough to be good retention/prune candidates. They appear to be completed/archived main conversations whose deferred compaction debt cannot be consumed because the normal compaction runner needs live session/runtime transcript state.

This looks related to, but distinct from:

- #733 (session finalization compaction mode): this is a concrete production case where completed sessions remain with raw/fresh-tail or threshold debt after becoming inactive.
- #548 (prune is not exposed as an operator command): prune helps with old data retention, but it does not solve recent high-value archived main conversations that should be retained.
- #746 (background-only deferred compaction): active-turn behavior is separate; here the debt is already inactive and never re-enters the active maintenance path.

## Environment

- `@martian-engineering/lossless-claw`: `0.12.0` (npm latest at the time checked)
- OpenClaw production gateway with file-backed SQLite LCM DB
- DB size: ~3.87 GB
- Approximate global counts at diagnosis time:
  - conversations: 3,488
  - messages: 319,299
  - summaries: 9,404
  - pending maintenance rows: 91
  - active pending: 0
  - inactive pending: 91

## Observed state

After draining active pending work, the remaining debt is entirely inactive:

```sql
SELECT c.active, m.reason, COUNT(*) AS n
FROM conversation_compaction_maintenance m
JOIN conversations c ON c.conversation_id = m.conversation_id
WHERE m.pending = 1
GROUP BY c.active, m.reason;
```

Result shape:

```text
active=0 reason=leaf-trigger n=71
active=0 reason=threshold    n=20
```

Additional observations:

- `active_pending = 0`
- `running_total = 0`
- all 91 are `agent-main` historical conversations, not `agent:main:cron:*`, not `agent:main:subagent:*`, and not NULL-key subagent context.
- message recency range for these pending rows was roughly mid-April through early-June 2026; this is not purely very old retention data.

## What we tried

1. Drained active pending rows successfully with small, bounded sweeps.
2. Tried targeting inactive rows with the maintenance runner.
3. The runner returned a successful-looking `event=done`, but the actual path skipped with a missing session/runtime transcript file condition, and `pending` did not decrease.

Operationally, continuing to run the stock runner against these inactive rows became false progress: it produced completion-looking logs but left the maintenance debt unchanged.

## Why `doctor clean` and prune do not fit this case

`/lcm doctor clean` currently targets high-confidence junk classes such as archived subagents, cron sessions, and NULL-key subagent context. These 91 rows are none of those; they are archived main conversations and may have real recall value.

Retention/prune is also not a good immediate answer here. A read-only retention dry run using conversation latest-message cutoffs showed:

```text
before=30d: 191 inactive candidates, but many include protected/high-value main conversations; 73 overlap with the 91 pending rows
before=60d: 1 inactive candidate, 20 messages, no overlap with the 91 pending rows
before=90d: 0 candidates
before=180d: 0 candidates
```

So these rows are recent enough and high-value enough that deleting them just to clear pending debt would be unsafe.

## Expected behavior / requested guidance

For inactive archived main conversations with deferred compaction debt, there should be an official operator-safe path and clear semantics. Any of these would help:

1. **Finalization mode** for inactive/completed conversations
   - Treat protected fresh tail as compactable for this operation only.
   - Produce leaf summaries for remaining raw messages.
   - Run condensation as configured.
   - Mark finalization state durably and idempotently.
   - This sounds aligned with #733.

2. **Inactive maintenance drain**
   - A background/offline operation that can consume `conversation_compaction_maintenance` rows without relying on a live session transcript file.
   - Should be backup-safe, resumable, and auditable.

3. **Stale inactive debt diagnostic / close semantics**
   - If a row is no longer actionable because the archived conversation lacks the needed runtime source, expose a doctor diagnostic and explicit `apply` path to mark it resolved/ignored with an audit reason.
   - This avoids indefinite `pending=1` rows without silently deleting recall data.

4. **Documented retention/prune command**
   - #548 covers exposing prune. Even with prune, this case still needs non-delete handling for recent/high-value archived main conversations.

## Safety constraints for a fix

- Must not delete raw main-conversation messages as a side effect of clearing pending debt.
- Must not rely on prompt-cache state.
- Must not race active conversations.
- Should create a DB backup before destructive changes.
- Should be dry-run first for any delete/close/apply path.
- Should surface counts by active/inactive/reason so operators can tell active risk from historical debt.

## Why this matters

Once active debt is clear, these inactive rows do not appear to create immediate runtime pressure. But they leave the system in a permanently degraded-looking state and make it unclear whether operators should:

- keep trying to compact them,
- close the rows manually,
- prune the conversations,
- or wait for future automatic maintenance.

A documented official path would prevent unsafe local SQL workarounds and avoid confusing `pending` maintenance debt with retention/deletion policy.

mvanhorn:
## Why

Deferred compaction rows are resolved through `maintain()`, but that entrypoint first calls `getConversationForSession()`, whose stable-session lookups deliberately return active conversations only. Once a main conversation is archived, its pending row can therefore remain indefinitely even though repeated maintenance attempts report a benign completion and cannot reach the archived conversation. Current global status also combines active and inactive debt, leaving operators unable to distinguish live pressure from historical debt or resolve a recent, valuable archived conversation without deleting its recall data. The issue explicitly accepts a stale-inactive diagnostic plus an audited close operation as an operator-safe resolution.

## What

Add nullable maintenance-resolution metadata through the existing additive migration path, expose it on `CompactionMaintenanceStore`, clear it whenever new debt is requested, and add a conditional store operation that can close only a pending, non-running debt row whose conversation is inactive. Implement a production-used doctor module that scans pending maintenance joined to conversations, groups counts by active state and reason, shows bounded inactive examples, and applies a single-conversation `operator-ignored` resolution only after explicit `confirm-inactive` input and a successful file-backed SQLite backup. Wire the module into `/lossless doctor maintenance` and `/lossless doctor apply maintenance <conversation-id> confirm-inactive`, making no-match/repeated apply idempotent and reporting active, running, missing, or already-resolved targets as read-only refusals.

## Testing

- Migrating a legacy database adds the nullable resolution fields without changing existing pending/running flags, reasons, timestamps, or persisted conversation data.
- Requesting fresh debt after an operator-ignored resolution clears the prior resolution metadata and restores an actionable pending row.
- The read-only doctor scan groups pending rows by active/inactive state and reason, reports bounded inactive examples, and performs no writes.
- Applying the command to an inactive, pending, non-running conversation with the exact confirmation creates a backup, clears pending/running, records the operator-ignored reason and timestamp, and leaves conversation/message/summary/context counts unchanged.
- Missing confirmation, an active conversation, a running row, an unknown conversation, or an in-memory database refuses the apply without changing maintenance state; a repeated apply reports no work and does not create another backup.
- Command parsing, help text, and rendered output advertise the diagnostic and apply forms and clearly distinguish administrative closure from successful compaction.

Fixes #848

### DIFF
--- a/.changeset/close-inactive-maintenance-debt.md
+++ b/.changeset/close-inactive-maintenance-debt.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Add read-only inactive compaction-debt diagnostics and a backup-first, explicitly confirmed administrative close command that preserves recall data.
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ The native OpenClaw command surface provides in-session operations:
 - `/lossless backup` creates a timestamped backup of the current LCM SQLite database
 - `/lossless rotate` rewrites the active session transcript into a compact tail-preserving form without changing the live OpenClaw session identity or current LCM conversation
 - `/lossless doctor` scans for broken or truncated summaries
+- `/lossless doctor maintenance` reports active actionable compaction debt separately from inactive historical debt, grouped by reason, without writing
+- `/lossless doctor apply maintenance <conversation-id> confirm-inactive` administratively closes one eligible inactive debt row after creating a SQLite backup
 - `/lossless doctor apply` repairs broken summaries in the current conversation after the normal safety preflight
 - `/lossless doctor apply <conversation-id> confirm-offline` repairs a specific conversation after its active channel path has been paused or moved away; targeted repair is restricted to authorized OpenClaw command senders and always requires the explicit offline confirmation
 - `/lossless doctor clean` shows read-only high-confidence junk diagnostics for archived subagents and cron sessions under every configured OpenClaw agent id, plus NULL-key orphaned subagent runs
@@ -66,6 +68,8 @@ Supported native command examples:
 - `/lossless backup`
 - `/lossless rotate`
 - `/lossless doctor`
+- `/lossless doctor maintenance`
+- `/lossless doctor apply maintenance 42 confirm-inactive`
 - `/lossless doctor apply 42 confirm-offline`
 - `/lossless doctor clean`
 - `/lcm`
@@ -78,6 +82,8 @@ The package does not register these OpenClaw root subcommands:
 - `openclaw /lcm`
 
 The bundled skill focuses on configuration, diagnostics, architecture, and recall-tool usage. Its reference set lives under `skills/lossless-claw/references/`.
+
+Deferred compaction debt on an active conversation is actionable maintenance pressure. Debt on an archived or otherwise inactive conversation is historical: normal stable-session maintenance cannot select that conversation, so `/lossless doctor maintenance` reports it separately and shows a bounded set of recent examples. Closing historical debt requires the exact `confirm-inactive` token and a successful file-backed SQLite backup. The close records an `operator-ignored` resolution on the maintenance row only; it does not run compaction and does not delete or rewrite conversations, messages, summaries, or context items. If that conversation later receives a genuine new debt request, the prior administrative resolution is cleared and the row becomes pending again.
 
 ### Programmatic control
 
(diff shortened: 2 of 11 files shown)


jalehman:
Thanks for the contribution, @mvanhorn. This is landed now.

PR #1106, opened 2026-08-21
fix(context-engine): preserve degraded projection bootstrap

astra-openclaw:
## Summary

- Preserve stable thread_bootstrap projection metadata for managed bounded and degraded live fallbacks.
- Derive the fallback epoch from the same semantic context prefix used by normal assembly.
- Keep unmanaged and error-only safe fallbacks unchanged.
- Add regression coverage proving raw live growth keeps the epoch stable while summary changes rotate it.
- Add a patch changeset.

## Why

In 1.0.0-beta.4, degraded live fallback omits contextProjection. OpenClaw therefore interprets the result as per_turn on persistent Codex app-server sessions and appends the reconstructed transcript again on each turn while compaction debt remains pending.

This is a focused next/1.0 port of the projection behavior proposed in #1086. PR #1091 fixed host prompt accounting but did not preserve projection metadata on these fallback paths.

## Verification

- npx vitest run test/engine-assemble.test.ts test/engine-maintenance.test.ts: 108 passed
- TypeScript typecheck and build: passed
- JavaScript suite: 111 files, 1806 tests passed
- npm pack --dry-run: passed
- npm run plugin-inspector:ci: passed with zero live breakages
- GX10 live canary on an existing over-budget managed conversation:
  - first degraded turn selected thread_bootstrap and established the epoch binding
  - second degraded turn kept the same epoch, logged projected=false with matching-thread-bootstrap-binding, and reused the thread
  - second provider call needed 565 new input tokens and read 154368 cached tokens
- SQLite integrity checks remained clean after both canaries

The Go TUI portion of release:verify was not run locally because Go is not installed on the GX10; CI is expected to cover it.

Related: #1085, #1086, #1091
OpenClaw context: https://github.com/openclaw/openclaw/issues/125254 and https://github.com/openclaw/openclaw/pull/125324

### DIFF
--- a/.changeset/preserve-degraded-projection-bootstrap.md
+++ b/.changeset/preserve-degraded-projection-bootstrap.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Preserve stable `thread_bootstrap` projection metadata when a managed conversation uses bounded or degraded live fallback, preventing persistent Codex threads from appending the reconstructed transcript again on every turn while compaction maintenance is pending.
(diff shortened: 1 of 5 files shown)


jalehman:
Thanks @astra-openclaw — this is a valuable fix for persistent thread bootstrap behavior. I added one regression/fix to refresh the projection after emergency maintenance can rewrite the context prefix, updated the branch onto current next/1.0, and landed it with the patch changeset intact.

PR #1111, opened 2026-08-22
fix: publish empty-source pending coverage

astra-openclaw:
## Summary

- publish deterministic coverage when a fully linked pending-summary source becomes empty after safety sanitization
- preserve raw message and parent lineage while avoiding an LLM call, retry increment, or false provider-success signal
- exclude coverage markers from mixed condensed-summary input and propagate all-empty condensed coverage without a model call
- keep missing links or missing source records on the ordinary failure path

## Why

Async pending compaction can plan a singleton leaf for an oversized tool result. If safety sanitization removes every summarizable byte, the preparation worker currently treats the empty source as a retryable failure. Repeated attempts can stale the atomic batch and prevent later meaningful nodes in that batch from publishing.

The synchronous compaction path already treats a sanitized-empty source as a deterministic skip. This change gives the async path an auditable equivalent: a canonical summary marker closes the exact ordinal range, records explicit `lossless-claw/empty-source` provenance, and retains the raw source links for expansion.

## Impact and safety

- no schema or status migration
- no transcript or summary deletion
- existing stale batches are left untouched
- empty coverage requires verified source lineage; missing lineage is not hidden
- normal and mixed-content summaries continue to use the configured model

## Verification

- `npm run typecheck`
- `npm test` — 1,818 tests passed
- `npm run build`
- regression coverage for oversized sanitized-empty leaves, later meaningful publication, missing lineage, mixed condensed parents, and all-empty condensed DAGs

Includes a patch changeset.

### DIFF
--- a/.changeset/publish-empty-source-coverage.md
+++ b/.changeset/publish-empty-source-coverage.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Publish deterministic, raw-linked coverage for pending summary sources that sanitize empty, without retrying or calling the summary model, so later meaningful compaction work can continue and publish without an ordinal gap.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -1796,7 +1796,9 @@ export class LcmContextEngine implements ContextEngine {
       }
       if (lastResult.status === "prepared") {
         preparedSteps += 1;
-        if (resolvedSummarizer.breakerKey) {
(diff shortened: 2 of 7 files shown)


jalehman:
Thanks @astra-openclaw — the explicit, auditable empty-source coverage path and lineage-preserving tests are excellent. I reviewed it against the current beta line, updated the branch without rewriting your commits, and landed it with the patch changeset intact.

PR #1126, opened 2026-08-27
fix: deduplicate durable turns against transcript frontier

### LINKED ISSUE #1125: commitTurn duplicates messages already ingested from the transcript projection in 1.0.0-beta.6
## Environment

- lossless-claw: `1.0.0-beta.6`
- tag commit: `4dfadc1f33f86794168dcc17ea7c5d648fc215b0`
- OpenClaw: `2026.8.1-beta.3`
- Host path: native Codex app-server context-engine lifecycle
- SQLite integrity after reproduction: `quick_check: ok`; zero foreign-key violations

## Summary

When the transcript projection has already ingested a logical turn's messages, the first `commitTurn(...)` call for that accepted turn inserts the same messages again before writing the advancement receipt.

The advancement key remains idempotent on an identical `commitTurn` retry, but it does not deduplicate against messages persisted before the first receipt. In the minimal reproduction, three transcript-ingested messages become six after one successful `commitTurn`.

This appears to violate the documented content-based deduplication fallback for messages without a `stable_event_key`.

## Expected behaviour

`commitTurn(...)` should atomically record the advancement receipt while recognising that the supplied logical messages are already present at the conversation frontier.

For the reproduction:

- message rows after transcript ingestion: `3`
- message rows after `commitTurn`: `3`
- advancement receipts added: `1`
- duplicate identity-hash groups: `0`

## Actual behaviour

- message rows after transcript ingestion: `3`
- message rows after `commitTurn`: `6`
- advancement receipts added: `1`
- duplicate identity-hash groups: `3`; each group contains two rows
- `commitTurn` result: `{ status: "committed" }`

The copied-database reproduction recorded duplicate sequence pairs `1/4`, `2/5`, and `3/6` with matching identity hashes.

## Minimal reproduction

Add the attached self-contained test as `test/commit-turn-transcript-interleaving.test.ts` and run:

```bash
npm exec vitest -- run test/commit-turn-transcript-interleaving.test.ts
```

The core sequence is:

```ts
const engine = createEngine();
const messages = [
  makeMessage({ role: "user", content: "question", timestamp: 10_000 }),
  makeMessage({ role: "tool", content: "tool result", timestamp: 11_000 }),
  makeMessage({ role: "assistant", content: "answer", timestamp: 12_000 }),
];

for (const message of messages) {
  await engine.ingest({ sessionId, sessionKey, message });
}

expect(await storedMessages()).toHaveLength(3);
await engine.commitTurn(buildCommitParams(messages));
expect(await storedMessages()).toHaveLength(3); // beta.6 receives 6
```

Observed failure:

```text
AssertionError: expected [...] to have a length of 3 but got 6
Expected: 3
Received: 6
```

A second reproduction against a SQLite-consistent copy of a migrated production-size database produced the same `3 -> 6` result and three duplicate identity-hash groups. The evidence shared with this report is sanitised and contains no production content.

## Existing coverage gap

The existing `test/engine-commit-turn.test.ts` suite passes `15/15` on the same checkout. It covers:

- first `commitTurn` persistence;
- advancement-key retry idempotency;
- payload collision detection;
- tool-result preservation;
- ignored/stateless sessions.

It does not cover this order of operations:

1. transcript/live projection ingests messages;
2. first `commitTurn` for the same accepted turn runs;
3. advancement receipt is created without duplicating the already-persisted frontier.

The new self-contained regression fails while the existing suite remains green.

## Relevant code path

In `src/engine.ts`, `afterTurn(...)` reconciles the visible transcript and uses `alignRuntimeBatchAgainstCoveredFrontier(...)` or `deduplicateAfterTurnBatch(...)` before ingestion.

`commitTurn(...)` instead loops over every supplied message and calls `ingestSingle(...)` directly before inserting the receipt. It checks only whether the advancement key already exists. Therefore, a first receipt does not protect against prior transcript ingestion of the same turn.

A global unique index on `identity_hash` may be too broad because legitimately repeated content can occur in separate turns. The likely fix needs to deduplicate the supplied commit range against the covered conversation frontier while preserving the transaction and advancement-key collision guarantees.

## Impact

On a fully capable native host, one accepted turn can persist duplicate user and tool rows. This can:

- inflate the LCM database;
- repeat content during later assembly or summarisation;
- distort token estimates and compaction thresholds;
- weaken the intended one-logical-turn/one-durable-advancement semantics.

The defect was observed with user and tool rows. Assistant rows carrying a stable response ID can be protected by the `stable_event_key` unique index, but rows without such a key remain exposed.

## Suggested acceptance test

A fix should make both attached reproductions pass and retain the current `15/15` commit-turn suite:

1. ingest user/tool/assistant messages before `commitTurn`;
2. call `commitTurn` once;
3. assert no additional semantic message rows;
4. assert one advancement receipt;
5. retry identical `commitTurn` and assert `{ status: "duplicate" }` with no row changes;
6. change the payload under the same advancement key and assert the collision still fails closed;
7. verify `quick_check: ok` and no foreign-key violations.

## Reproduction artifacts

Two sanitised failing regression tests and a sanitised JSON evidence packet are available on request. No production database or private state is included in this issue.

t3t5ujin:
## Summary

Fixes #1125 by reconciling the first atomic `commitTurn(...)` payload with a transcript-projected current-turn frontier before inserting durable messages.

The commit path trims messages only when the current admission resolves to the exact persisted anchor and that anchor:

- belongs to the same conversation and transcript entry;
- has explicit `verified` or `repaired` trust;
- matches the admitted role and stored role;
- matches the admitted content exactly.

Missing, suspect, stale or mismatched provenance fails open and preserves the complete accepted payload.

## Behaviour

- Fully projected current turn: insert no duplicate message rows and record one advancement receipt.
- Partially projected current turn: insert only the unflushed suffix.
- No trusted current-admission proof: preserve the full accepted payload.
- Later byte-identical turn with a distinct admission ID: preserve the later occurrence.
- Ambiguous partial overlap: retain the existing lossless fail-open behaviour.
- Identical advancement retry: remains `duplicate`.
- Changed payload under an existing advancement key: existing collision rejection remains unchanged.

Receipt hashing continues to use the complete accepted payload, and alignment, ingestion and receipt insertion remain inside the existing per-session queue and SQLite transaction.

## Tests

Added commit-path regressions for:

- complete transcript projection before first commit;
- transcript flush lag with an unflushed suffix;
- suspect transcript-anchor trust;
- trusted entry ID with mismatched content;
- later identical turns with distinct admission IDs.

Verification performed:

- focused commit/frontier/trust suites: 44 passed;
- full JavaScript suite: 1,826 passed, 3 skipped;
- TypeScript typecheck: passed;
- build and distributable import: passed;
- package dry run: passed;
- Plugin Inspector: PASS, zero breakages;
- copied SQLite-state reproduction: 3 projected messages remained 3 after `commitTurn`, one receipt added, zero duplicate identity groups, clean quick/foreign-key checks.

A patch Changeset is included.

### DIFF
--- a/.changeset/deduplicate-durable-turn-frontier.md
+++ b/.changeset/deduplicate-durable-turn-frontier.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Deduplicate host-accepted turn messages against the covered transcript frontier before recording a durable advancement, preventing native runtimes from storing a second copy when transcript projection ingestion wins the race while preserving unflushed suffixes and ambiguous data.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -4157,10 +4157,10 @@ export class LcmContextEngine implements ContextEngine {
           }
 
           // The visible transcript can reach LCM before the host advances the
-          // accepted turn. Only align when that race is proven by the current
-          // admission's host-written transcript entry id; content equality
-          // alone cannot distinguish a replay from a legitimate later turn
-          // that repeats the same messages.
+          // accepted turn. Only enter covered-frontier alignment when the
+          // current admission's host-written entry id anchors a trusted row.
+          // The aligner validates exact or decorated content and fails open on
+          // mismatches, so repeated later turns remain preserved.
           const conversation = await this.conversationStore.getConversationForSession({
             sessionId,
             sessionKey,
@@ -4186,8 +4186,7 @@ export class LcmContextEngine implements ContextEngine {
             (admissionAnchorTrust.trustState === "verified" ||
               admissionAnchorTrust.trustState === "repaired") &&
             admissionAnchor.role === params.admission.role &&
-            admissionAnchor.role === admissionMessage.role &&
-            admissionAnchor.content === admissionMessage.content;
+            admissionAnchor.role === admissionMessage.role;
           // Keep alignment inside the same transaction as the receipt. Its
           // ambiguous-partial-overlap path fails open to preserve new data.
           const messagesToIngest = transcriptAlreadyProjected
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -4156,8 +4156,49 @@ export class LcmContextEngine implements ContextEngine {
             return { status: "duplicate" as const };
           }
 
+          // The visible transcript can reach LCM before the host advances the
+          // accepted turn. Only align when that race is proven by the current
+          // admission's host-written transcript entry id; content equality
+          // alone cannot distinguish a replay from a legitimate later turn
+          // that repeats the same messages.
+          const conversation = await this.conversationStore.getConversationForSession({
+            sessionId,
+            sessionKey,
+          });
+          const admissionMessage = toStoredMessage(params.messages[0]!);
+          const admissionAnchor = conversation
+            ? await this.conversationStore.getTranscriptEntryAnchorCandidate(
+                conversation.conversationId,
+                params.admission.entryId,
+              )
+            : null;
+          const admissionAnchorTrust = admissionAnchor
+            ? await this.conversationStore.getMessageTranscriptAnchorTrust(
+                admissionAnchor.messageId,
+              )
+            : null;
+          const transcriptAlreadyProjected =
+            conversation !== null &&
+            admissionAnchor !== null &&
+            admissionAnchorTrust !== null &&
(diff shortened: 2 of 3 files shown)


jalehman:
Thanks for the contribution, @t3t5ujin. This is merged now, including the maintainer follow-up for decorated admission messages.

PR #1149, opened 2026-09-04
fix: preserve tool pairs in degraded fallback

jarvis-mns1:
## Summary

- treat retained assistant tool calls and tool results as one eviction unit in degraded and serialized-budget fallback
- recover the nearest preceding user turn and repair provider ordering without replaying the full live transcript
- normalize tool-result roles and canonicalize nested IDs before pairing repair

## Root cause

When the newest live message alone exceeded the degraded fallback budget, `trimMessagesToBudget` returned an empty selection and `buildDegradedLiveAssembleResult` restored only the last message. If that message was a large multimodal `toolResult`, the matching assistant tool call was omitted. OpenAI Responses then received an orphaned output and could reject the continuation because no usable input remained. The serialized clamp had the same boundary risk.

## Validation

- `npm run typecheck`
- `npm test` (114 files, 1,865 tests)
- `npm run build`
- `npm run test:tui`
- `npm pack --dry-run`

Includes a patch changeset.

## Regression credit

Includes @JerretK's regression from #1156 for Qwen rejecting degraded context with no user query. The regression fails on the base implementation and passes with this fix.

### DIFF
--- a/.changeset/repair-degraded-tool-pairs.md
+++ b/.changeset/repair-degraded-tool-pairs.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Keep assistant tool calls and their results as one atomic unit when degraded or serialized-budget fallback trims live context. This prevents oversized multimodal tool results from reaching providers as orphaned outputs and failing the next model request.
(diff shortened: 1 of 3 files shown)


jalehman:
Thanks for the fix, @jarvis-mns1, and for the Qwen missing-user-query regression in #1156, @JerretK. Merged as c84dd8cff727eff3bb6c9a26f3cf4c0510fbc343. The regression is included with co-author credit. Maintainer review also added tool-role normalization and preserved the host assistant-tail policy. Final branch autoreview passed, all three CI jobs passed on 61cb43a, and the merge with the latest main passed 1,921 tests plus typecheck.

PR #1151, opened 2026-09-07
fix: complete observedTokens fallback chain (caller > runtimeContext > telemetry > maintenance)

mpz4life:
## Summary

Fix the LCM compaction wedge that pinned long sessions (e.g. session 996 on a 128k third-party model deployment) into an unrecoverable "compacted but still over target" loop. The observed-token-count **fallback chain was incomplete**: `compact()` consulted only the caller-supplied count, and the deferred-compaction drain fell straight through to `maintenance.currentTokenCount` — LCM's own last-written counter — whenever the caller did not supply a fresh value. The maintenance counter as the final fallback tier is intentional and stays; what was missing were the fresher runtime-usage and telemetry tiers in front of it.

## Reproduction

Session 996 failure timeline:

| Time (UTC) | Event |
|---|---|
| 04:09:00 | compaction requested (`reason=threshold`) |
| 04:09:09 | `compact:done ok=false compacted=true reason=compacted_but_still_over_target` |
| 04:09:09 | 30-minute summary-spend backoff started |
| 04:11:57 – 04:17:42 | 15 consecutive `assemble: degraded live fallback reason=near-budget` (stored 77,794 → 77,997; projected 92,171) |
| 04:17:42 | `Compaction skipped: context is already under the compaction target • Context 88k/128k (69%)` |
| 04:17:45 | `Auto-compaction could not recover this turn` — session terminated |

Root cause: `compaction.evaluate()` received `observedTokens=92,171` (stale maintenance counter) instead of `88,235` (the actual API-reported prompt token count available via `runtimeContext.usage` and persisted in `conversation_compaction_telemetry.last_observed_prompt_token_count`). Because the stored-vs-observed gap stayed large, every sweep "succeeded" in storage but left the live transcript over threshold; subsequent sweeps repeated the same work and the session wedged.

## Fix

Add a single private helper `resolveLiveTokenCount(params, options)` implementing the full fallback chain, and route all three compaction entry points through it:

| Priority | Source | Notes |
|---|---|---|
| 1 | caller-supplied `currentTokenCount` (incl. legacy) | explicit caller override; always wins |
| 2 | `runtimeContext.usage` / `lastCallUsage` / `promptCache.lastCallUsage` (sum of `input + cacheRead + cacheWrite`, or raw `prompt_tokens`) | model-returned prompt tokens from the last API call |
| 3 | persisted telemetry `last_observed_prompt_token_count` | fresh snapshot; only consulted by the deferred-drain entry point |
| 4 | `maintenance.currentTokenCount` | **retained** as the final fallback for legacy debt rows; emits a `warn`-level log when taken |

Before/after per entry point:

- `executeCompactionCore` (`compact()`): caller-only → caller → runtimeContext usage → stored-only.
- `consumeDeferredCompactionDebt` (the `maintain()` drain): straight to maintenance counter → caller → runtimeContext usage → telemetry → maintenance counter (warn-logged).
- `evaluatePostTurnCompaction` (`afterTurn`): two-tier runtimeContext/caller → the same shared four-tier chain, so all three entry points now resolve identically.

This **completes** the chain rather than removing the maintenance fallback: no data path changes shape, existing rows remain readable, and the stale-counter branch is now visible in logs.

## What is NOT in this PR

Per the scope constraint, the following are explicitly out-of-scope and tracked as future changes:

- Widening the transcript-wedge trigger (`lossless-claw-30b.4`) to fire without an explicit `observedTokens` argument.
- Tuning the summary-spend backoff duration so drained debt does not re-arm on the next turn.
- ~~Fixing the overhead-truncation behavior in `observedRuntimeOverhead`~~ — addressed separately in Martian-Engineering/lossless-claw#1154.
- Automating the `/new` reset recovery after a transcript wedge so users do not have to invoke it manually.

## Backward compatibility

- Public API surface (`ContextEngine.compact()`, `ContextEngine.maintain()`, `consumeDeferredCompactionDebt` params) is unchanged.
- Database schema is unchanged; existing rows continue to be readable.
- Existing call paths that already had a correct runtime value flowing through are unchanged; caller-supplied `currentTokenCount` keeps winning wherever it previously won.

## Tests

11 vitest cases covering every chain tier, all passing. TDD flow: 6 cases failed against `main` HEAD before the fix (the bug), 3 passed as regression guards, and after the fix all 11 pass — the touched test files (`engine-compaction.test.ts`, `engine-maintenance.test.ts`, `engine-after-turn.test.ts`) show 103 passing.

Cases added:

- `executeCompactionCore prefers runtimeContext.usage over maintenance.currentTokenCount`
- `executeCompactionCore prefers runtimeContext.promptCache.lastCallUsage when top-level usage is absent`
- `executeCompactionCore caller-supplied currentTokenCount wins over runtimeContext.usage (caller override)`
- `executeCompactionCore falls back to params.currentTokenCount when runtimeContext has no usage`
- `executeCompactionCore with no observation proceeds with stored-only counts`
- `consumeDeferredCompactionDebt prefers runtimeContext.usage over telemetry and stale maintenance.currentTokenCount`
- `consumeDeferredCompactionDebt falls back to telemetry when runtimeContext has no usage record`
- `consumeDeferredCompactionDebt uses maintenance.currentTokenCount as last resort and emits a warn log`
- `consumeDeferredCompactionDebt uses runtimeContext.currentTokenCount when supplied and no usage record is present`
- `consumeDeferredCompactionDebt emits the maintain debug log when runtimeContext.usage is selected`
- `consumeDeferredCompactionDebt proceeds with undefined resolvedCurrentTokenCount when no observation is available`

## Changeset

Added `.changeset/fix-compaction-live-tokens-from-model.md` (`patch` bump per AGENTS.md smallest-appropriate-bump table — this is a bug fix with no public API change, no schema change).

### DIFF
--- a/.changeset/fix-compaction-live-tokens-from-model.md
+++ b/.changeset/fix-compaction-live-tokens-from-model.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix LCM compaction `observedTokens` to prefer the host-observed, model-returned prompt token count from `runtimeContext.usage` / `lastCallUsage` / `promptCache.lastCallUsage` over the stale `maintenance.currentTokenCount` counter. The deferred-compaction drain now consults the persisted compaction-telemetry snapshot as an additional fallback. Resolves the "compacted but still over target" wedge that pinned long sessions (e.g. session 996 on the `zte / co-claw` 128k model) into an unrecoverable loop.
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -958,6 +958,86 @@ export class LcmContextEngine implements ContextEngine {
     return Math.floor(value);
   }
 
+  /**
+   * Resolve the host-observed, model-returned live token count from a
+   * compact()/maintain() params bag using a documented priority chain.
+   *
+   * Order (highest priority first; first non-`undefined` wins):
+   *   1. `runtimeContext.usage` / `lastCallUsage` / `promptCache.lastCallUsage`
+   *      (the actual prompt token count returned by the model on the last API
+   *      call — the authoritative baseline).
+   *   2. Caller-supplied `currentTokenCount` and legacy `currentTokenCount`.
+   *   3. Persisted compaction telemetry snapshot (`lastObservedPromptTokenCount`)
+   *      — only consulted when `includeTelemetry` is true, which is the
+   *      deferred-drain entry point that may not have a live runtimeContext.
+   *   4. `staleFallback` — a last-resort value LCM wrote into its own debt
+   *      row. When used, a warn log is emitted so the operator can see the
+   *      session needs a fresher baseline.
+   *
+   * Returned `source` is one of `"runtimeContext" | "caller" | "telemetry" |
+   * "staleFallback" | "none"`. Callers use it for both the priority-chain
+   * debug log and the staleness warning.
+   */
+  private async resolveLiveTokenCount(
+    params: {
+      runtimeContext?: Record<string, unknown>;
+      currentTokenCount?: number;
+      legacyParams?: Record<string, unknown>;
+    },
+    options: {
+      conversationId?: number;
+      includeTelemetry?: boolean;
+      staleFallback?: number;
+      subject: string;
+      sessionLabel: string;
+    },
+  ): Promise<{ value: number | undefined; source: "runtimeContext" | "caller" | "telemetry" | "staleFallback" | "none" }> {
+    const runtimePromptTokens = extractRuntimePromptTokenCount(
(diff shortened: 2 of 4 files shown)


mpz4life:
Hi @jalehman  👋 — would you mind taking a look at this one when you get a chance?
It fixes a critical wedge where compaction was reading from a stale maintenance counter instead of the model-returned runtime usage (Session 996's "compacted but still over target" loop). The fix unifies token resolution via resolveLiveTokenCount, with +11 new tests and no API/schema changes.
Would especially appreciate eyes on the priority chain and the new telemetry fallback tier. Thanks in advance! 🙏

PR #1154, opened 2026-09-07
fix: preserve sign of observedRuntimeOverhead so negative LCM-vs-host gaps surface

mpz4life:
## Summary

`executeCompactionCore` derived `observedRuntimeOverhead` as `Math.max(0, compactableObservedTokens - decisionStoredTokens)`, silently clamping the case where LCM's stored-side counter exceeds the host-observed prompt token count. The two counters come from different estimators (LCM uses a rough JSON-length heuristic, the host reports a tiktoken-style precise count), and after cache hits the host-visible prompt is frequently smaller than LCM's self-estimate. When the clamp truncated the discrepancy to zero, the runtime-adjusted sweep target stayed unset and sessions wedged on "compacted but still over target".

## Behavior change

**Before** (`src/engine.ts:2048-2051`):

```typescript
const observedRuntimeOverhead =
  params.compactionTarget === "threshold" && compactableObservedTokens !== undefined
    ? Math.max(0, compactableObservedTokens - decisionStoredTokens)
    : 0;
```

**After**:

```typescript
const observedRuntimeOverhead =
  params.compactionTarget === "threshold" && compactableObservedTokens !== undefined
    ? compactableObservedTokens - decisionStoredTokens
    : 0;
```

Plus a debug-level log line emitted exactly once per `executeCompactionCore` call when the overhead is negative, surfacing the LCM-vs-host methodology gap that was previously hidden:

```typescript
if (observedRuntimeOverhead < 0) {
  this.deps.log.debug(
    `[lcm] compact: observed runtime overhead negative conversation=${conversationId} ${sessionLabel} storedTokens=${decisionStoredTokens} observedTokens=${compactableObservedTokens} observedRuntimeOverhead=${observedRuntimeOverhead} — LCM stored counter exceeds host-observed prompt; methodology gap, no runtime sweep target adjustment`,
  );
}
```

**Why this is safe**: the downstream guard at `src/engine.ts:2052-2057` (`runtimeAdjustedSweepTargetTokens = observedRuntimeOverhead > 0 && compactableObservedTokens > targetTokens ? … : undefined`) already restricts the side effect to positive values. The observable behavior is:

| `observedRuntimeOverhead` | Before | After |
|---|---|---|
| `> 0` (host sees more than LCM stored) | sweep target adjusted | sweep target adjusted — **identical** |
| `== 0` (host and LCM agree) | no adjustment | no adjustment — **identical** |
| `< 0` (host sees less than LCM stored) | no adjustment, discrepancy hidden as `0` | no adjustment, discrepancy visible in debug log — **new observability** |

## What is NOT in this PR

Out of scope per the failure-analysis report; each will get its own OpenSpec change proposal:

1. Widening the transcript-wedge trigger so it fires without an explicit `observedTokens` argument.
2. Tuning the summary-spend backoff duration so drained debt does not re-arm on the next turn.
3. Automating the `/new` reset recovery after a transcript wedge so users do not have to invoke it manually.

No public API surface changes. No persisted schema changes. No backoff, transcript-wedge trigger, or `/new` automation behavior touched.

## Backward compatibility

- **Public API surface**: `ContextEngine.compact()` and the internal `executeCompactionCore` parameter shapes are unchanged.
- **Persisted schema**: No migration, no column add/remove/rename. Existing rows continue to be readable without a one-time backfill.
- **Persisted compaction-telemetry snapshots and maintenance counters**: Continue to be read and written exactly as before.
- **Public log format**: New line uses the existing `[lcm] compact:` prefix; no breaking change to log consumers.
- **Pre-existing call paths**: All previously working call shapes continue to work; only the sign of the discrepancy is now exposed instead of being silently clamped.

## Tests

TDD red → green flow:

- **Red phase** (current `main`): 1/3 tests pass (positive overhead — regression guard). 1/3 tests fail (negative overhead — debug log not emitted because `Math.max(0, …)` truncated the discrepancy). 1/3 tests pass (zero overhead — regression guard).
- **Green phase** (after fix): 3/3 tests pass.
- **Full vitest on touched files** (`test/engine-compaction.test.ts`, `test/engine-maintenance.test.ts`, `test/engine-after-turn.test.ts`): **95/95 pass**.
- **Full vitest**: **1788/1789 pass**, with 5 test files failing purely on the pre-existing `Missing "./plugin-sdk/session-transcript-runtime" specifier in "openclaw" package` environmental error (unrelated to this change, reproduces on a clean checkout).

Three new vitest cases added to `test/engine-compaction.test.ts` under a new `describe("executeCompactionCore observedRuntimeOverhead", …)` block:

1. `preserves positive overhead and arms sweep target when observed exceeds stored` — `compactFullSweep` receives `stopAtTokens: 3_200`, `result.result.details.observedOverheadTokens === 5_000`.
2. `preserves negative overhead, omits stopAtTokens, and emits debug log` — `compactFullSweep` does NOT receive `stopAtTokens`, a debug log line prefixed `[lcm] compact: observed runtime overhead negative` is emitted once with `storedTokens=92171`, `observedTokens=88235`, `observedRuntimeOverhead=-3936`.
3. `keeps zero overhead behavior unchanged: no stopAtTokens, no debug log` — neither side effect fires (gate is `< 0`, not `<= 0`).

## Changeset

`.changeset/fix-compaction-runtime-overhead-sign.md` — `patch` bump (bug fix, no public API change, no schema change).

### DIFF
--- a/.changeset/fix-compaction-runtime-overhead-sign.md
+++ b/.changeset/fix-compaction-runtime-overhead-sign.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Fix LCM compaction `observedRuntimeOverhead` to preserve the sign of the LCM-vs-host methodology gap instead of silently truncating negative values to zero. Previously `Math.max(0, compactableObservedTokens - decisionStoredTokens)` hid the case where LCM's stored-side counter exceeds the host-observed prompt token count (e.g. cache hits shrinking the visible prompt below LCM's JSON-length estimate); the runtime-adjusted sweep target stayed unset and sessions wedged on "compacted but still over target". The fix keeps the downstream `> 0` guard intact so positive overhead still arms the sweep target, and adds a debug log line that surfaces the discrepancy when overhead is negative.
--- a/openspec/changes/fix-compaction-runtime-overhead-sign/proposal.md
+++ b/openspec/changes/fix-compaction-runtime-overhead-sign/proposal.md
@@ -1,0 +1,186 @@
+# Proposal: fix-compaction-runtime-overhead-sign
+
+## Why
+
+`executeCompactionCore` (src/engine.ts:2048-2051) derives the
+runtime-side overhead as
+
+```typescript
+const observedRuntimeOverhead =
+  params.compactionTarget === "threshold" && compactableObservedTokens !== undefined
+    ? Math.max(0, compactableObservedTokens - decisionStoredTokens)
+    : 0;
+```
+
+The `Math.max(0, …)` is an implicit assumption that the host's
+observed prompt is always at least as large as LCM's internal stored
+counter. In reality the two are computed by different methods and
+`storedTokens > observedTokens` is observable:
+
+- LCM uses `JSON.stringify().length/4` (rough heuristic, source
+  tokens perspective).
+- The host reports a tiktoken-style precise count, frequently after
+  cache hits and after prompt caching has collapsed repeated system
+  blocks. Host-visible prompt is smaller than LCM's self-estimate.
+- Session 996 evidence (a 128k third-party model deployment):
+  `decisionStoredTokens = 92_171`, true
+  `observedTokens = 88_235`, so
+  `Math.max(0, 88_235 - 92_171) = 0` — the discrepancy is silently
+  truncated and the runtime-adjustment branch
+  (`runtimeAdjustedSweepTargetTokens`) is never armed, so the sweep
+  target stays at the stored-side threshold and the session keeps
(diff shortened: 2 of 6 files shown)


jalehman:
Thanks for digging into this and documenting the negative LCM-versus-host token gap. We are closing this as superseded by #1157.

#1157 fixes the operational problem by preventing a lower observed runtime count from discounting stored context, and by clearing stalled threshold debt when the projected prompt is safely within budget. This PR would preserve the signed gap in diagnostics, but its downstream `> 0` guard leaves behavior unchanged for the motivating negative case.

If we need the signed estimator delta for production diagnostics, we can add it separately under a name that does not treat a negative delta as runtime overhead.

Issue #1146, opened 2026-09-01
Severe prompt-processing slowdown and stale transcript ranges with Lossless Claw 1.0.0


PBL-3421:
## Summary

Lossless Claw 1.0.0 caused a substantial slowdown in new OpenClaw conversations. A simple greeting took approximately 44 seconds with Lossless Claw active.

After disabling Lossless Claw and explicitly changing `plugins.slots.contextEngine` to `legacy`, the same type of request completed in approximately 3 seconds.

## Environment

- macOS on Apple Silicon Mac Studio
- OpenClaw: 2026.8.1
- Lossless Claw: 1.0.0
- Model provider: LM Studio
- Model: Qwen3.6 35B A3B, 4-bit
- Model context length in LM Studio: 262,144 tokens
- OpenClaw context-engine token budget reported in one warning: 128,000 tokens
- OpenClaw interface: Control UI/webchat
- Lossless Claw summary and expansion model: `venice/kimi-k2-5`

## Observed behavior

In a newly created conversation, I sent:

> Hi KITT

Response time was approximately 44 seconds.

A direct request to the same loaded LM Studio model, without OpenClaw’s context assembly, completed in approximately 0.46 seconds. This indicated that LM Studio and the model itself were responsive.

The OpenClaw logs repeatedly reported warnings similar to:

```text
[context-engine] durable turn advancement is blocked: [ID]: transcript range is stale


***
There were approximately 13 stale transcript-range warnings during the affected request.
Another warning appeared after the request:
***

[lcm] commitTurn: tokenBudget not provided; using default 128000

***

The LM Studio request was accepted almost immediately:

***

[model-fetch] response provider=lmstudio
api=openai-completions
model=qwen/qwen3.6-35b-a3b
status=200
elapsedMs=10

***

However, the complete response took roughly 44 seconds.
Troubleshooting and results
1. I disabled the lossless-claw plugin but initially left:

***

"plugins": {
  "slots": {
    "contextEngine": "lossless-claw"
  }
}

***

With the plugin disabled but the slot still pointing to it, OpenClaw logged:

***

Context engine "lossless-claw" degraded to "legacy" for this logical turn:
context engine "lossless-claw" is not registered.

***

Response time improved to approximately 15 seconds, but OpenClaw attempted the unavailable context engine on every turn.
2. I then explicitly changed the slot to:

***

"plugins": {
  "slots": {
    "contextEngine": "legacy"
  }
}

***

[...]

PR #1155, opened 2026-09-07
fix: avoid scanning unrelated message history when resolving summary source ranges

FtlC-ian:
## Summary

`getSummaryMessageSeqRange` walks the parent-lineage CTE (`source_summaries`), then joins `summary_messages` and `messages` to compute `MIN/MAX(seq)`. Because plain `JOIN` lets SQLite reorder freely, on analyzed databases the optimizer can pick a global `SCAN sm` over the whole `summary_messages` table — repeated per lineage summary — stalling context assembly on large stores.

This change converts those two joins to `CROSS JOIN`, which SQLite documents as preserving join order, so the optimizer visits the lineage CTE first and then uses index lookups (`SEARCH sm USING COVERING INDEX sqlite_autoindex_summary_messages_1 (summary_id=?)`). No schema or index changes; asymptotic cost now scales with lineage size rather than total table size.

## Reproduced query plans (node:sqlite, in-memory, 4096 seeded messages across one conversation, one message link per summary, `ANALYZE`)

**Before (plain JOIN):**
```
... | SCAN sm | BLOOM FILTER ON source (summary_id=?) | SEARCH source USING AUTOMATIC COVERING INDEX (summary_id=?) | SEARCH m USING INDEX sqlite_autoindex_messages_1 (message_id=?)
```

**After (CROSS JOIN):**
```
... | SCAN source | SEARCH sm USING COVERING INDEX sqlite_autoindex_summary_messages_1 (summary_id=?) | SEARCH m USING INDEX sqlite_autoindex_messages_1 (message_id=?)
```

## Observational evidence from a live deployment

Context-assembly timings recorded around a deployment of this fix (different workloads before vs. after, **not** a controlled benchmark or claimed universal effect):

- last 30 pre-reload context assemblies: median **7311.5 ms**, max **52893 ms**
- 21 post-reload context assemblies: median **40 ms**, max **484 ms**

## Test evidence

- Focused suite: `npx vitest run test/summary-store.test.ts` → 29 passed (includes 6 in `pending-summary-store.test.ts` picked up by same directory run)
- New regression coverage:
  - Range resolution through shared ancestors, correctly excluding unrelated conversations (`shared`/`leaf`/`left`/`right`/`root`/`empty`/`missing` cases).
  - Query-plan assertion that `sm` is `SEARCH`ed by `summary_id` rather than `SCAN`ed, verified via `EXPLAIN QUERY PLAN` on a seeded, analyzed 4096-message database.

## Related

- #1146 reports a severe slowdown on large stores with Lossless Claw 1.0.0; this fix addresses one plausible contributing query path (per-lineage source-range resolution), not the full scope of that report.

## Worked on by

- Ian F (authored/reviewed via Zed agent, delegated work)


jalehman:
Thanks for the contribution, @FtlC-ian. This is landed now.

PR #1156, opened 2026-09-07
fix: re-seat a user turn in the degraded live-assembly fallback

JerretK:
Under sustained context pressure with an in-flight compaction, buildDegradedLiveAssembleResult budget-trims the live tail and can drop every user message on a long tool loop, returning a context with only system/assistant/tool turns. Strict chat templates (e.g. Qwen3) reject that with HTTP 400 "No user query found in messages", which fails the whole agent turn. Unlike the main assemble path, this degraded fallback skipped the no-user-turns guard.

Re-seat the most recent evicted user turn after the protected prefix, mirroring clampMessagesToSerializedBudget. Adds a regression test that reproduces the shape (one early user turn, long tool loop, tight budget) and fails without the fix.


jalehman:
Thanks for isolating the missing-user-turn failure and adding the focused regression, @JerretK. PR #1149 now covers the same degraded-fallback case through `buildProviderValidSuffix()`: it restores the most recent preceding user turn whenever trimming removes all user messages, and it also restores matching tool-call/result partners for both degraded assembly and serialized-budget clamping. I’m closing this PR as superseded so we can review and maintain one implementation path. Your report and test case helped establish the required behavior.

PR #1157, opened 2026-09-08
fix: clear threshold sweep debt when projected prompt stays within budget

mpz4life:
## Summary

A threshold compaction sweep can reduce stored context and then stall above the ideal target because protected context and runtime framing remain. Treat that sweep as settled when an observed prompt count is available and its post-sweep projection fits the token budget. This clears deferred debt without opening a summary spend backoff for work that cannot make further progress.

Sweeps stopped by a work limit or still reducing context retain their debt. Both threshold convergence and safe-watermark clearance use a projection no smaller than stored context, so a negative observed-versus-stored estimator gap cannot hide real budget pressure. This also preserves the budget checks when combined with the signed-overhead diagnostics proposed in #1154.

Provider auth failures, host-owned prompt accounting, sweeps without observed counts, and over-budget results retain their failure or strict-threshold behavior.

## Changes

- Add the safe-watermark threshold verdict.
- Preserve debt after bounded partial progress.
- Keep projections above stored token counts.
- Add maintenance and token-accounting regressions.

## Testing

- `npm run typecheck`
- `npm run build`
- `npx vitest run test/engine-compaction.test.ts test/engine-maintenance.test.ts test/lcm-integration-sweep-bounds.test.ts test/lcm-integration-compaction.test.ts test/engine-after-turn.test.ts test/compaction-maintenance-store.test.ts` — 176 tests pass.
- Temporarily apply #1154's signed-overhead arithmetic and run the compaction and maintenance suites — 97 tests pass. The sign change itself is not included here.
- Regression cases cover sweep/chain/deadline/spend limits, positive and negative overhead, provider auth failure, and host-owned counts.

## Changeset

Patch: user-facing compaction behavior fix, recorded in `.changeset/threshold-sweep-safe-watermark.md`.

### DIFF
--- a/.changeset/threshold-sweep-safe-watermark.md
+++ b/.changeset/threshold-sweep-safe-watermark.md
@@ -1,0 +1,5 @@
+---
+"@martian-engineering/lossless-claw": patch
+---
+
+Treat a threshold sweep that made progress but still misses the ideal target as settled when the projected post-sweep prompt stays within the token budget. Stored compaction cannot shrink fixed runtime framing (anchors, tool schemas, fresh tail), so an in-budget projection no longer pins deferred-compaction debt behind a summary spend backoff — avoiding near-budget degradation getting stuck on the host's raw-transcript precheck. Without an observed prompt token count, the previous strict judgment is preserved.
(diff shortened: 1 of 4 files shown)


mpz4life:
@jalehman This PR is ready for review.

jalehman:
Thanks for the contribution, @mpz4life. This is merged now. Maintainer review added guards to preserve debt after bounded partial sweeps and to keep negative estimator gaps from discounting stored context, including the interaction with #1154. The patch changeset is included.

PR #1159, opened 2026-09-09
fix(assemble): preserve retained user replay identity and runtime context

Marvinthebored:
## Summary

Fix prompt-prefix changes caused by the loss of retained user replay identity and adjacent host runtime-context messages during assembly.

LCM reconstructs raw user messages without their timestamps or `idempotencyKey` values. OpenClaw can then fail to recognize an admitted user and submit another rendered copy. The separate `openclaw.runtime-context` message also disappears from reconstructed history. These changes invalidate cached history even when the summary prefix remains unchanged.

## Change

- Carry verified transcript IDs through raw-message assembly with existing symbol metadata.
- Restore timestamps and replay keys using canonical entry IDs, unique live keys, and content checks.
- Preserve adjacent host runtime-context messages only for retained raw users.
- Include these messages in the serialized budget check. Remove a restored runtime-context message if its parent is evicted, and update the token estimate.
- Retain assembled context if the canonical transcript read fails.

This does not restore summarized turns, replace externalized content, or infer identity from text alone. It adds no schema, configuration, dependency, or host API requirement. Canonical lookup uses one scoped transcript read when keyed live users are present, without a persistent replay cache.

## Evidence

Six pre-fix request pairs showed the rendered user and adjacent runtime-context message disappear at the next user turn. Tools, system blocks, summaries, and earlier messages stayed unchanged. The deletion also occurred before OpenClaw 2026.9.3.

The patch is deployed locally on lossless-claw 1.0.0 with OpenClaw 2026.9.3. A post-fix pair exercised database reconstruction with seven unchanged summaries. All 116 earlier messages and the adjacent runtime-context message survived replay. Original and provider-transformed requests agreed. Private transcripts and request bodies are not published.

## Validation

Two focused tests cover reconstructed replay across a summarized turn and repeated raw users, canonical identity guards, budget eviction, token accounting, read failure, and preservation of externalized content.

- The replay regression failed before the fix because the user's replay key was absent, then passed with the fix.
- The budget regression caught a stale token estimate after carrier removal, then passed after correction.
- Full suite: 1,852 tests passed on the PR base.
- Typecheck and plugin build passed.
- An independent adversarial review found the token-accounting defect. The correction received a second review with no further findings.

```sh
npx vitest run test/engine-user-replay.test.ts
npm test -- --maxWorkers=2
npm run typecheck
npm run build:plugin
git diff --check
```

## Related work

- #688 preserves volatile live inputs, rather than these replay fields and adjacent runtime-context messages.
- #787 filters leaked assistant runtime context during ingestion. This change does not store runtime-context messages in the LCM database.

## Release and disclosure

AI-assisted investigation, implementation, and review. Publication is approved by the submitting user.

This is a user-visible patch. Per the repository policy, a maintainer needs to add a patch changeset before release.


jalehman:
Thanks for the contribution, @Marvinthebored. This is landed now.

PR #1162, opened 2026-09-10
fix: preserve structured tool anchor identity

Marvinthebored:
## Summary

- Preserve structured tool identity when transcript anchor parent text is blank.
- Fail closed on ambiguous or corrupt anchors instead of adopting weak matches.
- Keep all message content and make externalized replay behavior idempotent.

## Why

Blank parent text can produce weak identity collisions even when structured tool parts differ. This change canonicalizes structured-part identity and validates provenance and payloads before anchor adoption or replay deduplication.

## Validation

- Full test suite: 1,862 tests passed across 115 files.
- Typecheck passed.
- Synthetic regression baseline had 3 failures; the current suite passes.

## Release note

Includes a patch changeset for `@martian-engineering/lossless-claw`.


jalehman:
Thanks for the contribution, @Marvinthebored. This is landed now.

PR #1164, opened 2026-09-11
fix: `enableSummaryThinking: false` could not disable reasoning

### LINKED ISSUE #1160: `enableSummaryThinking: false` does not disable reasoning — summaries silently truncated on Qwen chat-template models
**Version:** lossless-claw 1.0.0 · OpenClaw 2026.9.2 · provider api `openai-completions`, model `compat.thinkingFormat: "qwen-chat-template"`, `reasoning: true`

## Documented behaviour
`docs/configuration.md`:
> `enableSummaryThinking` … When true, requests low reasoning budget from the model during summarization calls. **Set to false to disable reasoning** and keep summarization output concise.

## Actual behaviour
Setting it to `false` does not disable reasoning — it makes reasoning *larger*.

The summarizer call only ever produces two shapes:
```js
// initial call
...shouldRequestSummaryThinking ? { reasoningIfSupported: "low" } : {}, ...reasoning ? { reasoning } : {}
// retry
const retryReasoning = shouldRequestSummaryThinking ? "low" : void 0;
```
and the plugin's own `complete` wrapper forwards **only** `reasoning`, never `reasoningIfSupported`:
```js
runtimeLlm({ messages, ...model, ...maxTokens, ...temperature, ...systemPrompt,
             purpose: "lossless-claw compaction summarization",
             ...reasoning !== void 0 ? { reasoning } : {} });
```
So the host receives `undefined`, and:
```js
resolveOpenAICompletionsReasoningEffort(o) { return o?.reasoningEffort ?? o?.reasoning ?? "high" }
isOpenAICompletionsThinkingEnabled(e) { const n = e.trim().toLowerCase(); return n !== "off" && n !== "none" }
```
`"high"` → enabled → for `qwen-chat-template` the host writes `chat_template_kwargs.enable_thinking = true`.

| `enableSummaryThinking` | sent to host | host effort | thinking |
|---|---|---|---|
| `true` (default) | nothing (`reasoningIfSupported` dropped) | `"high"` | **on** |
| `false` | nothing | `"high"` | **on** |

There is no code path that can emit `"off"` / `"none"`.

## Measured impact
Same payload shape as the summarizer, against a local vLLM Qwen (`max_tokens: 3000`):

| condition | finish_reason | reasoning tokens | latency | visible chars |
|---|---|---|---|---|
| as shipped (thinking on) | **length** (truncated) | **1840 / 3000 (61%)** | 17.8 s | 6,323 |
| `enable_thinking: false` | stop | 0 | 11.9 s | **11,382** |
| `max_tokens: 8000`, thinking on | stop | 3398 (64%) | 35.8 s | 10,185 |

Reasoning tokens expand to fill whatever budget is given. With `leafTargetTokens: 2400` the visible summary is routinely cut off.

## Secondary issue: the truncation is undetectable
`extractIncompleteResponseSignals` looks for `finishReason` on the result, but the plugin's `complete` wrapper returns `{content, provider, model, agentId, usage, audit}` — no `finishReason`/`stopReason`. So truncated summaries never trigger the retry path and produce no log signal.

Observed in our `lcm.db`: one leaf summary with `token_count = 1`, content `":"`, generated from `source_message_token_count = 20037`. Non-empty and no finishReason ⇒ silently accepted.

Aggregate compression ratio on the same corpus: Qwen leaf **0.0263** vs Claude Sonnet leaf **0.0677** (2.6× more lossy).

## Suggested fix
Emit an explicit disabling effort when the operator opts out, so the documented behaviour matches:
```diff
- ...shouldRequestSummaryThinking?{reasoningIfSupported:"low"}:{},...reasoning?{reasoning}:{}
+ ...shouldRequestSummaryThinking?{reasoningIfSupported:"low"}:{reasoning:"off"},...reasoning?{reasoning}:{}

- const retryReasoning=shouldRequestSummaryThinking?"low":void 0;
+ const retryReasoning=shouldRequestSummaryThinking?"low":"off";
```
`enableSummaryThinking: true` (the default) is unchanged, so this is a no-op unless the operator opts out. We are running this locally and it resolves the truncation.

Separately, forwarding `finishReason`/`stopReason` through the `complete` wrapper would let `extractIncompleteResponseSignals` and the retry path work as intended.

igs-rogenlo:
Fixes #1160.

## What is broken

`docs/configuration.md` says of `enableSummaryThinking`: *"Set to false to disable reasoning and keep summarization output concise."* No code path can emit a value the host reads as off, and the `true` path cannot request `"low"` either.

`src/summarize.ts` asks for `reasoningIfSupported: "low"`, and withholds the field when summary thinking is disabled. `src/plugin/index.ts` destructures `reasoningIfSupported` but forwards only `reasoning` to `runtime.llm`, using the former solely for `request_reasoning` log metadata. The host therefore receives nothing in both cases, and an absent effort resolves to the strongest one:

```js
resolveOpenAICompletionsReasoningEffort(o) { return o?.reasoningEffort ?? o?.reasoning ?? "high" }
isOpenAICompletionsThinkingEnabled(e) { const n = e.trim().toLowerCase(); return n !== "off" && n !== "none" }
```

For `compat.thinkingFormat: "qwen-chat-template"` that becomes `chat_template_kwargs.enable_thinking = true`.

| `enableSummaryThinking` | sent to host | host effort | thinking |
|---|---|---|---|
| `true` (default) | nothing — `reasoningIfSupported` dropped | `"high"` | on |
| `false` | nothing | `"high"` | **on** |

## What this changes

- **Forward `reasoningIfSupported` as `reasoning`** when no explicit `reasoning` is set. `runtime.llm` has no separate "if supported" field, and the effort it falls back to is the strongest one, so passing the weaker request through is what makes the field mean anything.
- **Send `reasoningIfSupported: "off"`** when an operator set `enableSummaryThinking: false`. Silence cannot express "off" to a host that defaults to `"high"`, so the disabled state has to be said out loud. It reaches the retry too, which previously also sent nothing.

The ollama exclusion keeps sending nothing. That one is the plugin's own decision rather than the operator's, so it should not start asserting an effort where it never did.

| `enableSummaryThinking` | after this change | host effort | thinking |
|---|---|---|---|
| `true` (default) | `reasoning: "low"` | `"low"` | on, bounded |
| `false` | `reasoning: "off"` | `"off"` | **off** |
| any, provider `ollama` | nothing | unchanged | unchanged |

## Why it is worth fixing

Measured against a local vLLM Qwen with the summarizer's own payload shape, `max_tokens: 3000`:

| condition | finish_reason | reasoning tokens | latency | visible chars |
|---|---|---|---|---|
| as shipped (thinking on) | **length** (truncated) | **1840 / 3000 (61%)** | 17.8 s | 6,323 |
| `enable_thinking: false` | stop | 0 | 11.9 s | **11,382** |
| `max_tokens: 8000`, thinking on | stop | 3398 (64%) | 35.8 s | 10,185 |

The summariser's budget is spent on hidden reasoning and the summary is truncated mid-write, which is the outcome the setting exists to avoid.

## Tests

Three existing cases pinned the old behaviour at the two boundaries this touches and now pin the documented one:

- `test/index-runtime-llm-complete.test.ts` asserts the exact `runtime.llm.complete` payload — it now carries `reasoning: "low"`.
- `test/lcm-summarizer-reasoning.test.ts` asserts the initial call and the retry both carry `reasoningIfSupported: "off"` when the setting is false.
- `test/summarize.test.ts` asserts the same on the headroom case.

Mutation-checked: reverting the `src/plugin/index.ts` forwarding fails the first; reverting the `src/summarize.ts` branch fails the other two. Ollama cases are untouched and still assert that neither field is sent.

Full suite: 117 files, 1874 tests passing. `tsc --noEmit` clean.


jalehman (changes requested):
I cannot merge this yet. The plugin-side forwarding is useful, but the disabled path does not fix the reported Qwen behavior through the OpenClaw runtime.

1. **The host drops `off` before the Qwen transport.** In OpenClaw 2026.9.2, `runtime.llm.complete` forwards `params.reasoning` into `completeWithPreparedSimpleCompletionModel`, whose `normalizeSimpleCompletionReasoning` converts `off` to `undefined` except for Claude Sonnet 5/Opus 5. The managed OpenAI-completions transport then defaults an absent effort to `high`, which sets `chat_template_kwargs.enable_thinking = true`. This affects both the initial request and retry. The same normalization exists in the declared 2026.7.2-beta.2 host and stable 2026.9.4. The tests here mock the runtime boundary and therefore do not cover it.

   Sources: [runtime forwarding at v2026.9.2](https://github.com/openclaw/openclaw/blob/v2026.9.2/src/plugins/runtime/runtime-llm.runtime.ts#L606-L615), [normalization](https://github.com/openclaw/openclaw/blob/v2026.9.2/src/agents/simple-completion-runtime.ts#L689-L699), [transport default](https://github.com/openclaw/openclaw/blob/v2026.9.2/packages/ai/src/transports/openai-completions-params.ts#L67-L69), [Qwen payload](https://github.com/openclaw/openclaw/blob/v2026.9.2/packages/ai/src/transports/openai-completions-params.ts#L224-L229), [v2026.9.4 normalization](https://github.com/openclaw/openclaw/blob/v2026.9.4/src/agents/simple-completion-execution.ts#L104-L114).

   A source-level executable check using those upstream functions, with only the Claude identity checks stubbed false for Qwen, gives:

   ```text
   requested=low -> normalized=low -> effort=low -> enable_thinking=true
   requested=off -> normalized=undefined -> effort=high -> enable_thinking=true
   ```

   This needs a supported host contract and a regression test through that normalization boundary before we can claim the disabled setting works. A plugin-only workaround or host-version change needs separate evaluation.

2. **The disabled branch bypasses the Ollama exclusion.** `src/summarize.ts:1643` sends `reasoningIfSupported: "off"` for Ollama when `enableSummaryThinking` is false. This contradicts the PR's stated provider exception and the omission preserved by #944. Preserve omission for both settings and cover the disabled initial/retry cases.

Validation at `fa39bfe89457cf3c73ebf4a7736ca62a4b122d07`: typecheck passed; 117 focused tests across the runtime bridge, summarizer, and reasoning guardrails passed. Branch autoreview against freshly fetched `origin/main` reported the Ollama finding, which I verified against the code and #944. Existing CI is successful on this exact PR head, but these runtime findings remain unresolved.

Release metadata: this is intended to be a patch fix for 1.0.1 and needs a maintainer-owned changeset once the implementation is ready. No merge or push performed. Also, #1160 includes missing finish/stop-reason forwarding that this diff does not address; that report should remain tracked even after the reasoning fix is completed.

igs-rogenlo:
Both findings are right. One is fixed here; the other is not mine to fix, and I would rather say so plainly than paper over it.

## 2 — the disabled branch bypassed the ollama exclusion. Fixed in `63c496a`.

Your reading was exact. `shouldRequestSummaryThinking` is false for ollama for a reason that has nothing to do with the operator's setting, so with `enableSummaryThinking: false` the ollama rows fell through to the second condition and sent `reasoningIfSupported: "off"`. #944 removed that field for ollama because sending it at all is what broke there, and `"off"` is as much a sent field as `"low"` — so the PR contradicted its own stated provider exception.

The two nested conditions are now one three-way decision, so the ollama row cannot be reached through the disabled branch:

| provider | `enableSummaryThinking` | sent |
|---|---|---|
| ollama | either | field omitted (#944) |
| not ollama | enabled | `"low"` |
| not ollama | disabled | `"off"` |

The initial call and the retry both build their request in `runSummarizerCall`, so one change covers both; the retry's separate `reasoning` argument was already `undefined` in the disabled case.

Two tests added, both driving the retry path so each one asserts the initial call *and* the retry: ollama with thinking enabled (the #944 control, which passes either way) and ollama with thinking disabled (the regression). The regression fails against `fa39bfe` with `expected 'off' to be undefined`. The disabled case passes `"  Ollama  "` on purpose, so the trim-and-lowercase normalization the exclusion depends on is pinned rather than incidental.

`tsc --noEmit` clean; full suite 1873 tests across 117 files pass.

## 1 — the host drops `off` before the Qwen transport. Confirmed, and this PR does not fix it.

I checked your chain against the source rather than taking it on trust, and it holds end to end on current main:

- `src/agents/simple-completion-execution.ts:104-114` — `case "off":` returns `"off"` only when `resolveClaudeSonnet5ModelIdentity(model) || resolveClaudeOpus5ModelIdentity(model)`, and `undefined` otherwise.
- `packages/ai/src/transports/openai-completions-params.ts:76` — `resolveOpenAICompletionsReasoningEffort` is `options?.reasoningEffort ?? options?.reasoning ?? "high"`.
- `packages/ai/src/transports/openai-transport-shared.ts:328-331` — `isOpenAICompletionsThinkingEnabled` is false only for `off` and `none`, so `"high"` enables it, and `setQwenChatTemplateThinking` then writes `chat_template_kwargs.enable_thinking: true`.

The detail I found most useful for a host fix: **the transport already handles `off` correctly.** The only thing between this plugin and the documented behavior is the layer above discarding `off` for every non-Claude model. That makes it a reachability hole rather than a missing capability, which should make the host change smaller than it looks.

So I am explicitly not claiming this PR fixes the reported Qwen behavior, and I would not want it merged on that basis. Two ways forward — your call, not mine:

1. **Hold this PR** until a host contract exists, and I file the host defect upstream with the above plus your executable check. One thing worth knowing before you pick: openclaw closed the matching planning item, `openclaw/openclaw#86099` ("PRD: Model-aware thinking controls for vLLM and Pi OpenAI harness"), as `NOT_PLANNED` on 2026-05-24 and moved it to Linear as REN-688, with the ClawSweeper-recommended maintainer design decision split out as REN-689. So the design item exists, just not on GitHub. I would file a narrow released-code defect referencing REN-688 rather than another PRD.
2. **Land the plugin-side forwarding only** — the half you called useful — and leave the disabled path out, so the plugin is already correct when a host contract arrives and nothing claims to work before then.

No preference from me on the release metadata; the changeset is yours to own once the implementation is settled.

## #1160

Agreed, and untouched here — I have not gone near the finish/stop-reason forwarding. That report should outlive the reasoning fix.

Issue #1161, opened 2026-09-10
`contextThresholdOverrides` rules matching on `modelContextWindowMin/Max` can never match — host never supplies the field


igs-rogenlo:
**Version:** lossless-claw 1.0.0 · OpenClaw 2026.9.2

## Symptom
Any `contextThresholdOverrides` rule that uses `match.modelContextWindowMin` or `match.modelContextWindowMax` is silently inert. It never matches, never logs an error, and `threshold: selected` reports `source=global reason=no_override_matched`.

Across 3,986 consecutive `threshold: selected` entries on our deployment:

| field | value | count |
|---|---|---|
| `modelContextWindow` | `none` | **3986 / 3986 (100%)** |
| `source` | `global` | 3986 / 3986 |
| `ruleName` | `none` | 3986 / 3986 |

## Cause
`ruleMatches` rejects immediately when the runtime context window is unknown:
```js
if (rule.match.modelContextWindowMin !== void 0 || rule.match.modelContextWindowMax !== void 0) {
  if (runtime.modelContextWindow === void 0) { return false }
```
`runtime.modelContextWindow` comes from `readRuntimeModelContext`, which scans the runtime-context bags for:
```js
MODEL_CONTEXT_WINDOW_KEYS = ["modelContextWindow","modelContextWindowTokens","contextWindow",
                            "contextWindowTokens","maxContextTokens","contextWindowMax"]
```
OpenClaw's `buildAfterTurnRuntimeContext` supplies `tokenBudget` but none of those six keys, so the value is always `undefined`.

This also means the documented example in `docs/configuration.md` (`{"match": {"modelContextWindowMin": 900000}, ...}`) cannot work as written on a stock OpenClaw host.

## Notes
- `match.model` does work, but on hosts that set `agents.defaults.compaction.model` the resolved `modelRef` is the **compaction** target, not the conversation model (95% of our samples resolve to the compaction model), so it is not a substitute for window-based matching.
- The host does compute a context window internally (`resolveCompactionContextTokenBudget` reads `model.contextWindow`), so the data exists — it just isn't placed in the bag passed to the context engine.

## Suggestions
Either
1. document that window-based matching requires host support that current OpenClaw releases do not provide, and/or validate the config at load time so dead rules are reported instead of silently ignored; or
2. derive the window inside the plugin from the `tokenBudget` already present in the runtime context; or
[...]

PR #1165, opened 2026-09-11
fix: name the case where a `modelContextWindow` rule can never match

igs-rogenlo:
Refs #1161. This takes the reporting half of option (1) in that issue, and deliberately not option (2).

## Why not derive the window from `tokenBudget`

`ruleMatches` already refuses that on purpose, and says so:

```ts
// Window-range matchers require explicit runtime window metadata — the token
// budget is never used as a proxy.
```

Checking the host confirms the refusal is right. `resolveCompactionContextTokenBudget` returns `min(requested ?? fallback ?? resolvedBudget, resolvedBudget)`, where `resolvedBudget` is derived from the model's context window. So `tokenBudget` equals the window only when no explicit budget is configured, and is smaller otherwise. Matching `modelContextWindowMin` against a low estimate merely under-matches, but `modelContextWindowMax` against one **over**-matches: a 1M-window model whose operator capped the budget at 200k would satisfy `modelContextWindowMax: 250000` and silently take a rule meant for small models. That is worse than the current inertness, so the proxy stays rejected.

## What this changes

The inertness itself is documented. What is missing is any way to see it happening: a rule that can never match reports `reason=no_override_matched`, the same string reported when the ranges simply did not cover the model. One of those means "widen the range", the other means "window matchers cannot work on this host" — and they are indistinguishable in the logs.

This reports `no_override_matched_window_metadata_absent` instead, whenever a window-range rule is configured *and* the host supplied no window size. The flag is computed once in the constructor, so `resolve` still just scans the compiled rules. Everything else is untouched: matching behaviour, specificity, threshold selection, and the plain reason in every other case.

The configuration guide gains a pointer from its own example to the caveat already below it — `large-context-models` and `small-context-models` there are exactly the rules that go inert — and a line naming the new reason so the string is searchable.

## What it looks like

Before, on a stock OpenClaw host, across 3,986 consecutive `threshold: selected` entries: `modelContextWindow=none` in 3,986, `ruleName=none` in 3,986, `reason=no_override_matched` in 3,986 — with nothing saying the rules were unevaluable rather than unmatched.

After, the same deployment reports `reason=no_override_matched_window_metadata_absent`, which answers the question the operator actually has.

## Tests

Three cases added to `test/context-threshold.test.ts`:

- a window-range rule with no runtime window reports the new reason
- the same rule with a runtime window that misses the range keeps `no_override_matched`
- a non-window rule with no runtime window keeps `no_override_matched`

Full suite: 117 files, 1874 tests passing. `tsc --noEmit` clean.

## The other half

Option (3) in the issue — having OpenClaw put a recognised key in `buildAfterTurnRuntimeContext`, so the matchers work rather than merely announce that they cannot — is the real fix, and it is a host change rather than a plugin one. I am happy to send that upstream as well if you would like the feature to work end to end; this PR stands on its own either way, since the diagnostic stays useful for any host that reports no window.


jalehman:
Thanks for the contribution, @igs-rogenlo. Merged with a maintainer-owned patch changeset after review and CI passed. The diagnostic is useful independently of the host change: it makes missing window metadata visible without changing threshold selection. Issue #1161 remains open for the host-side metadata fix.

Issue #1024, opened 2026-07-22
Feature request: configurable toggle to preserve system events (Heartbeat Poll) from compaction filtering


cv-forever:
## Summary

lossless-claw v0.14.0 appears to filter or exclude OpenClaw system events (specifically `[OpenClaw heartbeat poll]` messages) from the active context during compaction or context assembly. These system events are no longer visible to the agent after upgrading from v0.11.3 to v0.14.0. Reverting to v0.11.3 immediately restores normal heartbeat poll delivery.

There is currently no configuration option to control whether system events are preserved in the fresh tail or excluded during context assembly.

## Reproduction steps

1. Configure OpenClaw heartbeat:
   ```json
   {
     "heartbeat": {
       "every": "90m",
       "activeHours": { "start": "08:00", "end": "23:50" },
       "prompt": "read HEARTBEAT.md from the workspace..."
     }
   }
   ```

2. Upgrade lossless-claw from v0.11.3 to v0.14.0 (bundled with OpenClaw 2026.7.1-2).

3. Manually trigger or wait for a heartbeat poll event.

4. Observe that the `[OpenClaw heartbeat poll]` message does not appear in the conversation context.

5. Revert lossless-claw to v0.11.3 and trigger another heartbeat poll — the message appears normally.

## Environment

- OpenClaw 2026.7.1-2 (0790d9f)
- lossless-claw v0.11.3 → v0.14.0 → reverted to v0.11.3 (confirmed working after revert)
- Model: deepseek/deepseek-v4-flash
- Platform: macOS (Apple Silicon)
- Channel: WebChat direct conversation
- LCM database: freshly rebuilt after upgrade

## Evidence

- After upgrade to v0.14.0: heartbeat poll events confirmed firing in gateway logs (`heartbeat: started` at correct intervals) but the agent never received the system event message.
- After revert to v0.11.3: next heartbeat poll arrived normally as `[OpenClaw heartbeat poll]` in the active conversation.
- No error or warning in logs indicating suppression — the events simply did not reach agent context.

## Related configuration options examined

- `LCM_PRUNE_HEARTBEAT_OK` — controls pruning of `HEARTBEAT_OK` agent replies, not the system event messages themselves.
- `ignoreSessionPatterns` — for session-level exclusion, not message/event-level filtering.
- `statelessSessionPatterns` — for read-only sessions, not applicable to main agent sessions.

None of these address system event preservation.

## Feature request

[...]

PR #1170, opened 2026-09-14
feat(config): add preserveHeartbeatPoll to keep heartbeat poll events

cv-forever:
> 心跳不是噪声——它是会话还在活着的证据。
> A heartbeat is not noise — it is evidence that the session is still alive.

## Problem

OpenClaw heartbeat polls arrive as synthetic system events, and Lossless Claw short-circuits them out of the transcript pipeline in two places:

- `LcmContextEngine.ingestSingle` returns `{ ingested: false }` for `isHeartbeat` messages.
- `reconcileVisibleTranscriptProjectionForAfterTurn` returns early for `params.isHeartbeat`, so the after-turn projection never sees them either.

The runtime perceives the poll, but nothing lands in LCM storage. For a session whose only traffic is heartbeats, the stored context can look as if the session never happened at all.

## Change

Add an opt-in `preserveHeartbeatPoll` flag that keeps heartbeat poll events in LCM storage and in assembled context.

| Config | Default | Env |
| --- | --- | --- |
| `preserveHeartbeatPoll` | `false` | `LCM_PRESERVE_HEARTBEAT_POLL=1` / `true` |

The default is `false`, so behaviour is unchanged for existing users.

Pure `HEARTBEAT_OK` acknowledgements remain prunable. `pruneHeartbeatOkTurns` now accepts `{ keepPoll }`, which deletes only the ack message and preserves the poll prompt — so `preserveHeartbeatPoll: true` + `pruneHeartbeatOk: true` expresses "keep the events, drop the noise" instead of accumulating thousands of empty ack turns.

## Files

- `src/engine.ts` — both `isHeartbeat` short-circuits gated behind the flag; prune call sites pass `keepPoll`.
- `src/heartbeat-filter.ts` — `pruneHeartbeatOkTurns(..., { keepPoll })`.
- `src/db/config.ts` — flag + env override.
- `openclaw.plugin.json` — config schema + uiHint.
- `docs/configuration.md`, `skills/lossless-claw/references/config.md` — docs synced per `AGENTS.md` (Config Schema Sync).
- `test/preserve-heartbeat-poll.test.ts` — new coverage; config literals in `test/{expansion,helpers,lcm-log}` updated.

## Verification

- `npm run typecheck` — clean
- `npm run build` — clean
- `npm test` — 118 files / 1880 tests passing
- New suite: 9 cases (default drop, opt-in ingest, poll preserved, ack still pruned, substantive replies untouched)

Changeset intentionally omitted per `AGENTS.md` ("Do not expect external contributors to know or run the Changesets workflow") — happy to add one if you'd prefer.

Refs #1024


jalehman:
Thanks for the contribution, @cv-forever. This is merged now. I added the minor changeset and regression coverage for opt-in pruning across durable commits, after-turn projection, and bootstrap reconciliation.

PR #1174, opened 2026-09-16
fix: cache stability - adopt repeated projection bodies by created time instead of importing a second row

Marvinthebored:
## Summary

- Adopt a projected transcript entry id onto a live-ingested user row by the row's created time when the message body repeats in the projection.
- Applies to both adoption paths: the decorated-body path in `engine.ts` and the exact-match tail path in `batch-dedup.ts`.
- A single exact match keeps adopting as before; the created-time check only decides between several matching rows, and several rows within the same second still fail closed.

## Why

Reconcile stamps a transcript entry id onto the live-ingested row for the same turn. That adoption required the body to be unique in the projection (`adoptDecoratedProjectionEntryId`) or a single exact match in the recent tail (`adoptRecentTranscriptEntryIdForMessage`). Bodies that repeat word for word, such as OpenClaw's restart-recovery prompt `[System] Your previous turn was interrupted by a gateway restart…` or a plain `ok`, matched neither rule, so the entry was imported as a second row next to the live one. Assembly then alternated between the two rows across runs, one with a restored timestamp and one without, and every message after that index changed bytes for the provider. On a 300k to 500k token session that was a full prompt-cache rewrite per run boundary.

Live ingest stores the message's own timestamp as `created_at`, so the transcript entry's time lands within milliseconds of the correct row. `filterByCreatedAt` in `src/transcript.ts` selects candidates within two seconds and parses timezone-less transcript strings as UTC, matching how the store reads `created_at`. `listRecentUnstampedMessagesByRole` now returns `createdAt`; it has one caller.

This is the engine-side counterpart of OpenClaw PR openclaw/openclaw#149690, which fences the engine's bootstrap read at the admitted user turn so the pending turn is not imported at all. That PR covers the turn currently being submitted; this change covers repeated bodies on later bootstraps, which the fence does not reach.

## Validation

Regression tests in `test/sqlite-bootstrap-projection.test.ts`:

- `adopts a repeated runtime body by created time instead of importing a second row`: two identical restart prompts one minute apart, both live-ingested; the projection's entry ids land on the matching rows and no third row is imported. Fails on the unpatched engine (a second row is imported, the older row stays unstamped) and passes patched.
- `adopts by created time even when store order disagrees with projection order`: store rows ingested newest-first; an oldest-first or first-in-tail pick would stamp the wrong row.
- `fails closed when repeated bodies share the same created second`: twins within one second receive no id.
- Existing `does not adopt an ambiguous repeated decorated runtime body` and `does not adopt an ambiguous decorated projection candidate` are unchanged and still pass.

Full suite: 118 files, 1,924 tests passed. Typecheck passed.

Observed on a production session before the fix: `lcm.db` held two rows for each restart prompt, one without `transcript_entry_id` from live ingest and one with it from the next reconcile, and the assembled prefix diverged at that turn on alternate runs.

## Release note

Includes a patch changeset for `@martian-engineering/lossless-claw`.

---

Filed with AI assistance (Claude Fable 5.1).


pavonis-martian:
Thanks for the contribution, @Marvinthebored. This is landed now. Final review of the post-conflict head found no actionable findings, all three CI checks passed, and the patch changeset is included.

PR #1175, opened 2026-09-16
perf: avoid redundant work during transcript reconciliation

PollyBot13:
## Summary

Reduce avoidable work while reconciling an already-ingested transcript, without changing transcript identity, persisted token counts, or anchor validation.

- Extract `toStoredMessageIdentity()` for role/content normalization and use it in four identity-only audit/reconciliation sites. Keep token accounting in `toStoredMessage()` for persistence and budgeting.
- Return before identity-hash queries and incoming-payload normalization when every row in the recent adoption window already has a transcript ID (including an empty window).
- Add three regression cases and a patch changeset.

## Boundary and compatibility

The owner is LCM's transcript reconciliation path. No schema, configuration, dependencies, host API requirements, retention, compaction policy, or public tool contracts change. No persisted data is deleted. Existing audit/epoch checks, structured tool-anchor verification, ambiguous-match refusal and mixed-tail selection remain authoritative.

Based on `c84dd8c`; preserves the newer structured-anchor behavior from #1162. Related #1174 addresses repeated-body adoption correctness; this PR does not implement or supersede that work. The only remaining batch-dedup production delta here is the fully-anchored-tail early return.

## Verification

Exact tested head: `615d95842e652c00d0e216576ad419b3022511ca`.

- `npm test -- --reporter=dot`: **119 files, 1,924 tests passed**.
- `npm run typecheck`: passed.
- `npm run build`: passed (plugin, migration and CLI bundles).
- `npm run test:tui`: passed.
- `git diff --cached --check` before commit: passed.
- Independent source review: no blocking findings.

The new regression tests use real in-memory SQLite migrations/stores. They check identity equivalence without structured token-accounting calls, avoid hash queries/adoption for fully stamped tails, and verify that a mixed tail adopts the unstamped row while preserving existing IDs. Existing structured-anchor and replay tests cover the newer upstream correctness rules. These are deterministic work-elimination checks, not wall-clock assertions. The helper test is not exhaustive instrumentation of every engine call site.

### Performance evidence and limits

Earlier qualification of this patch on the stable 1.0.0 source (`988dee85592b9066ffe1c859542e8e18c23f2345`) used three alternating baseline/candidate pairs with a frozen long-history transcript and separate consistent SQLite copies. Model/Gateway calls were disabled. Median bootstrap was **5,467 ms → 2,391 ms** (56% lower). Each pair returned the same bootstrap result and matching semantic database state; administrative update/audit timestamps were excluded from comparison. The unmodified control build matched the installed original bundle.

That historical experiment used private data and is **not independently reproducible from this PR**, nor is it a benchmark of this rebased head. No private transcripts, database dumps, or raw operational logs are published. The committed synthetic regressions and current-head suite are the reproducible evidence here.

The stable-base build also passed an isolated real-model/Discord canary and a subsequent deployment smoke check. Small, nonpaired live samples showed context preparation around 4.8–5.5 seconds versus an earlier ~7.5-second sample, but **did not establish an end-to-end latency improvement**. Emergency compaction still caused multi-minute prompt preparation. Fixing that, and core/Codex preparation optimizations, are explicit non-goals. The rebased PR build has not replaced that deployed stable-base build.

## Review and rollback

Please review normalization equivalence and whether the anchored-tail guard preserves adoption eligibility. This is one small performance change (three production files, one test file, one changeset). Reverting it restores the previous work without a data migration. Merge remains a maintainer decision.

AI-assisted implementation and review using Codex.


jalehman:
Thanks for the contribution, @PollyBot13. This is landed now as 6586d56efa7c53af7eb8c6807335f79cdcc9e42e. Reviewed head 615d958 against c84dd8c: structured autoreview found no actionable issues, 85 focused tests plus typecheck/build passed locally, and exact-head CI was green. The patch changeset is included. I also tested the conflict-free combination with #1174 at c90353d: 70 focused tests and typecheck passed. The historical benchmark was not used as evidence for this head.

PR #1180, opened 2026-09-17
fix(assemble): stop replaying the persisted copy of the live current turn on the loop-hook path

### LINKED ISSUE #1168: Current user turn appears twice in the model input on 2026.9.x: replayed ingested row plus core's live envelope
On every turn, the model input contains the current user turn twice:

1. A bare replayed row of the current message: no timestamp, no conversation
   metadata, no injected memory blocks. This is lossless-claw's own store
   replaying a message it already ingested, surfaced through `assemble()`.
2. A decorated live envelope of the same message: timestamp, conversation
   info, and injected memory blocks, appended by [openclaw/openclaw](https://github.com/openclaw/openclaw)
   after `assemble()` returns.

Observed on a couple of different lanes (one agent's Slack DM, a dashboard
turn) with a plain one-word message ("ping") reproducing it cleanly, so the
duplication is not content-dependent.

### Mechanism

[openclaw/openclaw](https://github.com/openclaw/openclaw) core 2026.9.2 does **not** include the incoming user
turn in the `messages` array it hands to `ContextEngine.assemble()`; the
current turn is passed separately as a `prompt: string` field ("the incoming
user prompt for this turn, useful for retrieval-oriented engines"). After
`assemble()` returns, core unconditionally builds a fresh `{role:"user",
content, timestamp}` message for the current turn and appends it to whatever
`assemble()` returned, before sending the request to the model. This append
is not gated on what `assemble()` produced, so if the engine's own returned
`messages` already contains something representing the current turn, the
model sees both.

On the plugin side (v1.0.0), `assemble()` already accounts for this contract
(`hostDeliversCurrentTurnSeparately = params.prompt !== undefined && ...`),
and `src/live-coverage.ts` carries the matching dedup logic:
`assembledRowIsStructuralBareCurrentTurn`,
`resolveStructuralCurrentTurnLiveIndex`, and
`messagesHaveSameLiveCoverageSignature` are meant to recognize "this bare
row I'm about to replay is the same turn as the live row the host will
append" and suppress the replay. In practice this structural match (content
containment / decorated-body matching between a bare stored row and a
timestamp+conversation-info+memory-decorated live row) misses in some cases,
and the dedup fails safe toward keeping both rows rather than risking
dropping real content.

### Nearest existing issues

- [#1158](https://github.com/Martian-Engineering/lossless-claw/issues/1158): different symptom, not this same-turn double-count.
- [#1028](https://github.com/Martian-Engineering/lossless-claw/issues/1028): pre-dates the current live-coverage structural matching path.
- [#1167](https://github.com/Martian-Engineering/lossless-claw/issues/1167): adjacent replay/coverage report, but not the bare-row-vs-decorated-envelope case here.

(Flagging these so triage can confirm this is a genuinely new case, not a
variant worth folding into one of them.)

### Suggested fix direction

- When core will append the live message itself (`prompt` present in
  `assemble()`'s params), skip replaying the store row whose ingest belongs
  to the current turn, instead of relying on post-hoc structural matching
  against the decorated envelope.
- When the live message is genuinely absent from `messages` (no separate
  `prompt`), fall back to matching on raw content rather than only on
  structural/decorated-body containment, so the dedup doesn't fail safe into
  keeping a duplicate whenever decoration breaks the substring match.

gorkem2020:
Fixes [#1168](https://github.com/Martian-Engineering/lossless-claw/issues/1168).

## What happens on current hosts

OpenClaw 2026.9.x calls the context engine at two points per turn. The pre-prompt call hands assemble the prior history and the current turn as the separate `prompt` scalar, then the host submits its own decorated copy of the turn. The mid-turn loop hook (`transformContext`) first calls afterTurn with the transcript projection, which persists the live user turn, and then calls assemble with the provider messages, which still contain that live turn.

On the second path the engine replayed the row it had just persisted next to the live copy. Two outcomes, depending on the live copy's shape:

- the structural matcher recognised the live copy as the decorated face of the stored row, dropped the live copy and shipped the transcript form, so the host's prompt-only injections (timestamp, memory and rule blocks) never reached the model;
- the match failed (memory or rule blocks sit between the sender metadata and the body) and both copies shipped, so the model saw the current turn twice.

Fleet trace on one turn (plugin log, two assemble lines 32 ms apart): pre-prompt `inputMessages=2 outputMessages=2`, then loop hook `inputMessages=3 outputMessages=4` with `volatileLiveInputsAppended=1`; the fourth message was the second copy of the current turn.

The pre-prompt path shows the same defect from the other side on hosts whose visible transcript projection already carries the admitted turn (the chat pipeline on our fleet): bootstrap imports it right before assemble, assemble replays it as the store's final user row, and the host then submits its own decorated copy. On that session the plugin logged `inputMessages=1 outputMessages=112` and the provider request carried the turn twice.

## Change

- `afterTurn` records the user rows it persisted for the open turn (row ids, stored content, host entry ids), keyed by session queue key.
- `bootstrap` records the user rows it imported the same way, since on some host paths the projection already carries the admitted turn.
- `assemble` on the loop-hook path drops those rows by identity and force-appends the live copy. On the pre-prompt path (`prompt` delivered separately) it drops the store's final user row when it is one of those rows and is absent from the host history, and appends nothing: the host owns the current turn there.
- The record is cleared by `commitTurn` when the turn advances and expires after 30 minutes as a safety net.
- `collectUncoveredVolatileLiveInputs` accepts caller-identified live indexes so an identity-selected live turn is never treated as covered by a text match.

Identity instead of text shape: the engine already knows which rows it wrote for the turn; the previous heuristic re-derived that correspondence from body text and every new host decoration was a silent miss.

## Tests

`test/host-carried-live-turn.test.ts` on the visible-transcript projection harness: first loop iteration, later iterations of the same turn, a live copy carrying injected memory blocks, a projection-imported turn on the pre-prompt path, a history row the host also carries, a pre-prompt assemble whose history holds the row, and a session without a persisted current turn. Four cases fail on the previous code. Full suite and typecheck green.


gorkem2020:
Live proof on our fleet (OpenClaw 2026.9.2, this branch built on main 6586d56, one agent, host-route turn with a tool call), plugin telemetry per assemble call:

Before (same session, one turn earlier, unpatched 1.0.0 build):

```
loop hook   contextItems=5 inputMessages=5 outputMessages=6 volatileLiveInputsAppended=1
```

After:

```
pre-prompt  contextItems=6 inputMessages=6 outputMessages=6
loop hook   contextItems=7 inputMessages=7 outputMessages=7 volatileLiveInputsAppended=1 hostCarriedLiveTurnRowsSuppressed=1
loop hook   contextItems=9 inputMessages=9 outputMessages=9 volatileLiveInputsAppended=1 hostCarriedLiveTurnRowsSuppressed=1
```

The provider request captured by our tracing carried the current turn exactly once, as the host's decorated copy (timestamp plus injected memory context plus the body). Before the change the same request shape carried it twice.

Second revision, pre-prompt path on the chat pipeline (same fleet, the agent's main session, host projection already carrying the admitted turn):

```
before  contextItems=117 inputMessages=1 outputMessages=112                                   (request carried the turn twice)
after   contextItems=119 inputMessages=3 outputMessages=113 hostCarriedLiveTurnRowsSuppressed=1  (request carried it once, the host's copy)
```

Host status, for scoping: OpenClaw main fixed the bootstrap half of this on 2026-09-16 ([openclaw#149690](https://github.com/openclaw/openclaw/pull/149690), not in the 2026.9.4 release) and its loop hook now defers ingestion for admitted turns, so on that contract nothing is persisted before assemble and this change is a no-op. It matters for released hosts (2026.9.2 through 2026.9.4), where both paths still replay the persisted copy. Contract discussion upstream: [openclaw#150956](https://github.com/openclaw/openclaw/issues/150956).

gorkem2020:
Closing this one: OpenClaw 2026.9.5 (released 2026-09-19) carries the host-side fix, [openclaw#149690](https://github.com/openclaw/openclaw/pull/149690), which the 1.1.0 changelog already names as the resolution for the bootstrap duplicate.

Verified on a live gateway today with lossless-claw 1.1.0: on core 2026.9.2 this branch's guard fired on every turn (`hostCarriedLiveTurnRowsSuppressed=1`) and kept the model input single-copy; after the upgrade to 2026.9.5 the guard never fires on either the pre-prompt or the loop-hook path and the model input still holds exactly one copy of the current turn (checked in Langfuse on both the webchat and the `openclaw agent` route). So the plugin-side exclusion is redundant on hosts at or above 2026.9.5, and the maintainer's reading (host issue, fixed upstream) holds. The write-up in [#1168](https://github.com/Martian-Engineering/lossless-claw/issues/1168) stays as the record of the two host paths for anyone still on 2026.9.2 through 2026.9.4.

PR #1184, opened 2026-09-18
fix: own background summarization through host maintenance

### LINKED ISSUE #1183: Background summarization can inherit a closed OpenClaw async work scope
## Summary

Lossless starts deferred compaction and pending-summary preparation with unreturned `setImmediate()` callbacks. These inherit OpenClaw's caller-owned async work scope, but do not keep that scope alive. Once the caller finishes and the host drains it, a delayed summarization request can fail with `Async work scope is closed` before contacting the provider.

Observed with Lossless 1.0.0 source (the scheduling code is present on main at `6586d56`) and OpenClaw 2026.9.4. Pending summaries themselves are normal precomputation; this report concerns failed completion admission, not an unpublished batch simply being pending.

## Affected paths

- `LcmContextEngine.schedulePendingSummaryPreparationDrain()`
- `LcmContextEngine.scheduleDeferredCompactionDebtDrain()`
- The workers can reschedule themselves or attach unreturned continuations to busy session queues.
- `dispose()` is a no-op for the shared engine and does not join these timers.

## Local reproduction

Using the actual Lossless scheduler and OpenClaw `AsyncWorkScope`:

1. Invoke the scheduler inside `scope.track(...)`.
2. Let the scheduling method return, then `await scope.drain()`.
3. Let its `setImmediate` callback call the actual OpenClaw `createRuntimeLlm().complete()` with synthetic configuration.
4. Completion admission rejects before provider acquisition. Network calls were prohibited by the reproduction.

Stack:

```text
Error: Async work scope is closed
    at AsyncWorkScope.track (src/shared/async-work-scope.ts)
    at captured async work tracker
    at Object.complete (src/plugins/runtime/runtime-llm.runtime.ts:588)
    at background pending-summary worker
```

Both schedulers reproduce the closed-scope failure using the host's completion-admission primitive. A control using awaited work inside `runContextEngineMaintenanceWork()` admits the same delayed work successfully.

Production logs also show `runtime.llm.complete error: Async work scope is closed`. The isolated reproduction establishes the scheduling defect; it does not establish that every lifecycle error or plugin hot-reload timeout has this cause.

## Expected / proposed fix

Use the existing host-owned background `maintain()` lifecycle and await summary preparation/deferred work within it. Preserve below-threshold prepare-only behavior and threshold publication rules. Do not bypass host work tracking or retain call-bound LLM capabilities beyond their owner.

Add regression coverage showing foreground ingestion returns without detached model work, background maintenance retains completion ownership until settlement, and below-threshold preparation does not publish summaries.

Related error handling: Lossless currently labels this host lifecycle rejection as a provider failure and may reach deterministic fallback; it should not be mistaken for a provider outage or credential failure.

jalehman:
## Summary

Fixes #1183.

Move pending-summary preparation and deferred compaction out of detached foreground callbacks and into awaited, host-owned background maintenance. The previous callbacks could inherit a foreground async work scope that had already closed before the summarizer requested completion.

- Foreground scheduling records data-only requests; background `maintain()` owns one bounded pass using its current runtime context.
- Keep foreground ingestion available during model calls and preserve below-threshold preparation versus threshold publication behavior.
- Await ready-only foreground publication and remove unowned queue continuations.
- Classify closed async scopes as runtime lifecycle errors, avoiding provider retries and deterministic fallback for that condition.
- Include regression coverage, architecture documentation, and a patch Changeset.

## Scope

Isolated from the locally tested `feat/context-explorer` branch onto current `main`. No context-explorer UI changes are included. This addresses summary completion ownership; it does not claim to resolve the separate plugin hot-reload drain timeout.

## Validation

- Reported working in the maintainer's local deployment.
- `npm run typecheck` — passed.
- `npm run build` — passed.
- `npm test` — 120 files, 1,929 tests passed on the isolated branch.
- `git diff --check` — passed.


digitalfrost84:
Follow-up for the carve-out in your scope note ("does not claim to resolve the separate plugin hot-reload drain timeout"): we have a live reproduction of that drain timeout on 1.1.0 (published after this merge), filed as #1194.

Host side is OpenClaw 2026.9.5: the replacement's drain of the context-engine instance exhausts the 5 s observer bound and the 60 s recovery window, the generation is abandoned ("replacement not applied"), and every plugin instance — this one, a channel plugin, a harness plugin and local hook plugins — stays retired until a full process restart. The changed config was unrelated to this plugin each time.

Since 1.1.0 postdates this fix, the instance-holding work appears to be outside the summarization-completion path. #1194 lists what we can collect at replacement time (in-flight queue lengths, scheduling timers, engine state) if you want to attribute it precisely.

Issue #746, opened 2026-05-23
Feature request: background-only deferred compaction drain to avoid mid-turn/tool-loop stalls


lysopen:
## Summary

`proactiveThresholdCompactionMode: "deferred"` is currently not equivalent to "background-only" compaction. Deferred threshold debt can still be consumed from `assemble()` / pre-assembly paths, and in OpenClaw tool-call loops this can block the active user turn midway through tool execution.

I would like a mode that only drains deferred threshold compaction debt after the assistant response has been delivered, or from an idle/background maintenance worker, never from prompt assembly / pre-assembly during the active turn.

## Why this matters

In a Telegram production workflow with frequent tool calls, the current behavior is disruptive:

1. User sends a message.
2. Agent starts working and may call tools.
3. During the tool loop, OpenClaw calls context-engine `afterTurn()` / `assemble()` per iteration.
4. Lossless sees pending deferred threshold compaction debt.
5. `assemble()` drains that debt and starts compaction before the agent can finish replying.
6. The user experiences the conversation as "the agent started working, then froze to compact history".

The desired behavior is:

```text
user message
→ agent runs all needed tool calls
→ assistant response is delivered to user
→ deferred compaction runs after the response / in background / when idle
```

## Current behavior observed

Current setup:

- OpenClaw production version: `2026.5.18`
- lossless-claw: `0.11.2`
- `proactiveThresholdCompactionMode: "deferred"`
- summary model: `cliproxyapi/gpt-5.4-mini`
- tuned sweep caps tested:
  - `contextThreshold: 0.82`
  - `maxSweepIterations: 3`
  - `sweepDeadlineMs: 45000`
  - `compactUntilUnderDeadlineMs: 90000`
  - `sweepMaxDepth: 1`

After a gateway restart, the new settings were confirmed live: threshold target changed from ~150k to ~164k and threshold compactions used 3 passes as configured.

Example observed compactions after restart:

- Conversation `1952` / Telegram topic `43162`:
  - `164756 -> 80026`
  - 3 passes
  - ~56.5s
- Conversation `2375` / Telegram topic `46887`:
  - `164909 -> 92000`
  - 3 passes
  - ~57.2s

These caps are useful, but they only limit how long each sweep can run. They do not prevent compaction from running on the active turn path.

## 0.11.2 makes the issue more visible

The new 0.11.2 controls are helpful:

- `maxSweepIterations`
- `sweepDeadlineMs`
[...]

Issue #1171, opened 2026-09-15
commitTurn compaction evaluation falls back to a hardcoded 128000 budget; breaks large-context-window models


BottaniCals:
### Summary

When lossless-claw's `commitTurn` handler runs proactive compaction evaluation, the resolved token budget is undefined — the host context-engine hook for `commitTurn` doesn't supply a budget by design — so the plugin falls back to a hardcoded `defaultTokenBudget = 128e3` and runs the entire threshold/target computation against that value. `maxAssemblyTokenBudget` is applied only as a ceiling (`Math.min(budget, cap)`) and cannot raise it.

For a 1M-context model this means the compaction target becomes `contextThreshold × 128000 ≈ 108k`, which triggers spurious compaction on any conversation whose frontier exceeds ~109k tokens. The pass then almost never reaches the target — the protected fresh tail plus already-summarized context dominate the post-compaction size — so the run records `compacted_but_still_over_target`, maintenance stays pending, and `/lossless` reports `lcm health: degraded` until something else (e.g. a fresh session) interrupts the loop. This affects every sufficiently-large conversation, not just edge cases.

### Plugin config in use

The `plugins.entries.lossless-claw.config` block tuned for a 1M-token model, after experimentation to find values that wouldn't break a 1M context window:

```json
{
  "ignoreSessionPatterns": [
    "agent:*:dreaming:**",
    "agent:*:dreaming-narrative-**",
    "agent:*:cron:**",
    "agent:*:**:active-memory:**"
  ],
  "freshTailCount": 128,
  "contextThreshold": 0.85,
  "maxAssemblyTokenBudget": 900000,
  "leafChunkTokens": 100000,
  "leafTargetTokens": 12000,
  "condensedTargetTokens": 10000,
  "largeFileThresholdTokens": 120000,
  "maxExpandTokens": 25000,
  "sweepDeadlineMs": 600000,
  "compactUntilUnderDeadlineMs": 1500000,
  "delegationTimeoutMs": 600000,
  "summaryTimeoutMs": 300000
}
```

[...]

PR #1191, opened 2026-09-19
fix: keep deferred summary preparation out of foreground assembly

PollyBot13:
## Summary

Prevent an over-budget user turn from waiting for a sequence of deferred model-backed summaries that may still leave the batch unpublished. The emergency assembly path now attempts only validated publication of already-ready summaries; incomplete work stays durable for host-owned background maintenance and assembly uses the existing live-context fallback.

- Preserve publication validation, session queue ordering, persisted history, manual/forced compaction and host admission.
- Do not clear model-spend backoff or erase persisted provider retry state when ready-only publication makes no progress.
- Add owner-boundary regression coverage for incomplete/ready/invalidated batches, durable recovery and retry state.

Related: #746, #313. Builds on host-owned maintenance from #1184; it does not replace that lifecycle contract.

## Validation

- Full typecheck and full build pass.
- 127 test files / 2,009 tests pass.
- Unmodified current-main negative control fails both new zero-foreground-summary-call cases.
- Independent source review completed; retry-state finding fixed and reviewed.

## Isolated real-model / Discord qualification — PASS

Tested LCM head `550fae7e4903b4fc8a775de816ea0b2c780e7880` (base `c168562dc6137e40440654c814170666b121f3e4`). All ten changed source/document files match the recorded source hashes. Packaged executable SHA256: `27317f6913fe7d749ed9004931ab73c5a493f877ea07e3bcbe54f89f385e0fd4`.

**Host prerequisite:** this preserves #1184's existing host-owned `maintain()` contract; it does not add a new plugin API. The frozen 2026.9.3 host failed background recovery, and inspected installed 2026.9.5 also lacks the durable-turn maintenance handoff. Successful qualification used an isolated source build: frozen base `1391f7cd2d40ab5bbcf2f5f831d3a64f520e72d7` plus the compatible runtime-facts prerequisite hunks from openclaw/openclaw#142871 (`d6997b6bbf2e269887f03bddae6a488d74908af2`) and maintenance handoff hunks from openclaw/openclaw#151936 (`dc5b321b5f89f9cf0e704ee202ea3059fb1ce63d`). Unrelated subagent-registry hunks were excluded. This is **not** a claim that stock 2026.9.3/2026.9.5 provides background liveness, nor qualification of all changes in those upstream PRs. No first stable release containing the repair has been established; do not infer a new stable version floor.

- Host package SHA256: `2b545032c9e0733e686c99ad04128d1ea2c6d55270fe3da14f435577311e97c2`
- Immutable isolated image: `sha256:65e91edb6e2ba5168c052c95b7e93f4e539e50b092b047412e83c0832b63359f`
- Host build/typecheck, 48 targeted tests, package integrity and image identity passed.

**Reproduction:** dedicated test Discord bot/channel, real `openai/gpt-5.6-terra` turns, disposable SQLite state in a read-only/no-host-mount Colima container. After a real conversation, stop the isolated Gateway and append 15 synthetic historical observations plus a checkpoint, plan an incomplete batch, and record threshold debt. Restart before inbound; assert >12 unfinished nodes and 12,256 stored tokens against a 4,000-token assembly cap. Send two queued normal Discord turns, observe emergency ready-only fallback, and let host-admitted maintenance prepare/publish. One ordinary follow-up after observed progress admits the remaining maintenance pass; no direct/manual `maintain()` or fabricated summaries are used. Finally verify restart/health, platform replies and cleanup.

Public-safe extracted result:

```json
{
  "scope": "Emergency LCM step only; not full reply latency or SLA",
  "emergencyStepMilliseconds": [
    4.0,
    2.0
  ],
  "initialStoredTokens": 12256,
  "initialPendingNodes": 15,
  "initialHistoryRowsVerified": 22,
  "historyDigestsPreserved": true,
  "newCanonicalSummaries": 15,
  "fallbackSummaryCountUnchanged": true,
  "activeBatchAbsent": true,
  "maintenancePending": false,
  "maintenanceRunning": false,
  "ordinaryRecoveryFollowups": 1,
  "canaryResult": "pass"
}
```

The 4 ms / 2 ms measurements cover the emergency LCM step only, **not total reply latency, an SLA or a controlled comparison with the historical stall**. Unit regressions prove zero foreground completion calls; real telemetry proves actual pending/fallback and later background publication. All 22 initial rows retain IDs, sequence, transcript-entry identity, roles, content and structured-part digests; no fallback summaries were added. All 40 exported manifest entries were independently rehashed; secret scan passed; owned container absence was verified. The queued replies and recovery follow-up were read back from Discord. Evidence manifest SHA256: `c5d6d1685321f40ee1e289cdcf457e1c3221610c5fc5673b7aa7c4f925248c89`.

**Remaining maintainer decision:** rollout/release timing relative to a host providing the repaired durable-turn maintenance handoff. Keep this PR draft until that prerequisite is accepted; users on affected hosts should retain a known-working deployment or explicitly choose legacy inline behavior if foreground work is acceptable. Production was not changed.

## Scope and recovery

No stored-history deletion, new dependency in this PR, production deployment or Gateway configuration change. The change uses existing publication and fallback paths; fallback preserves protected current context and is not an unconditional hard token cap. Revert this patch to restore the prior emergency preparation behavior.

AI-assisted implementation and independent review; validation claims are limited to the checks described above.


a-faheem:
@jalehman, does this PR also cover a new user message waiting for background maintenance that is already running?

We saw this on OpenClaw 2026.9.6 with Lossless Claw 1.1.0. Someone was having a normal conversation with the assistant in a Slack thread. The assistant answered one question, then appeared to ignore the follow-up. The user edited the follow-up to add an explicit mention, and the assistant eventually replied, making it look as though the mention had woken it up.

Checking the logs showed a different sequence. Here are relative timings, with private details removed:

- T+0s: the original follow-up was received. Its recorded context marked it as an implicit mention because the bot was already participating in the thread; an explicit mention was not required.
- T+55s: the user edited the message to mention the bot explicitly.
- T+91s: context-engine background maintenance finished, after several model-backed summarization calls.
- T+92s: the original follow-up entered the model transcript and processing began.
- T+104s: the Slack response started appearing.

The timing and retained original message suggest that the follow-up was waiting on maintenance rather than being rejected by mention gating. This is an observed incident, not a controlled reproduction, and we have not tested this PR against it.

The distinction we are trying to understand: this PR removes deferred summary preparation from foreground assembly, whereas our case appears to involve a foreground turn waiting for an already-running background maintenance operation. In the installed host code, `beginForegroundSessionMaintenance()` requests cancellation of preemptible owners but then waits for the relevant owners' `done` promises. We have not established which cancellation/settlement path was taken during the incident.

Does #1191 address that wait as well, or should we open a separate issue for background-maintenance preemption/settlement, possibly in OpenClaw? Happy to keep that separate from the 128,000-token budget issue in #1171/#1192.

